PR #3, opened 2025-04-04
chore: fix license headers and add linting

jackye1995:
Add missing license headers

Add spotless and checkstyle for all modules


jackye1995:
Removed version check logic copied from lance repo, merging to test effectiveness.

PR #24, opened 2025-05-27
feat: add documentation

jackye1995:
Claude 4 Vibe-coded the documentation side, pending human fixes. Preview: https://jackye1995.github.io/lance-spark/

cc @yanghua


jackye1995:
@yanghua cleaned up all the garbage things, some examples might still not be fully correct but I think this is a good start we can merge and then iterate, what do you think?

PR #70, opened 2025-08-06
chore: refactor pom.xml, connect profile with modules

jiaoew1991:
Only load the required modules

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -116,18 +116,7 @@
     </properties>
 
     <modules>
-        <module>lance-spark-base_2.12</module>
-        <module>lance-spark-3.5_2.12</module>
-        <module>lance-spark-bundle-3.5_2.12</module>
-        <module>lance-spark-3.4_2.12</module>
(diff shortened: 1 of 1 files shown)


jackye1995:
> Similar projects like Spark and Iceberg also load modules based on profiles, which is more reasonable I think.

the whole project was written in maven not in gradle, so it's not as flexible in a few aspects. Some historical context: https://github.com/lancedb/lance/issues/3637. 

At least when I was trying to publish the jars to Maven central, I faced issue that I cannot publish jars for different versions separately because it will publish the same base jar multiple times which is not allowed. If there are other ways to solve that, then I think we could go with the approach in this PR more easily.

I checked what similar Maven-based multi-version Spark project like Hudi does, looks like the solution is also not ideal, and some package names are even wrong in Maven (e.g. some 2.13 jars are under 2.12 package). In the current way in this repo, at least everything is pretty cleanly separated. Not sure if @yanghua you have any better suggestions.

jiaoew1991:
close it

PR #73, opened 2025-08-07
fix: add field's metadata to spark schema

fangbo:
This PR adds Lance field's metadata to Spark schema's field. If the metadata is missed, the large_binary column which is blob is written as binary. This is conflicted with old data.


fangbo:
@jackye1995 This PR is ready, could you review it ?

jackye1995 (approved):
makes sense to me, we also need to fix this at create table level so user can directly create a blob encoding column from Spark, but that can be a separated task

PR #75, opened 2025-08-07
feat: add vector column support for Lance via  arrow FixedSizeList conversion

LuQQiu:
This PR adds support for writing Spark DataFrame `ArrayType` columns as Arrow `FixedSizeList` to enable vector embeddings storage and indexing in Lance format. Users can now mark array columns as vectors using metadata, enabling efficient similarity search and vector indexing for ML workloads.

copied Spark Dataframe to Arrow Type converter and add custom Spark Array<Float/Double> to Arrow FixedSizeList conversion.

### DIFF
--- a/lance-spark-base_2.12/src/test/java/com/lancedb/lance/spark/FixedSizeListDataFrameTest.java
+++ b/lance-spark-base_2.12/src/test/java/com/lancedb/lance/spark/FixedSizeListDataFrameTest.java
@@ -92,8 +92,8 @@ public class FixedSizeListDataFrameTest {
       Row firstRow = result.first();
       assertEquals(0, firstRow.getInt(0));
       assertEquals("text_0", firstRow.getString(1));
-      scala.collection.mutable.WrappedArray<Float> embeddings =
-          (scala.collection.mutable.WrappedArray<Float>) firstRow.get(2);
+      // Use getSeq for cross-version compatibility
+      scala.collection.Seq<Float> embeddings = firstRow.getSeq(2);
       assertEquals(128, embeddings.size(), "Embeddings should have 128 elements");
 
       // Verify values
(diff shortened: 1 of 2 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -37,6 +37,8 @@ import java.util.concurrent.atomic.AtomicInteger
 import scala.collection.JavaConverters._
 
 object LanceArrowUtils {
+  val ARROW_FIXED_SIZE_LIST_SIZE_KEY = "arrow.FixedSizeList.size"
```
jackye1995, line 40: is this an existing parameter, or something new? Feels strange to have carmel case here, if we can change it it should use something like `arrow.fixed-size-list.size` or `arrow_fixed_size_list_size`

lance-spark-base_2.12/src/test/java/com/lancedb/lance/spark/write/VectorWriteTest.java
```diff
@@ -0,0 +1,234 @@
...
+      rows.add(RowFactory.create(i, vector));
+    }
+
+    Dataset<Row> df = spark.createDataFrame(rows, schema);
+
+    // Write to Lance format
+    String outputPath = "/tmp/vector_test_float32.lance";
+    deleteDirectory(new File(outputPath));
+
+    df.write()
```
jackye1995, line 97: this probably covers the data frame write case. What about if we just run CREATE TABLE to create an empty table? for example https://github.com/lancedb/lance-spark/blob/main/lance-spark-base_2.12/src/test/java/com/lancedb/lance/spark/SparkLanceNamespaceTestBase.java#L135.

I think we can do something like

```
CREATE TABLE (vector ARRAY<FLOAT>)
TBLPROPERTIES (
  'vector.arrow.fixed-size-list.size'='1024'
)
```

And then you can do the conversion at https://github.com/lancedb/lance-spark/blob/main/lance-spark-base_2.12/src/main/java/com/lancedb/lance/spark/utils/SchemaConverter.java to take the additional table properties values at https://github.com/lancedb/lance-spark/blob/main/lance-spark-base_2.12/src/main/java/com/lancedb/lance/spark/LanceNamespaceSparkCatalog.java#L408

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
LuQQiu: it's a new parameter added for identify which column needs conversion, OKKK can do arrow.fixed-size-list.size or arrow_fixed_size_list_size

LuQQiu:
Require https://github.com/lancedb/lance-namespace/pull/187

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
LuQQiu: changed to arrow.fixed-size-list.size

lance-spark-base_2.12/src/test/java/com/lancedb/lance/spark/write/VectorWriteTest.java
LuQQiu: Good call! Forget about SQL statement!, have updated to use your approach

PR #80, opened 2025-08-26
chore: refector code to support features(update/delete) for different…

fangbo:
Different spark versions have different interfaces. If we add new features such as update/delete, we should add implementation according to specific spark versions. This PR is a code refactoring for some preparation. After this code refactoring, we can conveniently implement new features under specific spark modules.


jackye1995 (approved):
nice refactoring!

PR #87, opened 2025-09-03
fix: remove the restriction that path must have the suffix ".lance"

YannByron:
Spark Lance Connector requires that the ident of the table must have the suffix ".lance ", which is inconsistent with lance-core, so it is removed.


YannByron:
@jackye1995 @yanghua can you help to approve and review this PR?

PR #116, opened 2025-11-19
fix: set version of `maven-shade-plugin` used in `lance-spark-bundle-*` modules

geserdugarov:
Currently, there is no explicit version management of `maven-shade-plugin`, which is used in `lance-spark-budle-*` modules. And there is no `org.apache.maven.plugins:maven-shade-plugin:pom`, for instance, in [this Maven repository](https://repo1.maven.org/maven2/org/apache/maven/plugins/maven-shade-plugin/), which could lead to compilation fail by Maven.

This PR sets `maven-shade-plugin` version explicitly using in plugin management of the parent `pom`.


geserdugarov:
@jackye1995 Could you please take a look at the proposed changes?

Issue #127, opened 2025-11-28
Decimal column's filter can't correctly push down


fangbo:
# Table

```
decimal_test (
  id BIGINT,
  price DECIMAL(10,2))
```

# Query

```
select * from decimal_test where price > cast(2.0 as decimal(10, 2));
```

# Error

```
25/11/28 15:05:33 WARN TaskSetManager: Lost task 0.0 in stage 2.0 (TID 2) (core-1-2.emr-136bf49b537e35deb62e.cn-beijing.emr-volces.com executor 2): java.lang.IllegalArgumentException: Invalid user input: Error resolving filter expression (price IS NOT NULL) AND (price > 2.00): Invalid user input: Received literal Float64(2) and could not convert to literal of type 'Decimal128(10, 2)', /workspace/rust/lance-datafusion/src/logical_expr.rs:27:17, /workspace/rust/lance-datafusion/src/planner.rs:848:17
        at com.lancedb.lance.ipc.LanceScanner.openStream(Native Method)
        at com.lancedb.lance.ipc.LanceScanner.scanBatches(LanceScanner.java:122)
        at com.lancedb.lance.spark.internal.LanceFragmentScanner.getArrowReader(LanceFragmentScanner.java:109)
        at com.lancedb.lance.spark.internal.LanceFragmentColumnarBatchScanner.create(LanceFragmentColumnarBatchScanner.java:55)
        at com.lancedb.lance.spark.read.LanceColumnarPartitionReader.next(LanceColumnarPartitionReader.java:44)
        at org.apache.spark.sql.execution.datasources.v2.PartitionIterator.hasNext(DataSourceRDD.scala:120)
        at org.apache.spark.sql.execution.datasources.v2.MetricsIterator.hasNext(DataSourceRDD.scala:158)
        at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.$anonfun$hasNext$1(DataSourceRDD.scala:63)
        at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.$anonfun$hasNext$1$adapted(DataSourceRDD.scala:63)
        at scala.Option.exists(Option.scala:376)
        at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.hasNext(DataSourceRDD.scala:63)
        at org.apache.spark.sql.execution.datasources.v2.DataSourceRDD$$anon$1.advanceToNextIter(DataSourceRDD.scala:97)
[...]

PR #128, opened 2025-11-28
fix: fix decimal column's filter push down

fangbo:
Related to https://github.com/lance-format/lance-spark/issues/127


jiaoew1991 (commented):
Can you add a test case?

PR #131, opened 2025-12-03
feat: support dataframe load path API

jiaoew1991:
The current lance read and write requires catalog configuration, and it does not support s3://path/to/table.lance for direct read and write paths. The purpose of this PR is to not only support catalog mode, but also support dataframe load API to read and write lance.

example:
```
spark.read.format("lance").load("s3://path/to/a.lance")
```


jiaoew1991:
> Will this also be available for Scala 2.13 and Spark 4.0?

Theoretically, it's possible, but I haven't tested it in this environment yet.

PR #170, opened 2026-01-07
fix: persist() hanging bug

### LINKED ISSUE #169: Dataframe Persist() hanging
This infinite hanging seems to be caused by deadlock between executors and drivers where executors are waiting for serialized values (to begin reading LanceInputPartition). I took at the SparkUI and it seems like there are executors waiting but 0 tasks spawned.

Reproducing script:

```python
from pyspark.sql import SparkSession
from pyspark import StorageLevel

spark = SparkSession.builder \
    .appName("test") \
    .config("spark.jars.packages", "<your bundle version>") \
    .config("spark.sql.catalog.lance", "com.lance.spark.LanceNamespaceSparkCatalog") \
    .config("spark.sql.catalog.lance.impl", "<your catalog impl>") \
    .getOrCreate()

foo = spark.sql("select title_id from <dataset> 1")
foo.explain()
foo.persist(storageLevel=StorageLevel.MEMORY_AND_DISK_2) # hangs here
foo.count()
```

jtuglu1:
Fixes #169.

Prior to the fix, persist() would hang. After the fix, we are able to see the expected `1` return value.
```python
from pyspark.sql import SparkSession
from pyspark import StorageLevel

spark = SparkSession.builder \
    .appName("test") \
    .config("spark.jars.packages", "<your bundle version>") \
    .config("spark.sql.catalog.lance", "com.lance.spark.LanceNamespaceSparkCatalog") \
    .config("spark.sql.catalog.lance.impl", "<your catalog impl>") \
    .getOrCreate()

foo = spark.sql("select title_id from <dataset> 1")
foo.explain()
foo.persist(storageLevel=StorageLevel.MEMORY_AND_DISK_2) # hangs here
foo.count()
```

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceInputPartition.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceInputPartition.java
@@ -21,9 +21,10 @@ import org.apache.spark.sql.connector.expressions.aggregate.Aggregation;
 import org.apache.spark.sql.connector.read.InputPartition;
 import org.apache.spark.sql.types.StructType;
 
+import java.io.Serializable;
 import java.util.List;
 
-public class LanceInputPartition implements InputPartition {
+public class LanceInputPartition implements InputPartition, Serializable {
   private static final long serialVersionUID = 4723894723984723984L;
 
   private final StructType schema;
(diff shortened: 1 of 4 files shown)


jtuglu1:
Closing this for now as rebuilding from `main` seems to have fixed it.

PR #175, opened 2026-01-14
feat: support refreshing vended credentials in worker

jackye1995:
Based on https://github.com/lance-format/lance-spark/pull/174

Make the changes across the codebase so that we record namespace impl and properties in initial Spark catalog initialization time, and then pass it to worker to perform read and write to properly refresh vended credentials.

Also some refactoring to centralize global caches to `LanceRuntime`

cc @bryanck @jtuglu1

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceRuntime.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceRuntime.java
@@ -13,19 +13,27 @@
  */
 package org.lance.spark;
 
+import org.lance.namespace.LanceNamespace;
+import org.lance.namespace.LanceNamespaceStorageOptionsProvider;
+
 import org.apache.arrow.memory.BufferAllocator;
 import org.apache.arrow.memory.RootAllocator;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 /**
  * Runtime utilities for Lance Spark connector.
  *
- * <p>This class manages a global Arrow buffer allocator with size configurable via environment
- * variable.
+ * <p>This class manages a global Arrow buffer allocator and provides helper methods for namespace
+ * operations.
  *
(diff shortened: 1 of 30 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/LanceRuntime.java
```diff
@@ -78,6 +120,124 @@ private static long getAllocatorSize() {
...
+    return FRAGMENT_CACHE;
+  }
+
+  private static ReadOptions buildReadOptionsWithInitialStorageOptions(
+      LanceSparkReadOptions readOptions, Map<String, String> initialStorageOptions) {
+    if (initialStorageOptions != null && !initialStorageOptions.isEmpty()) {
+      // Merge initial storage options (from namespace.describeTable) with base storage options
+      Map<String, String> merged = new java.util.HashMap<>(readOptions.getStorageOptions());
+      merged.putAll(initialStorageOptions);
+      return new ReadOptions.Builder().setStorageOptions(merged).build();
```
bryanck: In a few places I noticed that only the storage options are set, does it matter that other read options are not cloned?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -99,6 +100,16 @@ case class AddIndexExec(
     val uuid = UUID.randomUUID()
     val indexType = IndexTypeUtils.buildIndexType(method)
 
+    // Get namespace info from catalog if available (for credential vending on workers)
```
jtuglu1: I see this logic is duplicated across OptimizeExec and AddExec. Can we make this a shared helper? I also wonder if it's easy to potentially forget to add this logic on new execs.

PR #179, opened 2026-01-19
fix: strip storage. prefix from catalog storage options

majian1998:
### Problem
When configuring the Lance Spark catalog with keys like `spark.sql.catalog.<cat>.storage.*`,
the connector previously stored the full catalog options map as `storageOptions`.
This caused keys such as `storage.oss_endpoint` / `storage.aws_access_key_id` to be passed down
to Lance native object store initialization, but the native providers expect **unprefixed** keys
(e.g. `oss_endpoint`, `aws_access_key_id`). As a result, cloud backends (OSS/S3/...) could fail
at runtime with errors like "OSS endpoint is required".

### Fix
Normalize catalog storage options in `LanceSparkCatalogConfig.from(...)` by:
- extracting only entries with the `storage.` prefix
- stripping the prefix before storing them in `storageOptions`

This aligns Spark catalog configuration with Lance native expectations and matches
the behavior of the Rust `DirectoryNamespaceBuilder::from_properties` parsing logic.

### Behavior change
Users should keep using `spark.sql.catalog.<cat>.storage.*` keys. The connector now correctly
passes them to native layers as unprefixed storage options.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkCatalogConfig.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkCatalogConfig.java
@@ -34,6 +34,8 @@ import java.util.Objects;
  */
 public class LanceSparkCatalogConfig {
 
+  private static final String STORAGE_PREFIX = "storage.";
+
   private final Map<String, String> storageOptions;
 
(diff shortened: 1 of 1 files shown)


majian1998:
Hi @jackye1995, this PR is approved—are there any remaining concerns before merging? Happy to make changes if needed.

PR #183, opened 2026-01-20
fix: update docker-build to use spark 3.5.8 and support docker compose v2

hamersaw:
The DOCKERFILE for testing uses spark 3.5.6 This version no longer exists at https://dlcdn.apache.org/spark/ which causes the `make docker-build` cmd to fail. Updating to spark 3.5.8, which as a minor version upgrade should not have dire consequences.

Also updating the Makefile to support docker compose v2.


jackye1995 (approved):
Thanks for the fix!

PR #194, opened 2026-01-29
feat: support UPDATE COLUMNS FROM syntax for column-level updates

jiaoew1991:
### Background

For the extremely large tables `target` and `source`, the data from source needs to be backfilled into target. When both tables exceed 1B rows, especially when `source` contains binary data, and given the massive data volume, which may reach several hundred TB, it is not possible to complete the task using the add column syntax in one go. The current add column syntax strictly requires that `target` and `source` do not have columns with the same name. In this scenario, it is desired to merge the data from source in batches (possibly around 100 times) into target, which requires the use of the update column syntax to update target in batches.

### Proposal
Add new SQL syntax: ALTER TABLE ... UPDATE COLUMNS ... FROM ... This allows updating existing columns using data from a source view, matching rows by _rowaddr. Only matching rows are updated; unmatched rows in source are ignored.

Key features:
- Column-level update using Lance's fragment.updateColumns() API
- Preserves _rowaddr and _fragid (no row rewrite)
- Lower write amplification compared to MERGE INTO

Usage:
```sql
CREATE TEMPORARY VIEW update_source AS
SELECT _rowaddr, _fragid, new_value as col1
FROM target_table WHERE condition;

ALTER TABLE target_table UPDATE COLUMNS col1 FROM update_source;
```

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -49,7 +49,8 @@ Lance provides SQL extensions that add additional functionality beyond standard 
 
 The following features require the Lance Spark SQL extension to be enabled:
 
-- [ADD COLUMN with backfill](operations/dml/add-columns.md) - Add new columns and backfill existing rows with data
+- [ADD COLUMNS with backfill](operations/dml/add-columns.md) - Add new columns and backfill existing rows with data
+- [UPDATE COLUMNS with backfill](operations/dml/update-columns.md) - Update existing columns using data from a source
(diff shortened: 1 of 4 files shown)


jackye1995 (approved):
nice! looks good to me overall, could you make sure we have the documentation updated as well?

fangbo:
@jiaoew1991 

We have a customer encountering the same issue — updating a very large table with a small table. Spark's native merge into scans all data from the huge table and performs heavy shuffling. Your feature perfectly solves our problem. 

Thank you very much!!

PR #198, opened 2026-02-02
feat: enable create / drop tags and using tags as version on select queries

hamersaw:
Adding support for tags in various APIs:

## Spark SQL
To create a new tag using the specified "<version>" or latest if not provided.
`ALTER TABLE <table> CREATE TAG <tag> [VERSION AS OF <version>]`

To delete an existing tag
`ALTER TABLE <table> DROP TAG <tag>`

To query a table using tag as version
`SELECT * FROM <table> VERSION AS OF <tag>`

## Spark API
```
spark.read()
    .option("version", "<tag>")
    ...
```

### DIFF
--- a/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -15,7 +15,7 @@ package org.apache.spark.sql.catalyst.parser.extensions
 
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedIdentifier, UnresolvedRelation}
 import org.apache.spark.sql.catalyst.parser.ParserInterface
-import org.apache.spark.sql.catalyst.plans.logical.{AddColumnsBackfill, AddIndex, LogicalPlan, NamedArgument, Optimize, Vacuum}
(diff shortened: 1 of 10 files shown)


hamersaw:
> Hi, @hamersaw I have a question to discuss.
> 
> Lance has branch feature. So, if we support to query from a branch, what do you think the Spark SQL's grammar to specify branch and tag ?
> 
> If we use:
> 
> ```sql
> SELECT * FROM <table> VERSION AS OF <tag>
> ```
> 
> We can not define the specific branch in the sql.

@fangbo , this is next up on my "random backfill" TODO. I tried to add an `ON BRANCH <branch>` clause to spark SQL statements because that would be wildly ergonomic - ex. `SELECT * FROM <table> [ON BRANCH <branch>] [VERSION AS OF <tag>]` but there is no clean way to do that without rewriting most of the grammer to be lance specific (similar to how `ALTER TABLE <table> CREATE TAG` is proposed here -- built on your COLUMN work).

I looked around a bit and saw Iceberg supports this by adding a prefixed ID to the table (ex. `SELECT * FROM db.table.branch_foo`). Without major grammer updates this seems the least intrusive approach because it needs to be supported across a TON of statments (ex. SELECT, ALTER TABLE, etc). I'm certainly going to open a discussion before submitting a PR on this and am VERY interested in others thoughts!

fangbo:
Hi, @hamersaw  will this work continue? We have a customer who needs to read and write branches in Spark.

fangbo:
> > Hi, @hamersaw will this work continue? We have a customer who needs to read and write branches in Spark.
> 
> @fangbo Unfortunately this has fallen down my priority list. I do not think I'll have time to pick this back up in the near term.

Hi, @hamersaw one of our customers is currently using branch/tag in production. I want to pick up this issue. I have submit a poc implementation about this feature https://github.com/lance-format/lance-spark/pull/563. I will split some small PRs to finish this work.

Issue #166, opened 2026-01-05
Support update using UpdateMode::RewriteColumns


fangbo:
This PR https://github.com/lance-format/lance-spark/pull/79 implements UPDATE using MOR. The update mode is delete-insert. The updated rows are deleted and new rows are inserted.

While the PR https://github.com/lance-format/lance/pull/4715 support fragment-level update columns using UpdateMode::RewriteColumns. The updated columns are written to new data files.

Each approach has its strengths: delete-insert works well when only a few rows need to be changed like:

```sql
UPDATE table_x SET col_a = xx, col_b = xx, col_c = xx WHERE id=xx
```

Whereas UpdateMode::RewriteColumns is better suited for bulk updates of a single column like:

```sql
UPDATE table_x SET col_a = xx
```

So, I plan to support two update modes in Spark. We can use a parameter `spark.sql.catalog.lance.update_rewrite_columns=true/false` to specify update mode. The default value is true means that using UpdateMode::RewriteColumns to do update. When the value is false, using MOR to do update.

@jackye1995 @jiaoew1991 @majin1102 What do you think about this idea ?

PR #199, opened 2026-02-03
feat: support update/merge-into using RewriteColumns mode

fangbo:
This is for https://github.com/lance-format/lance-spark/issues/166

1. add a new parameter `rewrite_columns` to specify update/merge-into to use RewriteColumns mode. If the parameter value is false RewriteRows mode is used which means that the rows are deleted and new updated rows are inserted.
2. UpdateColumnsExtractor rule is used to extract updated columns from spark sql: `update` or `merge into`. The specific updated columns will be injected to LancePositionDeltaOperation.
3. if `rewrite_columns` is true, LancePositionDeltaOperation.representUpdateAsDeleteAndInsert return false. It means that LanceDeltaWriter.update will be invoked for columns updating.
4. In LanceDeltaWriter.update , the new updated columns values are collected for a Fragment. Then Fragment.updateColumns is invoked to update the specific columns.

### DIFF
--- a/docs/src/operations/dml/update.md
+++ b/docs/src/operations/dml/update.md
@@ -34,3 +34,56 @@ SET tags = ARRAY('ios', 'mobile')
 WHERE event_id = 1001;
 ```
 
+## Column Rewrite Mode
+
+`lance-spark` introduces a column rewrite mode for `UPDATE` operations, which can significantly improve performance for narrow updates that only affect a few columns.
+
+This mode allows the Lance data source to perform column-level updates by writing new data files for only the modified columns, avoiding the need to rewrite the entire data file (the "delete and insert" pattern).
+
+!!! warning "Spark Extension Required"
+This feature requires the Lance Spark SQL extension to be enabled.
+See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+### Configuration
+
+You can enable or disable this feature using the Spark SQL session configuration `spark.sql.lance.rewrite_columns`.
+
+**Using SQL:**
+
+To enable the feature for the current session:
+```sql
+SET spark.sql.lance.rewrite_columns = true;
+```
+
+To disable it:
+```sql
+SET spark.sql.lance.rewrite_columns = false;
+```
+
+### Behavior and Semantics
+
+- When `spark.sql.lance.rewrite_columns` is set to `true`, `UPDATE` operations will attempt to perform column-level updates. Instead of deleting the matched rows and inserting new versions, the engine will only write new versions of the changed columns.
+- When the configuration is set to `false` (the default behavior), the operations fall back to rewriting the affected rows (a "delete and insert" operation).
+
+### Examples
+
+**UPDATE with RewriteColumns**
+
+Here is an example of enabling the mode and performing an `UPDATE`.
+
+```sql
+-- Enable column rewrite mode
+SET spark.sql.lance.rewrite_columns = true;
+
+-- Assume 'users' table has columns: id, name, address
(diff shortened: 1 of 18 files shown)


lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/optimizer/UpdateColumnsExtractor.scala
```diff
@@ -0,0 +1,132 @@
...
+              case _ => Seq.empty
+            }
+
+            op.setUpdatedColumns(updatedColumns.distinct.asJava)
+
+          case _ =>
+        }
+
+      } catch {
+        case _: Exception =>
```
jackye1995: this should not just fail silently

lance-spark-3.5_2.12/src/test/java/org/lance/spark/update/MergeIntoRewriteRowsTest.java
```diff
@@ -0,0 +1,42 @@
...
+        SparkSession.builder()
+            .appName("lance-merge-into-distribution-test")
+            .master("local[4]")
+            .config(
+                "spark.sql.catalog." + catalogName, "org.lance.spark.LanceNamespaceSparkCatalog")
+            .config(
+                "spark.sql.extensions", "org.lance.spark.extensions.LanceSparkSessionExtensions")
+            .config("spark.sql.catalog." + catalogName + ".impl", "dir")
+            .config("spark.sql.catalog." + catalogName + ".root", tempDir.toString())
+            .config("spark.sql.catalog." + catalogName + ".storage.rewrite_columns", "false")
```
jackye1995: seems strange that we are making this a storage option. I think this should be more like a Spark SQL conf, so we can do something like `SET` to enable/disable it if necessary.

lance-spark-3.5_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
```diff
@@ -316,7 +402,77 @@ public void delete(InternalRow metadata, InternalRow id) throws IOException {
...
+
+      currentUpdateWriter.finish();
+
+      ByteArrayOutputStream out = new ByteArrayOutputStream();
+      try (ArrowStreamWriter writer = new ArrowStreamWriter(currentUpdateData, null, out)) {
+        writer.start();
+        writer.writeBatch();
+        writer.end();
+      } catch (IOException e) {
+        throw new RuntimeException("Cannot write schema root", e);
```
jackye1995: I don't think you need to rethrow as RuntimeException? The method already does `throws IOException`

jackye1995 (commented):
mostly looks good to me, could you also update the related documentation?

docs/src/operations/dml/update.md
```diff
@@ -34,3 +34,74 @@ SET tags = ARRAY('ios', 'mobile')
 WHERE event_id = 1001;
 ```
 
+
+## Column Rewrite Mode
+
+`lance-spark` introduces a column rewrite mode for `UPDATE` and `MERGE INTO` operations, which can significantly improve performance for narrow updates that only affect a few columns.
```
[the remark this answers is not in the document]
fangbo: > I think you should also update the MERGE doc? Currently the update doc talks about both update and merge.

Thanks for your advice. I have added a new documentation [`merge-into.md`](https://github.com/lance-format/lance-spark/pull/199/changes#diff-9ab62816fbbf57c74474c83c1e600cd880740c172d2d33332985ef00b379d2c9)

PR #201, opened 2026-02-04
feat: support time travel

zhangyue19921010:
Support `TIMESTAMP AS OF ` and `VERSION AS OF`

```
spark.sql("SELECT * FROM " + fullName + " TIMESTAMP AS OF " + ts1);

spark.sql("SELECT * FROM " + fullName + " VERSION AS OF " + "2");
```

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -425,7 +425,7 @@ public abstract class BaseLanceNamespaceSparkCatalog
 
   @Override
   public Table loadTable(Identifier ident, long timestamp) throws NoSuchTableException {
-    return loadTableInternal(ident, Optional.of(timestamp), null);
+    return loadTableInternal(ident, Optional.of(timestamp), Optional.empty());
   }
 
   @Override


lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
```diff
@@ -0,0 +1,148 @@
...
+      builder.namespace(namespace);
+    }
+
+    if (versionId != null) {
+      builder.version(versionId.intValue());
+    }
+    return builder.build();
+  }
+
+  // Determine if the timestamp is in microseconds or nanoseconds and convert to Instant
```
hamersaw: In which cases is the timestamp given in either nanos or micros?

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -411,62 +416,69 @@ public boolean tableExists(Identifier ident) {
...
+    // Open dataset to get schema
+    StructType schema = getSchema(ident, location, readOptions, namespace);
+    return createDataset(
+        readOptions, schema, initialStorageOptions, namespaceImpl, namespaceProperties);
+  }
 
-    String location = describeResponse.getLocation();
-    Map<String, String> initialStorageOptions = describeResponse.getStorageOptions();
+  @Override
+  public Table loadTable(Identifier ident, String version) throws NoSuchTableException {
```
hamersaw: It looks like all of the loadTable instances are basically duplicated. Would it make sense to consolidate some of the code? maybe provide a "version" either from string parsing or timestamp lookup as an argument and if it doesn't exist than use latest?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
```diff
@@ -0,0 +1,148 @@
...
+public class Utils {
+
+  public static long parseVersion(String version) {
+    return Long.parseUnsignedLong(version);
+  }
+
+  public static long findVersion(List<Version> versions, long timestamp) {
+    long versionID = -1;
+    Instant instant = instantFromTimestamp(timestamp);
+    for (Version version : versions) {
```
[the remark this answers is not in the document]
zhangyue19921010: it is ordered based on https://github.com/lance-format/lance/blob/41fadef05062d0df834d90024c78ad67024681b4/rust/lance/src/dataset.rs#L1793

hamersaw (approved):
Looks good to me. I'm excited to build on this for `TAG` support!

jackye1995 (approved):
nice work!

Issue #196, opened 2026-01-29
EMR Spark 3.5 Arrow version conflict


jackye1995:
(from Discord)

 EMR (which still has spark 3.5) as the newer lance core also brings in arrow 18, which isn't compatible with the arrow 13 in spark 3.5 (no such method error for org.apache.arrow.vector.ipc.message.ArrowRecordBatch.<init>(...) )

PengleiShi:
@hamersaw Sorry, my mistake. This seems to be a Spark 3.4.1 specific issue. It works fine after upgrading to 3.4.4, so feel free to close this.

PR #203, opened 2026-02-04
fix: pin arrow version to 15.0

hamersaw:
The [recent upgrade of arrow version](https://github.com/lance-format/lance/commit/341a599dfaa376078f5c77bda7e000d1aaa249a3) (15.0 -> 18.3) in lance-core does not play well with Spark 3.5. This results in failures within the Lance Spark connector for simple operations (ex. `CREATE TABLE`). In this PR we pin the arrow version to 15.0 ([used for ~2 years](https://github.com/lance-format/lance/commit/035662907086f1bb284260043622e1c303c10696)) for all 3.5 and 3.4 profiles. The other alternative approaches are to (1) downgrade lance-core arrow dependency or (2) maintain separate lance-core branches. This seems like the most reasonable approach.

Closes: https://github.com/lance-format/lance-spark/issues/196

### DIFF
--- a/lance-spark-3.4_2.12/pom.xml
+++ b/lance-spark-3.4_2.12/pom.xml
@@ -15,9 +15,9 @@
     <packaging>jar</packaging>
 
     <properties>
-        <arrow.version>${arrow15.version}</arrow.version>
         <scala.version>${scala212.version}</scala.version>
         <scala.compat.version>${scala212.compat.version}</scala.compat.version>
+        <arrow.version>${arrow14.version}</arrow.version>
     </properties>
 
     <dependencies>
--- a/lance-spark-3.4_2.13/pom.xml
+++ b/lance-spark-3.4_2.13/pom.xml
@@ -15,9 +15,9 @@
     <packaging>jar</packaging>
 
     <properties>
-        <arrow.version>${arrow15.version}</arrow.version>
         <scala.version>${scala213.version}</scala.version>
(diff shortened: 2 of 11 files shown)


hamersaw:
> I recall we discussed about it in [lance-format/lance#5565 (comment)](https://github.com/lance-format/lance/pull/5565#issuecomment-3688163218), can we check if this would still work with java 21 and spark 4.0? We are currently only testing java 17: https://github.com/lance-format/lance-spark/blob/main/.github/workflows/spark.yml#L60

Arrow:
- [v15 tested on Java 8, 11, 17, 21](https://arrow.apache.org/docs/15.0/java/install.html).
- [v19 requires Java 11+](https://arrow.apache.org/java/main/developers/building.html#system-setup). So we can assume v18 supports Java 11.

Spark:
- We used arrow v15 on Spark 3.4 and 3.5 forever
- [Spark 4.0 uses arrow 18 internally](https://spark.apache.org/releases/spark-release-4-0-0.html)

I tested using arrow 15 on Spark 3.5 with Java 21 locally and everything worked. This PR doesn't change anything for our Lance Spark 4.0 story (still using arrow 18 on whatever Java is configured), but downgrades arrow version on 3.4 / 3.5. So if those work with newer Java versions IIUC that covers our concerns right?

hamersaw:
> Some reference for how Iceberg Spark does it:

... Java sorcery ...

PR #205, opened 2026-02-05
fix: align arrow version with spark and add docker integration tests

hamersaw:
The [recent upgrade of arrow version](https://github.com/lance-format/lance/commit/341a599dfaa376078f5c77bda7e000d1aaa249a3) (15.0 -> 18.3) in lance-core does not play well with Spark 3.5. This results in failures within the Lance Spark connector for simple operations (ex. `CREATE TABLE`). In this PR we pin the arrow version to 15.0 ([used for ~2 years](https://github.com/lance-format/lance/commit/035662907086f1bb284260043622e1c303c10696)) for all 3.5 and 3.4 profiles. The other alternative approaches are to (1) downgrade lance-core arrow dependency or (2) maintain separate lance-core branches. This seems like the most reasonable approach.

Future work will be to integrate these into our CI pipeline.

Closes: https://github.com/lance-format/lance-spark/issues/196

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,9 +20,6 @@ MODULE := lance-spark-$(SPARK_VERSION)_$(SCALA_VERSION)
 BUNDLE_MODULE := lance-spark-bundle-$(SPARK_VERSION)_$(SCALA_VERSION)
 BASE_MODULE := lance-spark-base_$(SCALA_VERSION)
 
-# Maven profiles for Spark/Scala version-specific settings
-MAVEN_PROFILES := -Pspark-$(SPARK_VERSION),scala-$(SCALA_VERSION)
-
 # Spark download versions for Docker
 include docker/versions.mk
 SPARK_DOWNLOAD_VERSION := $(SPARK_DOWNLOAD_VERSION_$(SPARK_VERSION))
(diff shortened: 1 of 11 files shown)


lance-spark-3.4_2.12/pom.xml
```diff
@@ -14,10 +14,6 @@
     <description>Apache Spark 3.4 Connector for Lance</description>
     <packaging>jar</packaging>
 
-    <properties>
```
jackye1995: why can we remove these? Don't we need them to ping the scala version?

hamersaw: We have [profiles at the root](https://github.com/lance-format/lance-spark/blob/acf3e169c53da278d232edd8052f7089464519ff/pom.xml#L440-L469) that set them. So we can either use those or set them explicitly in each module. I chose the former by updating the Makefile to set the profile based on the SPARK_VERSION env var. This way there is a single place to set this rather than within each modules pom.xml.

pom.xml
```diff
@@ -54,17 +54,23 @@
         <lance.version>2.0.0</lance.version>
         <lance-namespace-impl.version>0.1.3</lance-namespace-impl.version>
 
+        <arrow14.version>14.0.2</arrow14.version>
+        <arrow15.version>15.0.2</arrow15.version>
+        <arrow17.version>17.0.0</arrow17.version>
```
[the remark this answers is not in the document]
jackye1995: Oh I see. I always thought Spark 4.0 uses 18. Then let's use the one matching Spark one.

Makefile
```diff
@@ -12,14 +12,22 @@
 
 # Version parameters (can be overridden from command line)
 # Example: make install SPARK_VERSION=3.5 SCALA_VERSION=2.13
-SPARK_VERSION ?= 3.5
-SCALA_VERSION ?= 2.12
+SPARK_VERSION ?= 4.0
```
jackye1995: can we avoid changing the default for now?

lance-spark-3.4_2.12/pom.xml
hamersaw: Ok, then we should remove the profiles completely because there is no reason for them right?

PR #219, opened 2026-02-09
docs: add databricks vendor docs

hamersaw:
Add docs to outline compatibility with Databricks


docs/src/vendors/databricks.md
```diff
@@ -0,0 +1,137 @@
...
+    SPARK_VERSION=<version> SCALA_VERSION=<version> make bundle
+    ```
+
+    See the [lance-spark repository](https://github.com/lance-format/lance-spark) for more details.
+
+### 3. Configure Spark
+
+Configure the catalog to use `LanceNamespaceSparkCatalog`, then set catalog-specific and namespace-specific properties as needed. Refer to the [Lance Spark Config docs](https://lance.org/integrations/spark/config/#example-namespace-implementations) for all available namespace implementations.
+
+!!! note
```
jackye1995: I think this should be a part of the last step "Install the Lance Spark Library"?

docs/src/vendors/databricks.md
```diff
@@ -0,0 +1,137 @@
...
+    .format("lance") \
+    .saveAsTable("lancedb.default.my_table")
+
+# Read a table
+new_df = spark.read \
+    .table("lancedb.default.my_table") \
+    .collect()
+```
+
+## Known Issues
```
jackye1995: nit: Known Limitations

docs/src/vendors/databricks.md
```diff
@@ -0,0 +1,137 @@
...
+
+    The recommended approach is to use the Lance namespace catalog (`LanceNamespaceSparkCatalog`) directly, which bypasses Databricks catalog integration entirely. See [Configure Spark](#3-configure-spark) below.
+
+## Classic Compute Setup
+
+### 1. Create a Cluster
+
+Create a new cluster using **Databricks Runtime 16.4 LTS**. Other runtimes may work but have not been tested.
+
+The cluster **Access Mode** must be set to **No isolation shared**. This is required for using custom Spark extensions on Databricks. You can set this by using `Policy: Unrestricted` and configuring the access mode under **Advanced** cluster configuration.
```
jackye1995: this should also be a part of Known Limitations

hamersaw: Keeping "Supported Environments" at the top is not a hill I will die on, but I think this is very important to calling inline in cluster setup. I would argue that this isn't a "Known Limitation", but a setup detail that is required for it to work.

hamersaw:
I removed all the "examples" sections (ex. "Usage" and "Spark Configuration"). One second thought, I don't think they are very useful here as all of that is document in much more depth elsewhere.

docs/src/vendors/databricks.md
```diff
@@ -0,0 +1,59 @@
+# Databricks
```
jackye1995: we should add a "last updated at" date, since this kind of guide can easily go out of date.

docs/src/vendors/databricks.md
```diff
@@ -0,0 +1,59 @@
...
+### Supported Environments
+
+| Environment     | Catalog          | Support Status      | Notes                                                       |
+|-----------------|------------------|---------------------|-------------------------------------------------------------|
+| Classic Compute | Unity Catalog    | ❌ Not Supported     | Databricks uses a proprietary implementation not compatible with Spark extensions |
+| Classic Compute | Hive Metastore   | ❌ Not Supported     | Vended storage credentials through Databricks are not supported |
+| Classic Compute | Lance Namespace  | ✅ Supported         | Recommended approach; bypasses Databricks catalog integration |
+| SQL Warehouse   | —                | ❌ Not Supported     | Does not support custom Spark datasources or SQL Extensions |
+
+!!! warning "Databricks Catalogs"
```
jackye1995: this section seems like duplicated information with the table above?

PR #220, opened 2026-02-09
ci: run docker integration tests as part of ci

hamersaw:
Add an "Integration Tests" section to the CI that runs commands in an actual Spark env.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,7 +38,7 @@ else
 endif
 
 # Optional Docker build cache flags (set in CI for layer caching)
-# Example: make docker-build-minimal DOCKER_CACHE_FROM="type=gha" DOCKER_CACHE_TO="type=gha,mode=max"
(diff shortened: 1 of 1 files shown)


docker/tests/test_lance_spark.py
```diff
@@ -17,6 +17,12 @@
 
 SPARK_VERSION = Version(os.environ.get("SPARK_VERSION", "3.5"))
 
+# UPDATE and MERGE require Spark 3.5+ (RewriteUpdateTable/RewriteMergeIntoTable rules)
+requires_update_or_merge = pytest.mark.skipif(
```
hamersaw: Supposedly 3.4 does not have UPDATE or MERGE rules so we skip those tests.

hamersaw (commented):
All the files in `lance-spark-3.4_2.12` are copied from the 3.5 and enable deletes. This allowed a few more tests to pass for 3.4.

Issue #212, opened 2026-02-06
Run docker integration test in CI


jackye1995:
cc @hamersaw

PR #228, opened 2026-02-11
ci: azurite and minio integration tests

hamersaw:
Hoping to "best effort" catch issues with Azure Blob and S3 as part of our integration tests. We could run these against actual Azure Blob and S3 hosted buckets, but the requires large infrastructure. I think we should start with these and if we run into issues that are not caught, we can improve.

Related to https://github.com/lance-format/lance-spark/issues/212


jackye1995 (approved):
thanks for the work!

Issue #210, opened 2026-02-06
createOrReplaceTable does not fully work


jackye1995:
We need to have comprehensive test coverage and make sure we fully support the following:
- CREATE TABLE
- CREATE OR REPLACE TABLE
- REPLACE TABLE
- CREATE TABLE AS SELECT
- REPLACE TABLE AS SELECT

PR #251, opened 2026-02-12
feat: replace table and all variants (RTAS, CORT, CORTAS, etc)

hamersaw:
Adds a staging table implementation that supports:
`REPLACE TABLE`
`REPLACE TABLE AS SELECT`
`CREATE TABLE AS SELECT`
`CREATE OR REPLACE TABLE`
`CREATE OR REPLACE TABLE AS SELECT`

Closes https://github.com/lance-format/lance-spark/issues/210

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -71,6 +71,9 @@ venv.bak/
 **/.ipynb_checkpoints
 docker/*.jar
 
+# Python
+__pycache__/
+
 
 # mkdocs documentation
 /docs/site
--- a/docker/tests/test_lance_spark.py
+++ b/docker/tests/test_lance_spark.py
@@ -179,11 +179,6 @@ class TestDDLStagingTable:
         assert ids == [10, 20, 30]
         assert result[0].value == 100.0
 
-    # TODO(https://github.com/lance-format/lance/issues/5972): Investigate why this fails on Spark 3.4
-    @pytest.mark.skipif(
-        SPARK_VERSION < Version("3.5"),
-        reason="REPLACE TABLE with different schema fails on Spark 3.4 due to Lance SDK schema validation"
-    )
     def test_replace_table_as_select_different_schema(self, spark, test_table):
         """Test REPLACE TABLE AS SELECT with completely different schema."""
         # Create initial table with schema: (id INT, name STRING, value DOUBLE)
@@ -286,9 +281,6 @@ class TestDDLStagingTable:
         result = spark.table("default.test_table").collect()
         assert len(result) == 2
 
-    # TODO(https://github.com/lance-format/lance/issues/5972): Enable this test once Lance SDK
-    # supports passing explicit schema to Fragment.create() for REPLACE TABLE operations
-    @pytest.mark.skip(reason="Pending Lance SDK fix for schema override in fragment creation")
     def test_replace_table_schema_only(self, spark, test_table):
         """Test REPLACE TABLE with schema only (no data)."""
         # Create table with data


lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -519,6 +522,191 @@ public void renameTable(Identifier oldIdent, Identifier newIdent)
...
+  @Override
+  public StagedTable stageCreate(
+      Identifier ident, StructType schema, Transform[] partitions, Map<String, String> properties)
+      throws TableAlreadyExistsException, NoSuchNamespaceException {
+    Identifier actualIdent = transformIdentifierForApi(ident);
+    List<String> tableIdList = buildTableId(actualIdent);
+    StructType processedSchema = SchemaConverter.processSchemaWithProperties(schema, properties);
+
+    String location;
+    try (Dataset dataset =
```
jackye1995: This would actually create the table. I think we need to do `namespace.declareTable` when starting the stage, and then just pass in the location & storage options to be used to eventually do a Dataset.write when the writes are finished.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceStagedTable.java
```diff
@@ -0,0 +1,100 @@
...
+
+  /** The type of staging operation. */
+  public enum Operation {
+    CREATE,
+    REPLACE,
+    CREATE_OR_REPLACE
+  }
+
+  private final Operation operation;
+  private final Runnable commitAction;
```
jackye1995: why pass in Runnable? I don't see the logic really differ in different cases. Even they differ, it's better to just write the actual logic here I think instead of passing a runnable.

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -519,6 +522,191 @@ public void renameTable(Identifier oldIdent, Identifier newIdent)
...
+      DescribeTableRequest describeRequest = new DescribeTableRequest();
+      tableIdList.forEach(describeRequest::addIdItem);
+      DescribeTableResponse describeResponse = namespace.describeTable(describeRequest);
+      initialStorageOptions = describeResponse.getStorageOptions();
+
+      abortAction =
+          () -> {
+            DropTableRequest dropRequest = new DropTableRequest();
+            tableIdList.forEach(dropRequest::addIdItem);
+            namespace.dropTable(dropRequest);
```
jackye1995: prefer to use deregisterTable

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceStagedTable.java
hamersaw: Updated to pass individual fields and commit / abort functionality is executed accordingly.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
```diff
@@ -56,24 +58,43 @@ public class LanceBatchWrite implements BatchWrite {
...
     this.namespaceImpl = namespaceImpl;
     this.namespaceProperties = namespaceProperties;
     this.tableId = tableId;
+    this.commitFlag = commitFlag;
 
-    // Open dataset at start to capture version for commit
-    this.dataset = openDataset();
+    // For new tables (staged creates), create an empty dataset first.
+    // For existing tables, open to capture version for commit.
+    this.dataset = newTable ? createEmptyDataset() : openDataset();
```
hamersaw: `CREATE TABLE` now routes through the `stageCreate` flow so we need to handle scenarios where writing to a new table happen before the table actually exist (since we updated `stageCreate` to just `declareTable` rather than write the dataset to storage).

Issue #255, opened 2026-02-19
Unnecessarily ordering by row address on column backfill writers


hamersaw:
The `AddColumnsBackfillWrite` and `UpdateColumnsBackfillWrite` have order requirements (ie. `_rowaddr`) for the field. This can result an a relatively expensive shuffle for large amounts of data. The only reason for this order is so that the `BatchWriter` can iterate over the rows, have a single fragment writer open at a time, and sequentially iterate through writes.

To bypass this shuffling phase, we could instead update the `BatchWriter` to take a non-ordered list, and during processing bucket data into separate fragment buffers. Then on `commit` we iterate over these buffers and flush to storage in fragments.

We will need to make sure the memory requirements are not significantly impacted with this approach. Since all the data is in-memory, there should be a minimal overhead of maintaining separate buffers for each fragment within the writer. This should be negligible, but should be validated through testing.

PR #259, opened 2026-02-20
perf: remove sort from add and update columns flow

hamersaw:
Currently, we are sorting on `_rowaddr` so that in the BatchWriter we can partition by fragment ID, where we have a single fragment writer open at a time when we sequentially iterate over the (sorted) input data. For very large datasets this sort operation can be expensive (ex. time + memory utilization). This PR removes the sort and instead creates per fragment buffers on the writer that are packed during record ingestion and then flushed on `commit`. This _should_ be faster on large datasets and the initial benchmark below shows that the fragment buffer packing actually has lower memory utilization than the currently required sort operation.

Closes https://github.com/lance-format/lance-spark/issues/255.


jackye1995 (approved):
👍

PR #261, opened 2026-02-21
feat: improve caching

jackye1995:
In general, ensure that:
1. we use the same session across datasets and cache session
2. we fix dataset version to read so each version is cached


hamersaw (approved):
It looks like in Spark generally JVM is scoped to a single application so there shouldn't be concerns about resource contention across tenants within a Session cache. Is that accurate?

jackye1995:
Thanks for the review! Yeah we assume this is basically running in single person mode. Let me add some info in documentation.

jackye1995:
I made the additional improvement that caches are now isolated by catalog name, so that for user that wants multi-tenant setup, they can setup multiple Spark catalogs.

PR #263, opened 2026-02-22
refactor!: remove legacy LanceCatalog

jackye1995:


jackye1995:
> Then we could get rid of the isPathBasedIdentifier

I think we probably have to maintain that anyway, there is always a use case of people want to convenience to query a specific table at a location while all the other tables are organized in a namespace.

But let me do one more refactoring beyond this to be clear about how we handle path-based tables.

PR #271, opened 2026-02-26
refactor: replace newTransactionBuilder with CommitBuilder API

hamersaw:
Migrate all transaction commits from the deprecated SourcedTransaction pattern (dataset.newTransactionBuilder().operation(op).build().commit()) to the new CommitBuilder API from the Lance Java SDK. This uses explicit Transaction + CommitBuilder for clearer separation of concerns.

Notable: StagedCommit.commitNewTable() now uses URI-based CommitBuilder directly instead of the Dataset.write() + newTransactionBuilder workaround.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
@@ -26,7 +26,6 @@ import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.util.LanceArrowUtils;
 
 import java.time.Instant;
-import java.time.ZonedDateTime;
 import java.util.List;
 
 public class Utils {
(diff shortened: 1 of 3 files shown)


lance-spark-3.4_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
```diff
@@ -149,7 +151,11 @@ public void commit(WriterCommitMessage[] messages) {
                 .newFragments(newFragments)
                 .build();
 
-        dataset.newTransactionBuilder().operation(update).build().commit();
```
jackye1995: I think when we have a dataset, we still want to use that to open a transaction? Because that will be able to reuse the dataset that is already being opened and avoid loading the same dataset again even if the metadata file is already cached. Only when we are doing things like CTAS and don't have a dataset yet, we should use this

PR #297, opened 2026-03-13
feat: expose prefilter parameter to support vector search

nyl3532016:
When performing vector queries, if a specific fragment is targeted, prefilter must be set to true to ensure correct execution.
Vector queries are pushed down to shards for execution, but the returned Top-K results consist of K × number of shards records, which require manual processing by the business layer.
#189 
[lance-pr](https://github.com/lance-format/lance/pull/6040)


nyl3532016:
@hamersaw Thanks for your time. Will the lance-version in the main branch be upgraded to the recent 4.0.0-beta.x?

PR #300, opened 2026-03-13
test: try java 21 with spark 3.5

jackye1995:


LuciferYang:
It is not recommended to use Java 21 with Apache Spark 3.5, as it only receives preliminary support. Moreover, the Apache Spark community does not have a daily test combination of branch 3.5 + Java 21.

PR #304, opened 2026-03-16
fix: surfacing failures in async write threads

hamersaw:
Currently, Spark writers start a background thread to perform the actual writes and submit to it it batches (either directly, or through a queue). The problem is that if the background thread fails, the exception is lost and the parent thread is deadlocked waiting for the child to accept the next write.

In this PR we ensure the in all scenarios the child threads expose the exception to the parent thread, this is done by:
- using lock conditions in `SemaphoreArrowBatchWriteBuffer.java` to efficiently notify when batches can be accepted and when errors occur
- using an offer timeout on `QueuedArrowBatchWriteBuffer.java` to attempt to enqueue the next buffer, and check for errors intermittently. 

It also uses a `Future` on the background writing process to capture the exception rather than an `AtomicReference` that needs to be explicitly set. This should mitigate any future issues where we "forget" to set it on some branch.


jackye1995 (approved):
Thanks for the fix!

Issue #217, opened 2026-02-07
Table exists check does not work in Spark 4.0


hamersaw:
While running the new integration tests we found that the command `spark.catalog.tableExists("source")` does not work for the Spark connector in Spark 4.0. The resultant stack trace is:
```
>>> spark.catalog.tableExists("source")
26/02/06 19:55:59 WARN CheckAllocator: More than one DefaultAllocationManager on classpath. Choosing first found
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/opt/spark/python/pyspark/sql/catalog.py", line 745, in tableExists
    return self._jcatalog.tableExists(tableName)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/spark/python/lib/py4j-0.10.9.9-src.zip/py4j/java_gateway.py", line 1362, in __call__
  File "/opt/spark/python/pyspark/errors/exceptions/captured.py", line 282, in deco
    return f(*a, **kw)
           ^^^^^^^^^^^
  File "/opt/spark/python/lib/py4j-0.10.9.9-src.zip/py4j/protocol.py", line 327, in get_return_value
py4j.protocol.Py4JJavaError: An error occurred while calling o49.tableExists.
: java.lang.RuntimeException: Failed to describe table: source
	at org.lance.spark.BaseLanceNamespaceSparkCatalog.loadTable(BaseLanceNamespaceSparkCatalog.java:429)
	at org.apache.spark.sql.connector.catalog.CatalogV2Util$.getTable(CatalogV2Util.scala:438)
	at org.apache.spark.sql.connector.catalog.CatalogV2Util$.loadTable(CatalogV2Util.scala:413)
	at org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveRelations$.$anonfun$lookupTableOrView$2(Analyzer.scala:1234)
	at scala.Option.orElse(Option.scala:477)
	at org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveRelations$.org$apache$spark$sql$catalyst$analysis$Analyzer$ResolveRelations$$lookupTableOrView(Analyzer.scala:1229)
	at org.apache.spark.sql.catalyst.analysis.Analyzer$ResolveRelations$$anonfun$apply$12.applyOrElse(Analyzer.scala:1211)
[...]

PR #326, opened 2026-03-23
fix: handle JNI RuntimeException in describeTable for Spark 4.0+ tableExists

LuciferYang:
Closes https://github.com/lance-format/lance-spark/issues/217

### Problem

In Spark 4.0+, `spark.catalog.tableExists()` internally calls `catalog.loadTable()` (not `catalog.tableExists()` as in Spark 3.x). When a table doesn't exist, `DirectoryNamespace.describeTableNative()` (JNI) throws a raw `RuntimeException` instead of a typed `LanceNamespaceException`. Although `loadTableInternal` caught `RuntimeException`, it re-wrapped it as another `RuntimeException` rather than converting it to `NoSuchTableException`. Spark 4.0's `CatalogV2Util.loadTable()` only catches `NoSuchTableException` to determine table absence, so the re-wrapped `RuntimeException` propagated as a crash.

This does not affect Spark 3.x, where `tableExists()` is called directly and all exceptions are caught.

**Root cause**: `dir.rs` and `dir/manifest.rs` in `lance-namespace-impls` use `Error::namespace_source(String)` instead of `NamespaceError::TableNotFound`. The JNI bridge (`error.rs`) tries to downcast the source to `NamespaceError` to throw a typed `LanceNamespaceException`, but the downcast fails on a plain `String` and falls back to `RuntimeException`. Note: even when the downcast succeeds, the JNI bridge instantiates the base `LanceNamespaceException` class (not the `TableNotFoundException` subclass), so catching `LanceNamespaceException` and checking `ErrorCode` is the correct approach on the Java side.

### Fix

- Added a `describeTableOrThrow()` private helper that wraps `namespace.describeTable()` and translates table-not-found errors into Spark's `NoSuchTableException` via two catch blocks:
  1. `LanceNamespaceException` with `ErrorCode.TABLE_NOT_FOUND` check — forward-compatible with the upstream fix where `NamespaceError::TableNotFound` will be used, enabling the JNI downcast to succeed and produce a `LanceNamespaceException(TABLE_NOT_FOUND)`. This also covers `TableNotFoundException` (a subclass of `LanceNamespaceException`).
  2. `RuntimeException` with message matching — workaround for the current state where the JNI downcast fails and falls back to a raw `RuntimeException`. Two known message patterns are matched:
     - `dir.rs`: "Table does not exist: {table_name}" (single-level / directory listing mode)
     - `dir/manifest.rs`: "Table '{table_name}' not found" (manifest mode, the default)
- Applied the helper at `loadTableInternal` and `stageReplace` — the call sites where `NoSuchTableException` is the expected outcome for missing tables. The other two call sites (`createTable`, `stageCreateOrReplace`) call `namespace.describeTable()` directly since the table is known to exist there.
- Added a TODO in the Javadoc to remove the `RuntimeException` catch block once lance fixes `dir.rs` and `manifest.rs` to use `NamespaceError::TableNotFound`.

## Changes

| File | Change |
|------|--------|
| `BaseLanceNamespaceSparkCatalog.java` | Add `describeTableOrThrow()` helper; catch `LanceNamespaceException` with error code check + `RuntimeException` with message matching; apply at 2 call sites |
| `SparkLanceNamespaceTestBase.java` | Add 2 tests covering the manifest.rs "not found" path |
| `BaseTestSparkDirectorySingleLevelNsNamespace.java` | New test base class with `single_level_ns=true` covering the dir.rs "does not exist" path |
| `TestSparkDirectorySingleLevelNsNamespace.java` (3.4 + 3.5) | Concrete test classes for both Spark versions |

## Test plan

- [x] `testTableExistsReturnsFalseForNonExistentTable` (manifest mode) — verifies the reported bug scenario
- [x] `testLoadNonExistentTableThrowsNoSuchTableException` (manifest mode) — verifies exception type contract
- [x] `testTableExistsReturnsFalseForNonExistentTable` (single-level mode) — verifies dir.rs error path
- [x] `testLoadNonExistentTableThrowsNoSuchTableException` (single-level mode) — verifies dir.rs error translation
- [x] All 22 tests pass on Spark 3.4 (`TestSparkDirectoryNamespace` + `TestSparkDirectorySingleLevelNsNamespace`)
- [x] All 22 tests pass on Spark 3.5
- [x] `make lint` clean

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -16,6 +16,8 @@ package org.lance.spark;
 import org.lance.Dataset;
 import org.lance.WriteParams;
 import org.lance.namespace.LanceNamespace;
+import org.lance.namespace.errors.ErrorCode;
+import org.lance.namespace.errors.LanceNamespaceException;
 import org.lance.namespace.errors.TableNotFoundException;
 import org.lance.namespace.model.DeclareTableRequest;
 import org.lance.namespace.model.DeclareTableResponse;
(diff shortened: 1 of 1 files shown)


LuciferYang:
update the code and will feedback later

PR #328, opened 2026-03-23
fix: respecting managed versioning when calling declare / describe table

hamersaw:
Managed versioning is a REST namespace flag returned from describeTable and declareTable requests. It is used to indicate the REST server should be responsible for managing versions of the table, meaning specifically that manifest updates (ex. writing new rows, etc) should be routed through the implementation.

Currently, the Spark connector does not respect this flag, meaning regardless of setting all manifest updates are routed through the REST namespace implementation. This can cause errors if this functionality is not supported yet.

In this PR, we maintain the managed version flag and then use it to either populate the namespace on CommitBuilder (or not). This is analogous to how this behaves when using the WriteDatasetBuilder or OpenDatasetBuilder, we're just doing this explicitly while using the CommitBuilder directly (because of the distributed nature of this solution).


jackye1995 (approved):
looks good to me!

PR #351, opened 2026-03-27
feat: propagate enable_stable_row_ids through spark write path

ivscheianu:
Propagate `enable_stable_row_ids` through the Spark write path so that clients can enable stable row IDs via the standard Spark API:

```scala
df.write.option("enable_stable_row_ids", "true").save("lance://...")
```

instead of having to drop down to the native Lance Java API
```java
Dataset.write().enableStableRowIds(true)
```

The initial implementation tried to mirror the flag into `Dataset.getConfig()` via an explicit `UpdateConfig` commit after table creation. This was necessary because `Dataset.write().enableStableRowIds(true)` sets the internal Rust manifest flag but does **not** populate `Dataset.getConfig(), so downstream readers had no Spark-accessible way to check whether stable row IDs were enabled.

This approach had two problems:
1. **Unnecessary version bump.** Every table creation with stable row IDs produced two commits instead of on: the initial CREATE, followed by a separate UpdateConfig commit just to persist the flag in the config map. This means the table starts at version 2 instead of version 1.

3. The above mentioned action will also cause **CDF (Change Data Feed) breakage.** CDF tracks row-level changes across dataset versions, and tests assert exact version numbers for created/updated rows (e.g. "row inserted at version 1, updated at version 3"). The extra version bump shifts every subsequent version by +1, breaking all CDF assertions across the test suite. 

The fix:is to set `CommitBuilder.useStableRowIds(true)` directly on the commit operation. This sets the Rust manifest flag in a single atomic commit: no config map, no extra version, no version inflation. The config map workaround is eliminated entirely.

A companion PR on lance-core ([#3594](https://github.com/lance-format/lance/pull/6325)) adds `Dataset.hasStableRowIds()`, which reads the manifest flag directly via JNI, giving downstream readers a clean API to check the flag without relying on the config map.

**NOTE:** he changes are related, but can be merged independently. No synchronization is needed, they can exist without each other without problems

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/write/StagedCommit.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/write/StagedCommit.java
@@ -121,7 +121,9 @@ public class StagedCommit {
     final Overwrite operation = Overwrite.builder().fragments(fragments).schema(schema).build();
     final CommitBuilder builder =
         new CommitBuilder(datasetUri, LanceRuntime.allocator()).writeParams(storageOptions);
-    builder.useStableRowIds(enableStableRowIds);
+    if (enableStableRowIds) {
+      builder.useStableRowIds(true);
+    }
     applyManagedVersioning(builder);
     try (Transaction txn = new Transaction.Builder().operation(operation).build();
         Dataset committed = builder.execute(txn)) {


hamersaw (approved):
Looks great! Thanks for the iteration.

PR #352, opened 2026-03-27
feat: support rename table

wombatu-kun:
Closes https://github.com/lance-format/lance-spark/issues/249

Implements ALTER TABLE ... RENAME TO ... for namespace-based tables by wiring Spark's built-in rename DDL to the existing LanceNamespace.renameTable() API. No custom SQL grammar changes needed — Spark handles the parsing natively.

Main Changes in `lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java`:  
- Replaced `UnsupportedOperationException` stub with real implementation that transforms identifiers, builds RenameTableRequest (source ID + new name + new namespace), and delegates to namespace.renameTable()
- Path-based tables still throw `UnsupportedOperationException` 
- Supports cross-namespace rename via newNamespaceId  

All changes propagate to Spark 3.4–4.1 and Scala 2.12/2.13 via Maven source sharing — no version-specific module changes required.

### DIFF
--- a/docker/tests/conftest.py
+++ b/docker/tests/conftest.py
@@ -17,6 +17,11 @@ import urllib.error
 import pytest
 from pyspark.sql import SparkSession
 
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "requires_rest: test only runs on REST-based backends")
+
+
 # ---------------------------------------------------------------------------
 # Azurite (Azure Blob Storage emulator) configuration
 # ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


wombatu-kun:
> Looks great! Can we add the `ALTER TABLE ... RENAME ...` to the docs? Then I think we're good to merge!
> 
> PS. I like the addition of the optional LanceDB backend to the integration tests. I'm wondering if we should have an env var that controls which backends to run against?

Good idea! Added `TEST_BACKENDS` env var — you can now run `TEST_BACKENDS=local` make docker-test to test against a single backend. Defaults to all enabled backends so existing behavior is unchanged.

hamersaw (approved):
Thanks for rebasing, this looks great!

PR #353, opened 2026-03-27
refactor: remove Java-side dataset cache, rely on Rust-side Session

LuciferYang:
## Summary

- Delete `LanceDatasetCache`; the Rust-side `Session` already caches metadata and index blocks, so a Java-side `Dataset` cache only saves manifest deserialization (microseconds).
- Load fragment metadata on demand via `dataset.getFragment(id)` per partition, instead of eagerly building `Map<Integer, Fragment>` from `dataset.getFragments()` on every cache miss — O(N) in total fragment count, even though each partition reads one fragment.
- Each `LanceFragmentScanner` opens, owns, and closes its own `Dataset` through a new `LanceRuntime.openDataset()` helper.

## Changes

| File | Change |
|------|--------|
| `LanceDatasetCache.java` | Deleted (~291 lines). |
| `LanceRuntime.java` | Add `openDataset()` — wires the catalog `Session`, reconstructs the namespace from the `(namespaceImpl, namespaceProperties)` strings carried in `LanceInputPartition`, merges storage options, and pins the version. |
| `LanceFragmentScanner.java` | `create()` opens its own `Dataset` and loads only the target fragment; `close()` closes both scanner and dataset, using `Throwable.addSuppressed` so the first failure isn't masked by cleanup. |

## Before / after

```
Before (main)                          After
-------------                          -----
LanceFragmentScanner.create()          LanceFragmentScanner.create()
  LanceDatasetCache.getFragment()        LanceRuntime.openDataset()   // per partition
    Guava cache.get()                      (Rust Session: metadata + index)
      new CachedDataset(dataset)           dataset.getFragment(id)    // single fragment
        dataset.getFragments()  ← O(N)
    cached.getFragment(id)
```

The per-catalog Rust `Session` (configured via `LANCE_INDEX_CACHE_SIZE` / `LANCE_METADATA_CACHE_SIZE`) and the global Arrow `BufferAllocator` are unchanged.


hamersaw (commented):
Do you think it would make sense to just remove the `LanceDatasetCache` and house fragment caching in the `LanceRuntime`? It feels like there's just multiple layers of cache here.

hamersaw:
I think overall this looks reasonable, but we need some relatively comprehensive benchmarks to support any performance implications.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/FragmentLoadingBenchmarkTest.java
```diff
@@ -0,0 +1,206 @@
...
+ * Benchmark comparing eager fragment loading ({@code dataset.getFragments()}) vs lazy on-demand
+ * loading ({@code dataset.getFragment(id)}).
+ *
+ * <p>This benchmark supports the performance claim in PR #353: removing eager fragment pre-loading
+ * reduces per-partition initialization cost from O(N) to O(1) where N is the total fragment count.
+ *
+ * <p>Tagged with "benchmark" so it is excluded from normal test runs.
+ */
+@Tag("benchmark")
+public class FragmentLoadingBenchmarkTest {
```
LuciferYang: Is this microbenchmark sufficient to demonstrate the effect? We can run

```
mvn test -pl lance-spark-base_2.12 \
  -Dtest=FragmentLoadingBenchmarkTest \
  -Dtest.excludedGroups= \
  -Dgroups=benchmark
```

to verify:

```
=== Fragment Loading Benchmark ===
Fragments    |  getFragments() (ms) | getFragment(id) (ms) |    Speedup
----------------------------------------------------------------------
10           |             0.082 ms |             0.007 ms |      11.0x
50           |             0.329 ms |             0.008 ms |      43.2x
100          |             0.630 ms |             0.010 ms |      65.9x
500          |             2.929 ms |             0.008 ms |     380.9x
1000         |             6.064 ms |             0.008 ms |     760.4x
```

Notes:
- getFragments(): loads ALL fragment metadata (old eager approach)
- getFragment(id): loads ONE fragment by ID (new lazy approach)
- Each worker partition only needs one fragment, so the lazy approach avoids
  loading metadata for all other fragments in the dataset.

The project does not integrate JMH yet, so this benchmark is relatively simple.

LuciferYang: If end-to-end impact is needed, we are running tests on a 1TB TPC-DS dataset, but this improvement likely won’t show a noticeable difference there.

LuciferYang: @summaryzb Do you have any test results about this ?

hamersaw:
Ok, so diving a little deeper. The rust-side implementation has it's own caches; for metadata (LANCE_METADATA_CACHE_SIZE) and indexes (LANCE_INDEX_CACHE_SIZE). By caching the actual dataset handle, we're really only saving time in deserializing the dataset manifest bytes, which should be measureable in microseconds (or milliseconds at worst). I'm wondering if we should just remove the notion of Spark-side caches for everything (dataset + fragment) and rely on the rust-side Lance cache. This is basically what the `LanceRuntime ` handles, rust-size Lance cache.

LuciferYang:
That makes sense. Since the Session already caches metadata and index blocks on the Rust side, the Java-side `LoadingCache<DatasetCacheKey, Dataset>` only saves manifest deserialization time. I agree we should remove it and let each scanner open/close its own `Dataset` handle, relying on the Rust-side cache for the heavy lifting. I'll update the pr to:

1. Remove the Guava `LoadingCache`, `DatasetCacheKey`, `getCachedDataset()`, and `getFragment()`
2. Have `LanceFragmentScanner` open its own `Dataset` (via `openDataset()`) and close it in `close()`
3. Keep the per-catalog `Session` caching as-is

I'll also enhance `openDataset()` to honor per-query `blockSize`/`indexCacheSize`/`metadataCacheSize` settings in the same change — the cache path currently silently ignores these, and since we're already rewiring the open path, this is the natural place to fix it.

Do you think this is reasonable?

hamersaw:
> I'll also enhance `openDataset()` to honor per-query `blockSize`/`indexCacheSize`/`metadataCacheSize` settings in the same change — the cache path currently silently ignores these, and since we're already rewiring the open path, this is the natural place to fix it.
> 
> Do you think this is reasonable?

I'm not following exactly where these are going to be inserted. This gets a little bit tricky because these a global session configuration options and may not be reasonably applied to a single dataset? Maybe it's worth hacking together a proposal and we can iterate?

LuciferYang:
@hamersaw I have refactored the code, removed `LanceDatasetCache`, and updated the PR description. Can we do further iterations based on this version?

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentScanner.java
```diff
@@ -30,32 +32,46 @@
...
   public static LanceFragmentScanner create(int fragmentId, LanceInputPartition inputPartition) {
+    Dataset dataset = null;
     try {
       LanceSparkReadOptions readOptions = inputPartition.getReadOptions();
-      Fragment fragment =
-          LanceDatasetCache.getFragment(
-              readOptions,
-              fragmentId,
+      dataset =
+          LanceRuntime.openDataset(
```
LuciferYang: Let me check if we can use the utility methods from `Utils` here.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentScanner.java
```diff
@@ -30,32 +32,40 @@
...
     try {
       LanceSparkReadOptions readOptions = inputPartition.getReadOptions();
-      Fragment fragment =
-          LanceDatasetCache.getFragment(
-              readOptions,
-              fragmentId,
-              inputPartition.getInitialStorageOptions(),
-              inputPartition.getNamespaceImpl(),
-              inputPartition.getNamespaceProperties());
+      dataset =
```
LuciferYang, line 64: @hamersaw On your earlier concern about blockSize/indexCacheSize/metadataCacheSize: this pr doesn't add new API for them now, but switching to `Utils.openDatasetBuilder` does mean the fragment-scan path now passes these through (previously they were dropped by `LanceDatasetCache`). This matches what `LanceCountStarPartitionReader` and ~27 other call sites already do.  WDYT?

hamersaw (approved):
Thanks, I think this looks great! I added the namespace build back in so that we can ensure credential refresh on long-running operations.

LuciferYang:
Thank you @hamersaw

LuciferYang:
@hamersaw — following up on this. Removing the Java-side dataset cache to rely on Rust-side `Session` reuse exposed a thundering-herd issue in `ObjectStoreRegistry::get_store`: concurrent default-opens against the same URI under a shared `Session` all fall through to fresh `ObjectStore` builds because the registry's `Weak<Arc<ObjectStore>>` dedup is post-build.

Repro on a 34 GiB MinIO TPC-DS `store_sales` workload (Spark `local-cluster[4,3,4096]`, 234 Lance fragments), three full-scan SQL queries via `spark.read.format("lance")`:

- **q1** — `SELECT sum(ss_quantity) FROM store_sales` (single int32 column, full scan)
- **q2** — `SELECT sum(ss_item_sk), sum(ss_ext_sales_price), sum(ss_quantity) FROM store_sales` (multi-column incl. decimal128, full scan)
- **q3** — `SELECT sum(ss_net_profit) FROM store_sales WHERE ss_sold_date_sk BETWEEN <a> AND <b>` (date-range predicate; Lance has no row-group stats so still scans)

Without the fix, every JMH iteration pays the full cold open cost (q1-lance 12.7 s ± 146 ms, CV ≈ 1% — i.e. all 10 iterations identical because each one rebuilds the credential chain + HTTP client).

Fix is in lance-format/lance#6839: per-`CacheKey` build-lock in `get_store` + a process-wide `LazyLock<Arc<ObjectStoreRegistry>>` for the JNI default-open path. After the fix:

| query | baseline | fixed | speedup |
|---|---|---|---|
| q1 sum(ss_quantity)            | 12.7 s | 2.5 s | 5.0x |
| q2 multi-col incl. decimal128  | 7.3 s  | 4.3 s | 1.7x |
| q3 date-range filter           | 12.9 s | 2.9 s | 4.5x |

Standalone `concurrent_open_bench` (no Spark, 144 concurrent opens under one `Session`) drops 9.50 s -> 0.18 s (53x).

PR #355, opened 2026-03-27
feat: preserve blob data through Spark shuffle during JOIN + INSERT INTO

beinan:
## Summary

- When blob columns flow through Spark's shuffle (e.g., `INSERT INTO target SELECT ... FROM source_a JOIN source_b`), the actual blob data was previously **lost** because `getBinary()` returned empty byte arrays (`new byte[0]`). The target table ended up with zero-length blobs.
- This PR introduces a **blob reference mechanism** that preserves blob data through Spark's shuffle without materializing the full blob bytes (which could be MBs/GBs per row and would kill shuffle performance).

### How it works

**Read side** — compact blob references flow through shuffle:
- Instead of returning `byte[0]`, blob columns now serialize compact **~100-byte `BlobReference` descriptors** containing the source dataset URI, column name, and row address
- The scanner automatically requests `_rowaddr` when blob columns are present, and strips it from the output after extracting the row addresses
- These references are small enough to flow through Spark shuffle without overhead

**Write side** — blob references are resolved to actual bytes:
- A new `BlobResolvingLargeBinaryWriter` detects `BlobReference` magic headers in incoming binary values
- It opens the source dataset (with caching), calls `Dataset.takeBlobs()` to fetch the actual blob bytes
- The real blob data is then written to the target table through the normal blob write path

### Key files

| File | Status | Description |
|------|--------|-------------|
| `BlobReference.java` | **NEW** | Compact serializable blob reference format (`LBRF` magic + dataset URI + column name + row address) |
| `BlobReferenceResolver.java` | **NEW** | Write-side resolver: opens source dataset → `takeBlobs()` → returns actual bytes |
| `BaseBlobJoinTest.java` | **NEW** | Tests for blob preservation through simple INSERT INTO SELECT, JOIN + INSERT, and mixed blob/non-blob JOINs |
| `BlobJoinTest.java` (3.4 + 3.5) | **NEW** | Concrete test implementations |
| `BlobStructAccessor.java` | Modified | Added blob reference context (dataset URI, column name, row addresses) and `getBlobReference()` |
| `LanceArrowColumnVector.java` | Modified | `getBinary()` returns blob reference instead of `byte[0]` |
| `LanceFragmentScanner.java` | Modified | Detects blob columns, requests `_rowaddr`, exposes dataset URI and blob column names |
| `LanceFragmentColumnarBatchScanner.java` | Modified | Extracts row addresses from `_rowaddr`, sets blob context on `BlobStructAccessor`, strips implicit `_rowaddr` |
| `LanceArrowWriter.scala` | Modified | Added `BlobResolvingLargeBinaryWriter` for blob-encoded columns |
| `BaseBlobCreateTableTest.java` | Modified | Updated assertions for new blob reference behavior |

## Test plan

- [x] `BlobJoinTest.testBlobPreservedDuringInsertIntoSelect` — Simple `INSERT INTO target SELECT FROM source` preserves blob data
- [x] `BlobJoinTest.testBlobPreservedDuringJoinAndInsert` — `JOIN` of two blob tables preserves both blob columns in target
- [x] `BlobJoinTest.testNonBlobColumnsPreservedDuringJoinWithBlobs` — Non-blob columns survive JOIN+INSERT alongside blob columns
- [x] All existing `BlobCreateTableTest` tests pass (9/9)

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
@@ -19,6 +19,7 @@ import org.lance.spark.vectorized.BlobStructAccessor;
 import org.lance.spark.vectorized.LanceArrowColumnVector;
 
 import org.apache.arrow.vector.FieldVector;
+import org.apache.arrow.vector.UInt8Vector;
 import org.apache.arrow.vector.VectorSchemaRoot;
 import org.apache.arrow.vector.complex.StructVector;
 import org.apache.arrow.vector.ipc.ArrowReader;
@@ -36,6 +37,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 
 public class LanceFragmentColumnarBatchScanner implements AutoCloseable {
   private final LanceFragmentScanner fragmentScanner;
@@ -117,6 +119,10 @@ public class LanceFragmentColumnarBatchScanner implements AutoCloseable {
       actualFields.put(rootVectors.get(i).getField().getName(), rootVectors.get(i));
     }
 
+    // Extract row addresses for blob reference support
+    Set<String> blobColumnNames = fragmentScanner.getBlobColumnNames();
+    long[] rowAddresses = extractRowAddresses(rootVectors, blobColumnNames, root.getRowCount());
+
     List<ColumnVector> fieldVectors = new ArrayList<>(schema.size());
     StructField[] fields = schema.fields();
     for (StructField field : fields) {
@@ -150,10 +156,44 @@ public class LanceFragmentColumnarBatchScanner implements AutoCloseable {
           throw new IllegalStateException(
               "Lance scan did not return expected field '" + fieldName + "'");
         }
-        fieldVectors.add(new LanceArrowColumnVector(vector));
+        LanceArrowColumnVector colVec = new LanceArrowColumnVector(vector);
+
+        // Set blob reference context so getBinary() produces blob references
+        if (rowAddresses != null && blobColumnNames.contains(fieldName)) {
+          BlobStructAccessor blobAccessor = colVec.getBlobStructAccessor();
+          if (blobAccessor != null) {
+            blobAccessor.setBlobReferenceContext(
+                fragmentScanner.getDatasetUri(), fieldName, rowAddresses);
+          }
+        }
+
+        fieldVectors.add(colVec);
       }
     }
     return fieldVectors;
+  }
+
+  /**
+   * Extracts row addresses from the {@code _rowaddr} column appended by the native scanner. Row
+   * addresses are needed to construct blob references that allow the write side to fetch actual
+   * blob bytes from the source dataset.
+   */
+  private long[] extractRowAddresses(
+      List<FieldVector> rootVectors, Set<String> blobColumnNames, int rowCount) {
+    if (blobColumnNames.isEmpty()) {
+      return null;
+    }
+    for (FieldVector fv : rootVectors) {
+      if (LanceConstant.ROW_ADDRESS.equals(fv.getField().getName()) && fv instanceof UInt8Vector) {
+        UInt8Vector rowAddrVector = (UInt8Vector) fv;
+        long[] rowAddresses = new long[rowCount];
+        for (int i = 0; i < rowCount; i++) {
+          rowAddresses[i] = rowAddrVector.get(i);
+        }
+        return rowAddresses;
+      }
+    }
+    return null;
   }
 
   // Virtual column vector for blob position
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReference.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReference.java
@@ -30,6 +30,12 @@ import java.util.Arrays;
  * the source dataset, fetches the actual blob bytes via {@code Dataset.takeBlobs()}, and writes
  * them to the target table.
  *
+ * <p>The trailing {@code size} is the resolved blob's byte length, read straight from the source
+ * blob descriptor's size vector. It carries no addressing information — resolution only needs the
+ * dataset/column/rowAddress — but it lets the write side budget the (potentially huge) resolved
+ * bytes against {@code maxBatchBytes} while the value is still just a ~200-byte reference, so a
+ * batch flushes before materialization OOMs the executor (see {@code LargeBinaryWriter}).
+ *
  * <p>Wire format:
  *
  * <pre>
@@ -38,6 +44,7 @@ import java.util.Arrays;
  *   [2+N bytes] datasetUri (length-prefixed UTF-8)
  *   [2+N bytes] columnName (length-prefixed UTF-8)
  *   [8 bytes] rowAddress
+ *   [8 bytes] size (resolved blob byte length)
  * </pre>
  */
 public class BlobReference {
@@ -45,19 +52,26 @@ public class BlobReference {
   /** 8-byte magic header to identify a serialized BlobReference. */
   public static final byte[] MAGIC = {'L', 'A', 'N', 'C', 'E', 'R', 'E', 'F'};
 
-  /** Min byte length: magic(8) + version(1) + two empty strings(2+2) + rowAddress(8). */
-  private static final int MIN_SIZE = MAGIC.length + 1 + 2 + 2 + 8;
+  /** Min byte length: magic(8) + version(1) + two empty strings(2+2) + rowAddress(8) + size(8). */
+  private static final int MIN_SIZE = MAGIC.length + 1 + 2 + 2 + 8 + 8;
 
-  private static final byte VERSION = 1;
+  private static final byte VERSION = 2;
 
   private final String datasetUri;
   private final String columnName;
   private final long rowAddress;
+  private final long size;
 
+  /** Constructs a reference with an unknown resolved size ({@code 0}); see {@link #getSize()}. */
   public BlobReference(String datasetUri, String columnName, long rowAddress) {
+    this(datasetUri, columnName, rowAddress, 0L);
+  }
+
+  public BlobReference(String datasetUri, String columnName, long rowAddress, long size) {
     this.datasetUri = datasetUri;
     this.columnName = columnName;
     this.rowAddress = rowAddress;
+    this.size = size;
   }
 
   /**
@@ -90,7 +104,7 @@ public class BlobReference {
       if (colLen < 0 || colLen > remaining) {
         return false;
       }
-      int expectedRemaining = colLen + 8;
+      int expectedRemaining = colLen + 8 + 8; // columnName + rowAddress + size
       return remaining == expectedRemaining;
     } catch (IOException e) {
       return false;
@@ -99,16 +113,17 @@ public class BlobReference {
 
   /** Serialize this reference to a compact byte array. */
   public byte[] serialize() {
-    return appendRowAddress(serializePrefix(datasetUri, columnName), rowAddress);
+    return appendRowAddressAndSize(serializePrefix(datasetUri, columnName), rowAddress, size);
   }
 
   /**
-   * Serializes the constant portion of a reference: everything except the trailing 8-byte
-   * rowAddress (i.e. magic + version + datasetUri + columnName).
+   * Serializes the constant portion of a reference: everything except the trailing per-row
+   * rowAddress and size (i.e. magic + version + datasetUri + columnName).
    *
    * <p>{@code datasetUri} and {@code columnName} are constant for an entire scan batch, so callers
    * on the per-row read hot path should compute this prefix once and then call {@link
-   * #appendRowAddress(byte[], long)} per row instead of re-encoding the strings every time.
+   * #appendRowAddressAndSize(byte[], long, long)} per row instead of re-encoding the strings every
+   * time.
    */
   public static byte[] serializePrefix(String datasetUri, String columnName) {
     try {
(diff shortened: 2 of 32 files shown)


hamersaw (commented):
Thanks for the PR! Should be a pretty significant help in reducing memory utilization. A few comments to iron things out, but I think I need to dive through the regular read paths to make sure this doesn't negatively affect performance there as well.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
```diff
@@ -62,29 +64,90 @@ public boolean loadNextBatch() throws IOException {
 
     if (hasNext) {
       VectorSchemaRoot root = arrowReader.getVectorSchemaRoot();
+      List<FieldVector> rootVectors = root.getFieldVectors();
+      int rowCount = root.getRowCount();
+
+      Set<String> blobColumnNames = new java.util.HashSet<>(fragmentScanner.getBlobColumnNames());
```
hamersaw: I think rather than computing a new HashSet here for every single batch, why not just store the blob column names in the fragmentScanner as a HashSet?

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
```diff
@@ -62,29 +64,90 @@ public boolean loadNextBatch() throws IOException {
...
+    if (LanceConstant.ROW_ADDRESS.equals(lastVector.getField().getName())
+        && lastVector instanceof UInt8Vector) {
+      UInt8Vector rowAddrVector = (UInt8Vector) lastVector;
+      long[] rowAddresses = new long[rowCount];
+      for (int i = 0; i < rowCount; i++) {
+        rowAddresses[i] = rowAddrVector.getObjectNoOverflow(i).longValue();
+      }
+      return rowAddresses;
+    }
+    return null;
```
hamersaw: This only parses the last vector for `_rowaddr`, depending on the projection and requested metadata columns that column could be anywhere in the list.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReference.java
```diff
@@ -0,0 +1,146 @@
...
+        throw new IllegalArgumentException("Unsupported BlobReference version: " + version);
+      }
+      String datasetUri = readString(in);
+      String columnName = readString(in);
+      long rowAddress = in.readLong();
+      return new BlobReference(datasetUri, columnName, rowAddress);
+    } catch (IOException e) {
+      throw new RuntimeException("Failed to deserialize BlobReference", e);
+    }
+  }
```
hamersaw, line 183: We should probably add a unit test for round-trip serialization / deserialization here just to shield against future regressions.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -0,0 +1,110 @@
...
+    List<Long> rowAddresses = new ArrayList<>(1);
+    rowAddresses.add(ref.getRowAddress());
+    List<BlobFile> blobs = dataset.takeBlobs(rowAddresses, ref.getColumnName());
+    if (blobs.isEmpty()) {
+      return new byte[0];
+    }
+    try (BlobFile blob = blobs.get(0)) {
+      return blob.read();
+    }
+  }
```
hamersaw, line 89: This calls `takeBlobs` for every single input row I believe. This will be a dominate cost as noted in the comment. I think we should try to batch these somehow.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -482,3 +493,32 @@ private[arrow] class IntervalMonthDayNanoWriter(val valueVector: IntervalMonthDa
...
+ * actual blob content. This writer detects blob references, opens the source dataset, fetches
+ * the real blob bytes via takeBlobs(), and writes them to the target Arrow vector.
+ *
+ * For non-blob-reference binary values (e.g. normal user data), the bytes are written as-is.
+ */
+private[arrow] class BlobResolvingLargeBinaryWriter(val valueVector: LargeVarBinaryVector)
+  extends LanceArrowFieldWriter {
+
+  // Lazy-initialized resolver (shared across all rows in this writer's lifetime)
+  @transient private lazy val resolver = new org.lance.spark.utils.BlobReferenceResolver()
```
hamersaw: I don't think this is ever closed, for long running Spark jobs with a ton of data transfer this could be problematic from a resource perspective.

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseBlobJoinTest.java
```diff
@@ -0,0 +1,399 @@
...
+
+    // Row 1: blob size should match the original content
+    long blobSize1 = targetRows.get(0).getLong(2);
+    assertEquals(
+        blobContent1.length, blobSize1, "Blob size should match the original content length");
+
+    // Row 2: blob size should match the original content
+    long blobSize2 = targetRows.get(1).getLong(2);
+    assertEquals(
+        blobContent2.length, blobSize2, "Blob size should match the original content length");
```
hamersaw, line 150: Since we have the blob content ^^ we might as well use that to compare rather than just checking the byte length.

hamersaw:
Diving through this a bit more. I'm not sure there's a reason to have a `LargeBinaryWriter` (existing) and a `BlobResolvingLargeBinaryWriter` (new). Right now, if we're reading from a blob column and writing to a regular binary we're going to dump a ton of `BlobReferences` rather than the actual bytes because it uses the `LargeBinaryWriter` which doesn't resolve the `BlobReference`. I think we should collapse these into a single writer to mitigate future issues.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
beinan: Fixed — `LanceFragmentScanner` now stores blob column names as `Set<String>` directly, so `loadNextBatch()` just calls `fragmentScanner.getBlobColumnNames()` with no allocation per batch.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
beinan: Good catch — `extractRowAddresses` now iterates all vectors and searches by field name instead of assuming `_rowaddr` is at the last position.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReference.java
beinan: Added BlobReferenceTest with tests for round-trip serialization, unicode URIs, empty strings, invalid input rejection, and magic header verification.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
beinan: Implemented batching — BlobReferenceResolver now groups references by (datasetUri, columnName) and flushes them in batches of 64, calling takeBlobs() once per batch instead of once per row.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
beinan: Fixed — the resolver is now properly managed via the LargeBinaryWriter lifecycle: flush() is called in finish() and close() is called in reset(), so cached datasets are cleaned up between batches and at the end of the write.

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseBlobJoinTest.java
beinan: Updated — all three tests now resolve the blob references from the target table and use assertArrayEquals to compare actual blob content against the original bytes.

beinan:
Good point — collapsed BlobResolvingLargeBinaryWriter into LargeBinaryWriter so there is now a single writer that always resolves blob references via resolveIfNeeded(). This means blob references are correctly resolved regardless of whether the target column is blob-encoded or regular binary, preventing silent data corruption.

beinan:
@hamersaw All feedback has been addressed and CI is fully green. Ready for another look when you get a chance!

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
```diff
@@ -62,29 +64,90 @@ public boolean loadNextBatch() throws IOException {
...
+    if (blobColumnNames.isEmpty()) {
+      return null;
+    }
+    for (FieldVector fv : rootVectors) {
+      if (LanceConstant.ROW_ADDRESS.equals(fv.getField().getName()) && fv instanceof UInt8Vector) {
+        UInt8Vector rowAddrVector = (UInt8Vector) fv;
+        long[] rowAddresses = new long[rowCount];
+        for (int i = 0; i < rowCount; i++) {
+          rowAddresses[i] = rowAddrVector.getObjectNoOverflow(i).longValue();
+        }
```
hamersaw: Per-row BigInteger allocation extracting row addresses: extractRowAddresses() runs once per Arrow batch and loops over every row calling rowAddrVector.getObjectNoOverflow(i), which returns a freshly allocated BigInteger that is immediately reduced to a long. For a multi-million-row scan with a blob column this is millions of throwaway BigInteger allocations purely to read a 64-bit value, adding avoidable GC pressure on the scan hot path.

**Fix:** Use UInt8Vector.get(i), which returns a primitive long directly, instead of getObjectNoOverflow(i).longValue().

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -0,0 +1,170 @@
...
+    if (group.size() >= batchSize) {
+      flushGroup(groupKey, group);
+    }
+
+    if (pending.resolved == null) {
+      flushGroup(groupKey, pendingBlobs.remove(groupKey));
+    }
+
+    return pending.resolved;
+  }
```
hamersaw: Blob resolve batching is defeated — one takeBlobs() round-trip per row: resolve() adds the pending blob to its group, and only batch-flushes when group.size() >= batchSize (64). But the very next statement `if (pending.resolved == null)` flushes immediately whenever the batch threshold was NOT hit — which is every call in steady state, because flushGroup() also removes the group from pendingBlobs so it never accumulates. resolveIfNeeded()/resolve() must return bytes synchronously to LargeBinaryWriter.setValue(), so deferral is impossible by construction. Net effect: dataset.takeBlobs() is invoked with exactly one row address per blob row instead of once per 64. On a JOIN+INSERT INTO carrying millions of blob cells through shuffle, this is ~64x the JNI crossings and object-store read round-trips the batching was designed to eliminate, making the feature impractical at scale.

**Fix:** Make resolution genuinely deferred: have getBinary()/the writer accumulate references and resolve them in true batches (e.g. a two-pass write, or buffer references and back-fill the vector after a single batched takeBlobs per group), rather than resolving synchronously per row.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -314,10 +313,33 @@ private[arrow] class BinaryWriter(val valueVector: VarBinaryVector) extends Lanc
...
   override def setNull(): Unit = {}
   override def setValue(input: SpecializedGetters, ordinal: Int): Unit = {
     val bytes = input.getBinary(ordinal)
-    valueVector.setSafe(count, bytes)
+    if (bytes == null || bytes.length == 0) {
+      valueVector.setSafe(count, bytes)
+    } else {
+      val resolved = resolver.resolveIfNeeded(bytes)
+      valueVector.setSafe(count, resolved)
+    }
```
hamersaw: Every non-empty large-binary value written to a blob column is now passed through resolver.resolveIfNeeded(), which calls BlobReference.isBlobReference() — a 4-byte magic-header ('L','B','R','F') sniff plus a >=17 byte length check. Legitimate user-supplied blob bytes whose first four bytes happen to be 0x4C424652 and whose length is >=17 (e.g. an arbitrary file/binary payload) will be misclassified as a serialized BlobReference. deserialize() then either throws IllegalArgumentException (unknown version byte) — failing the INSERT with a RuntimeException — or, on a coincidental version match, parses bogus length-prefixed strings and a garbage dataset URI, causing a takeBlobs() failure or silently wrong data being written. This affects first-time user binary data on a plain INSERT into a blob column, not reference round-trips.

**Fix:** Disambiguate references from real payloads instead of content sniffing — e.g. carry an out-of-band flag/column indicating the value is a reference, or make the magic long/structured enough (and validate full structure) that collision with real data is negligible, and treat a failed parse as 'not a reference' rather than throwing.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -314,10 +313,33 @@ private[arrow] class BinaryWriter(val valueVector: VarBinaryVector) extends Lanc
...
+      resolver.flush()
+    } catch {
+      case e: java.io.IOException =>
+        throw new RuntimeException("Failed to flush blob resolver", e)
+    }
+  }
+
+  override def reset(): Unit = {
+    super.reset()
+    resolver.close()
```
hamersaw: reset() closes resolver, discarding the source-dataset cache: LargeBinaryWriter.reset() calls resolver.close(), which closes and clears BlobReferenceResolver.datasetCache — the cache whose stated purpose is 'avoid re-opening the same dataset for every row'. Arrow field writers are reset between batches, so the source Lance dataset (manifest + metadata I/O, JNI open) is re-opened on the first blob of every subsequent batch instead of once per task. Combined with the broken batching above, the write path repeatedly re-opens the source dataset and issues per-row blob fetches.

**Fix:** Release the resolver (and its dataset cache) once at writer finish/close, not on every reset(); reset() should only flush pending state, keeping cached datasets open for the task lifetime.

hamersaw (commented):
Thanks for the iteration here, I think we're getting close!

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceFragmentColumnarBatchScanner.java
beinan: Fixed — now uses UInt8Vector.get(i) which returns a primitive long directly, avoiding per-row BigInteger allocation.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
beinan: Completely reworked — the writer now does genuine two-pass batched resolution. setValue() buffers blob references with empty placeholders in the vector. finish() groups all pending references by (datasetUri, columnName) and calls takeBlobs() once per group, then back-fills the vector with resolved bytes. This reduces JNI+I/O from O(N) to O(groups).

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
beinan: Addressed in two ways: (1) Extended magic from 4 bytes (LBRF) to 8 bytes (LANCEREF), making random collision probability ~1 in 2^64. (2) isBlobReference() now does full structural validation — checks version byte, verifies both length-prefixed strings fit within the total byte count, and confirms the remaining bytes exactly equal colLen + 8 (the row address). Random binary data that happens to start with LANCEREF will fail the structural check.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
beinan: Fixed — reset() now only clears the pending blob reference buffer. The resolver (and its dataset cache) lives for the lifetime of the writer and is never closed/recreated between batches. Datasets opened on the first batch stay cached for all subsequent batches in the same task.

beinan:
@hamersaw Addressed all four comments from the second review — CI is green again. Would appreciate another look when you have a moment!

hamersaw (approved):
Looks great! Lets resolve merge conflicts and then I'll get it in.

hamersaw:
@beinan pushed a few commits here that:
- pass a "resolver" to the lance writer that contains a dataset URI with a set of storage options / etc that get passed to the writer. the `LargeBlobWriter` uses this to open the read dataset correctly (ie. respecting vendored credentials, etc)
- buffering take operations on writes so we don't do a single call for each row. this respects the configuration for max buffer bytes on writes.
- various other allocation removals / cleanups to help combat some of the many OOM issues you're running into

beinan:
@hamersaw Thanks for the approval! Could you merge this when you get a chance?

PR #359, opened 2026-03-28
feat: support alter table set unenforced primary key

wombatu-kun:
Closes https://github.com/lance-format/lance-spark/issues/248  

Added `ALTER TABLE ... SET UNENFORCED PRIMARY KEY (col1, col2)` SQL command.  

The command sets `lance-schema:unenforced-primary-key=true` and `lance-schema:unenforced-primary-key:position=N` as field metadata via `UpdateConfig + Transaction + CommitBuilder`.  
Validations: columns must exist, be non-nullable, be leaf fields, and no PK must already be set.

### DIFF
--- a/docs/src/operations/ddl/create-table.md
+++ b/docs/src/operations/ddl/create-table.md
@@ -93,7 +93,7 @@ Lance supports vector (embedding) columns for AI workloads. These columns are st
 
 ### Supported Types
 
-- **Element Types**: `FLOAT` (float32), `DOUBLE` (float64)
+- **Element Types**: `FLOAT` (float32), `DOUBLE` (float64), `FLOAT` with float16 flag (half-precision)
 - **Requirements**:
     - Vectors must be non-nullable
     - All vectors in a column must have the same dimension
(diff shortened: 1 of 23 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/SetUnenforcedPrimaryKeyExec.scala
```diff
@@ -0,0 +1,144 @@
...
+import scala.collection.JavaConverters._
+
+case class SetUnenforcedPrimaryKeyExec(
+    catalog: TableCatalog,
+    ident: Identifier,
+    columns: Seq[String]) extends LeafV2CommandExec {
+
+  override def output: Seq[Attribute] = SetUnenforcedPrimaryKeyOutputType.SCHEMA
+
+  override protected def run(): Seq[InternalRow] = {
```
hamersaw: Do we want to add validation for empty column list with something like `require(columns.nonEmpty)`?

PR #380, opened 2026-04-02
fix: missing metadata issues when converting between ArrowField and SparkField

Ppei-Wang:
fix bugs:
1. toArrowField: fix the issue of missing meta(comment) inside the struct data type.
2. toJsonArrowField: fix the issue of missing metadata in field.
3. toJsonArrowDataType: fix  type compatibility.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/SchemaConverter.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/SchemaConverter.java
@@ -405,9 +405,20 @@ public class SchemaConverter {
       dataType.setType("timestamp");
     } else if (sparkType instanceof DecimalType) {
       DecimalType decimalType = (DecimalType) sparkType;
-      dataType.setType("decimal128");
-      // This is the convention set in lance-namespace/src/schema.rs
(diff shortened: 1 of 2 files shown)


hamersaw (commented):
Thanks for the PR! Just the doc + one quick question. I think this fixes the metadata propagation for `StructType` children, but it doesn't for `MapType` if I'm understanding this correctly. Could we add that fix as well?

hamersaw (approved):
@Ppei-Wang I updated this to choose the smallest decimal type that supports the provided precision. If you think this is reasonable we can go ahead and merge. Thanks for the fix!

Ppei-Wang:
> @Ppei-Wang I updated this to choose the smallest decimal type that supports the provided precision. If you think this is reasonable we can go ahead and merge. Thanks for the fix!

I use `toJsonArrowSchema` for type conversion on our platforms because it uses the `lance namespace`. 

No problem, it's indeed dead code in this project only.

So, only keep the modifications to `LanceArrowUtils`; it works fine for me.

hamersaw:
Closing this as the code has been removed.

Ppei-Wang:
> Closing this as the code has been removed.

ok

PR #396, opened 2026-04-07
feat: add zonemap-based fragment pruning and storage-partitioned join (SPJ) support

beinan:
## Summary
- **Zonemap-based fragment pruning**: Loads per-column min/max/null_count statistics from zonemap indexes during scan planning. Fragments whose zones all fail the pushed filters are skipped entirely — analogous to partition pruning in Hive/Iceberg.
- **Storage-partitioned join (SPJ) support**: Detects partition-compatible columns from zonemap stats (where every fragment has a single distinct value, i.e., `min == max`). When detected, reports the data column as the partition key via `KeyGroupedPartitioning`, and implements `HasPartitionKey` on `LanceInputPartition` to return per-fragment partition values. This enables Spark to perform shuffle-free joins — both between Lance tables and cross-source with Iceberg/Delta tables that share the same partition column.
- **`ZonemapFragmentPruner`**: Analyzes pushed Spark filters (EqualTo, LessThan, GreaterThan, In, IsNull, IsNotNull, And, Or, Not) against zonemap statistics to determine which fragments can be safely eliminated.
- **`ZonemapPartitionDetector`**: Scans zonemap stats across all schema columns (including join keys) to find columns where data is physically grouped by value. Converts partition values to Spark `InternalRow` for the SPJ protocol.
- **CREATE INDEX fix for Spark 4.0+** (#400): Spark 4.0+ natively parses `CREATE INDEX`/`DROP INDEX` into its own logical plans instead of throwing `ParseException`. This caused an NPE in `QueryExecution.isLazyAnalysis`. Fixed by intercepting Spark's native plans and converting them to Lance's `AddIndex`/`LanceDropIndex`.
- **`outputPartitioning()` eagerness fix**: `V2ScanPartitioningAndOrdering` calls `outputPartitioning()` during logical plan optimization, *before* `planInputPartitions()` sets `numPartitions`. Fixed by using `partitionInfo.getFragmentPartitionValues().size()` when partition info is available, so SPJ is recognized early enough for shuffle elimination.

### How SPJ works
When Lance data is written ordered by a column (e.g., `OPTIMIZE TABLE ORDER BY region`), each fragment contains a single value for that column. The zonemap captures this (`min == max`). At scan time:
1. `LanceScanBuilder` loads zonemap stats for schema columns and detects partition-compatible columns
2. `LanceScan.outputPartitioning()` reports `KeyGroupedPartitioning(FieldReference("region"))`
3. Each `LanceInputPartition.partitionKey()` returns the partition value for its fragment
4. Spark's `EnsureRequirements` matches partition keys across both sides of a join → no shuffle

Cross-source joins work because `FieldReference("region")` and Iceberg's `IdentityTransform(FieldReference("region"))` resolve to the same Catalyst `AttributeReference`.

### Benchmark results

End-to-end benchmark on a Spark 4.0 cluster (Kubernetes, 2–5 executors with 4 cores / 16 GB each). Two Lance tables, 1M rows each, 10 fragments with 10 regions, zonemap index on `region`. Inner join on `region` column.

| Metric | Without SPJ | With SPJ | Speedup |
|---|---|---|---|
| **Inner join** (median of 3 runs) | 329.8s | 27.3s | **12.1×** |
| Filtered join (`region = 'east'`) | 16.2s | 31.4s | 0.5× |

**Key observations:**

- **12.1× speedup on full inner join**: SPJ completely eliminates the shuffle stage. Without SPJ, Spark must hash-partition and shuffle all 1M × 1M matching rows across executors; with SPJ, each executor joins its local partition in-place via co-located SortMergeJoin.

- **Filtered join is slower with SPJ (expected)**: When `region = 'east'` is applied, zonemap pruning reduces Table A to a single fragment (100K rows). Without SPJ, the left side is so small that the shuffle cost is negligible — and AQE (Adaptive Query Execution) can convert the SortMergeJoin to a BroadcastHashJoin at runtime, making it very fast. With SPJ, Spark is locked into the storage-partitioned execution strategy: it must align all 10 partitions 1:1 between the two sides and execute 10 SortMergeJoin tasks, even though 9 of them produce zero rows (the left side is empty for those partitions). SPJ also prevents AQE from converting to BroadcastHashJoin because the join is already marked as "storage-partitioned." This overhead dominates when one side is tiny. This is a well-known tradeoff — Iceberg and Delta SPJ exhibit the same behavior. SPJ's value is in large, unfiltered joins where shuffle elimination saves orders of magnitude more time than the partition alignment costs.

**Execution plan with SPJ enabled** (no Exchange/shuffle nodes):
```
== Physical Plan ==
*(2) SortMergeJoin [region#...], [region#...], Inner
:- *(1) Sort [region#... ASC NULLS FIRST]
:  +- BatchScan lance [...] (storage-partitioned)
+- *(1) Sort [region#... ASC NULLS FIRST]
   +- BatchScan lance [...] (storage-partitioned)
```

**Execution plan without SPJ** (Exchange/shuffle present):
```
== Physical Plan ==
*(5) SortMergeJoin [region#...], [region#...], Inner
:- *(2) Sort [region#... ASC NULLS FIRST]
:  +- Exchange hashpartitioning(region#..., 200)    ← SHUFFLE
:     +- BatchScan lance [...]
+- *(4) Sort [region#... ASC NULLS FIRST]
   +- Exchange hashpartitioning(region#..., 200)    ← SHUFFLE
      +- BatchScan lance [...]
```

### Known limitations
- SPJ only activates when **every fragment** has `min == max` for the partition column. Boundary fragments with mixed values (common after `OPTIMIZE ORDER BY`) will prevent SPJ detection. Future work: sub-fragment row-range splits to support partially-sorted data.

### Future work
- [ ] **Adaptive SPJ → BroadcastHashJoin fallback**: When one side of a storage-partitioned join is small (e.g., after filter pushdown reduces it to a single fragment), SPJ's partition alignment overhead outweighs its benefits. Investigate reporting accurate `SupportsReportStatistics` size estimates so AQE can fall back to BroadcastHashJoin when one side is below `autoBroadcastJoinThreshold`, rather than being locked into the SPJ execution path.
- [ ] **Sub-fragment row-range splits**: Support partially-sorted data where `min != max` within a fragment by splitting at row-range boundaries, enabling SPJ on tables where `OPTIMIZE ORDER BY` leaves boundary fragments with mixed values.

### Dependencies
Requires lance-core with `Dataset.getZonemapStats()` API — merged in https://github.com/lance-format/lance/pull/6421 but **not yet included in a published release**. CI will fail until a new lance-core version (> 5.0.0-beta.5) is released.

## Test plan
- [x] 29 unit tests for `ZonemapFragmentPruner` covering all filter types, multi-column, null handling, contradictions
- [x] 21 unit tests for `ZonemapPartitionDetector` covering detection, rejection, serialization, partition key types
- [x] 4 unit tests for SPJ behavior (`HasPartitionKey`, `outputPartitioning` with/without partition info, fallback to fragment ID)
- [x] All 260 tests pass locally (with locally-built lance-core that includes `ZoneStats`)
- [x] End-to-end SPJ validated on Spark 4.0 Kubernetes cluster with 1M-row Lance tables


hamersaw (commented):
It looks like this is pruning at the fragment-level based on ZoneMap stats. I'm not sure what performance numbers you need to reach, but two important things to note early on:
1. For ZoneMap to be effective in SPJ I think we need to make sure there a large sequences of sorted rows. This will probably mean sorting across multiple fragments (ex. Delta's liquid clustering Z-Cubes). You could tailor insertion of the dataset to be sorted as a test, but in production I suspect this will mean rewriting data.
2. Lance Spark partitions are single fragments right now. I think we're going to need to make this both finer and coarser grained; meaning to be optimally effective we should pack multiple fragment row ranges within a partition depending on how the ZoneMap stats layout. This we we can minimize read-amplification. For example:
```
partition 0:
   fragment A
       rows [0 .. 1024]
       rows[2048...4048]
partition 1:
    fragment D
        rows[..4048]
partition 2:
    fragment X
        rows[1024 ..]
    fragment Y
        rows[512 ... 2048]
```

hamersaw (commented):
Thanks for working on this! My biggest concern is the limited applicability of this solution. The `ZonemapPartitionDetector.java` only labels columns as partitionable if every fragment contains only a single value, then each fragment individually is partitioned on the key. The hope is that this solution (ie. ZoneMap indexes + Lance fast random access) would be able to relax these expectations.

I think in a perfect world we:
- Extend partitions to support multiple fragments
- Use ZoneMaps to compile the list of unique values / ranges for a specific column (or multiple columns). This is probably more difficult than it sounds because each zone just contains min / max values.
- Create partitions that operate over each fragment that contains data from the unique partition column value.
- Rely on the LanceScanner implementation within each partition (using ZoneMap internally) to only read data from zones that contain the data for the partition it's reading. This may require pushing down filters for the JOIN key as well.

With ^^^ we can have SPJ over ANY data distribution, just sorted / bucketed will be significantly more efficient.

hamersaw (approved):

PR #403, opened 2026-04-09
feat: add TPC-H benchmark

### LINKED ISSUE #242: Setup TPCH benchmark
Setup TPCH benchmark to understand regression from Iceberg/Parquet

wombatu-kun:
Closes https://github.com/lance-format/lance-spark/issues/242

Implemented on the basis of TPC-DS benchmark infrastructure and similarly to it.

### Files changed (4 in-place edits)
1. **`Makefile`** — Added `benchmark-tpch-generate`, `benchmark-tpch-run`, `benchmark-tpch` targets. Updated `make help` block to list all benchmark targets with TPC-DS vs TPC-H distinction.
2. **`benchmark/README.md`** — Restructured to cover both benchmarks with shared Architecture, Data Layout (explaining the `customer` schema collision), Docker, and External Cluster sections. Every Quick Start leads with `make`.
3. **`benchmark/pom.xml`** — Added `kyuubi-spark-connector-tpch_${scala.compat.version}` dependency (v1.11.0), renamed to "Lance Spark Benchmark (TPC-DS + TPC-H)", removed stale `<mainClass>TpcdsBenchmarkRunner</mainClass>` from both `maven-jar-plugin` and `maven-shade-plugin` (since there's no longer a single default runner).
4. **`BenchmarkReporter.java`** — Added 2-arg constructor `BenchmarkReporter(List<BenchmarkResult>, String benchmarkName)`. Existing 1-arg constructor delegates with `"TPC-DS"` (backward compatible — TpcdsBenchmarkRunner needs no changes). Line 94 header string now uses the parameter.

### Files added (31 new files)

**Java classes (4)** — all in `org.lance.spark.benchmark` package:
- `TpchDataGenerator.java` — 8 TPC-H tables via Kyuubi `tpch.sfN` catalog, reuses `TpcdsDataGenerator.toLancePath()` for cross-cloud path conversion
- `TpchDataLoader.java` — registers/unregisters TPC-H temp views
- `TpchQueryRunner.java` — loads from `/tpch-queries/` classpath, iterates q1..q22 only (no a/b variant loop), drops the dead 2-arg convenience constructor from the TPC-DS equivalent
- `TpchBenchmarkRunner.java` — app name "TPC-H Benchmark", uses new 2-arg `BenchmarkReporter` with "TPC-H", writes CSV with `tpch_` prefix

**SQL query files (22)** — `benchmark/src/main/resources/tpch-queries/q1.sql` through `q22.sql`, sourced from Apache Kyuubi's TPC-H test resources (Apache 2.0 licensed), each with a provenance comment header. **q15 uses the CTE form** (verified by `grep CREATE VIEW` returning no matches) to avoid the multi-statement row-count edge case in the `;`-splitting query runner.

**Shell scripts (4)** — all chmod +x, all with `set -euo pipefail`, all parametrized via `SPARK_VERSION`/`SCALA_VERSION`/`DATA_DIR`/`RESULTS_DIR` env vars (zero hardcoded versions):
- `generate-tpch-data.sh` — simple wrapper, defaults `DATA_DIR` to `${BENCHMARK_DIR}/data/tpch`
- `run-tpch-benchmark.sh` — simple wrapper, same `EXPLAIN`/`METRICS`/`QUERIES` env passthrough as TPC-DS
- `submit-tpch-datagen.sh` — advanced wrapper with full arg parsing, Kyuubi TPC-H catalog config, `APP_NAME` defaults to `${USER}-tpch-datagen-sfN`
- `submit-tpch-benchmark.sh` — advanced wrapper with cluster-ready configuration

### DIFF
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -7,8 +7,8 @@
     <version>0.3.0-beta.1</version>
     <packaging>jar</packaging>
 
-    <name>Lance Spark TPC-DS Benchmark</name>
-    <description>TPC-DS benchmark for Lance Spark connector</description>
+    <name>Lance Spark Benchmark (TPC-DS + TPC-H)</name>
+    <description>TPC-DS and TPC-H benchmarks for Lance Spark connector</description>
 
     <properties>
         <spark.compat.version>3.5</spark.compat.version>
@@ -33,6 +33,13 @@
             <artifactId>kyuubi-spark-connector-tpcds_${scala.compat.version}</artifactId>
             <version>${kyuubi.version}</version>
         </dependency>
+        <!-- Kyuubi TPC-H connector: pure-Spark data source for parallel TPC-H
+             data generation. No external dbgen binary needed. -->
+        <dependency>
+            <groupId>org.apache.kyuubi</groupId>
+            <artifactId>kyuubi-spark-connector-tpch_${scala.compat.version}</artifactId>
+            <version>${kyuubi.version}</version>
+        </dependency>
     </dependencies>
 
     <build>
@@ -48,13 +55,6 @@
             <plugin>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>3.4.2</version>
-                <configuration>
-                    <archive>
-                        <manifest>
-                            <mainClass>org.lance.spark.benchmark.TpcdsBenchmarkRunner</mainClass>
-                        </manifest>
-                    </archive>
-                </configuration>
             </plugin>
             <plugin>
                 <artifactId>maven-shade-plugin</artifactId>
@@ -67,9 +67,6 @@
                         </goals>
                         <configuration>
                             <transformers>
-                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
-                                    <mainClass>org.lance.spark.benchmark.TpcdsBenchmarkRunner</mainClass>
-                                </transformer>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
                             </transformers>
                             <filters>
--- a/benchmark/pom.xml
+++ b/benchmark/pom.xml
@@ -55,6 +55,13 @@
             <plugin>
                 <artifactId>maven-jar-plugin</artifactId>
                 <version>3.4.2</version>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <mainClass>org.lance.spark.benchmark.TpcdsBenchmarkRunner</mainClass>
+                        </manifest>
+                    </archive>
+                </configuration>
             </plugin>
             <plugin>
                 <artifactId>maven-shade-plugin</artifactId>
@@ -67,6 +74,9 @@
                         </goals>
                         <configuration>
                             <transformers>
+                                <transformer implementation="org.apache.maven.plugins.shade.resource.ManifestResourceTransformer">
+                                    <mainClass>org.lance.spark.benchmark.TpcdsBenchmarkRunner</mainClass>
+                                </transformer>
                                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer" />
                             </transformers>
                             <filters>
--- a/benchmark/scripts/generate-tpch-data.sh
+++ b/benchmark/scripts/generate-tpch-data.sh
@@ -46,6 +46,15 @@ echo "Spark master:    ${SPARK_MASTER}"
 echo "Spark version:   ${SPARK_VERSION}"
 echo "Scala version:   ${SCALA_VERSION}"
 echo "Data dir:        ${DATA_DIR}"
+if [ -n "${FILE_FORMAT_VERSION:-}" ]; then
+  echo "File format version: ${FILE_FORMAT_VERSION}"
+fi
+if [ -n "${MAX_BYTES_PER_FILE:-}" ]; then
+  echo "Max bytes/file:  ${MAX_BYTES_PER_FILE}"
+fi
+if [ -n "${MAX_ROWS_PER_FILE:-}" ]; then
+  echo "Max rows/file:   ${MAX_ROWS_PER_FILE}"
+fi
 echo ""
 
 # Step 1: Build benchmark jar if needed
@@ -81,12 +90,14 @@ ${SPARK_SUBMIT} \
   --driver-memory "${DRIVER_MEMORY:-4g}" \
   --executor-memory "${EXECUTOR_MEMORY:-4g}" \
   --jars "${BUNDLE_JAR}" \
-  --conf spark.sql.extensions=org.lance.spark.LanceSparkSessionExtension \
+  --conf spark.sql.extensions=org.lance.spark.extensions.LanceSparkSessionExtensions \
   --conf spark.driver.extraJavaOptions="-XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED -Dio.netty.tryReflectionSetAccessible=true" \
+  ${MAX_BYTES_PER_FILE:+ --conf spark.sql.catalog.lance_default.max_bytes_per_file="${MAX_BYTES_PER_FILE}"} \
+  ${MAX_ROWS_PER_FILE:+ --conf spark.sql.catalog.lance_default.max_row_per_file="${MAX_ROWS_PER_FILE}"} \
   "${BENCHMARK_JAR}" \
   --data-dir "${DATA_DIR}" \
   --scale-factor "${SCALE_FACTOR}" \
-  --formats "${FORMATS}"
+  --formats "${FORMATS}"${FILE_FORMAT_VERSION:+ --file-format-version "${FILE_FORMAT_VERSION}"}
 
 echo ""
 echo "=== Data generation complete ==="
--- a/benchmark/scripts/generate-tpch-data.sh
+++ b/benchmark/scripts/generate-tpch-data.sh
@@ -1,0 +1,92 @@
+#!/usr/bin/env bash
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# TPC-H data generation via Spark (using Kyuubi TPC-H connector).
+#
+# Generates TPC-H tables in parallel across Spark executors and writes
+# them directly into the target format(s) — no intermediate files.
+#
+# Note: defaults DATA_DIR to ${BENCHMARK_DIR}/data/tpch to avoid schema
+# collision with TPC-DS (both suites define a `customer` table with
+# different schemas).
+#
+# Usage:
+#   ./generate-tpch-data.sh [SCALE_FACTOR] [FORMATS] [SPARK_MASTER]
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+BENCHMARK_DIR="${SCRIPT_DIR}/.."
+
+# Configurable Spark/Scala versions (override via environment)
+SPARK_VERSION="${SPARK_VERSION:-3.5}"
+SCALA_VERSION="${SCALA_VERSION:-2.12}"
+
+SCALE_FACTOR="${1:-1}"
+FORMATS="${2:-parquet,lance}"
+SPARK_MASTER="${3:-local[*]}"
+
+# Object store / external paths — default to local when unset
+DATA_DIR="${DATA_DIR:-${BENCHMARK_DIR}/data/tpch}"
+
+echo "=== TPC-H Data Generation ==="
+echo "Scale factor:    ${SCALE_FACTOR}"
+echo "Formats:         ${FORMATS}"
+echo "Spark master:    ${SPARK_MASTER}"
+echo "Spark version:   ${SPARK_VERSION}"
+echo "Scala version:   ${SCALA_VERSION}"
+echo "Data dir:        ${DATA_DIR}"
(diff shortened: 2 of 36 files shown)


benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
```diff
@@ -0,0 +1,181 @@
...
+      }
+
+      for (int i = 1; i <= iterations; i++) {
+        BenchmarkResult result = runQuery(queryName, format, sql, i);
+        results.add(result);
+
+        String status = result.isSuccess() ? "OK" : "FAIL";
+        System.out.printf(
+            "  [%s] %s iter=%d time=%dms rows=%d%n",
+            status, queryName, i, result.getElapsedMs(), result.getRowCount());
```
hamersaw: Compile error: `BenchmarkResult` has no `getRowCount()` method. Verified with `mvn compile`:
```
TpchQueryRunner.java:[80,64] cannot find symbol: method getRowCount()
```
Either add a `rowCount` field + getter to `BenchmarkResult`, or drop the field from the log line.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
```diff
@@ -0,0 +1,181 @@
...
+        if (trimmed.isEmpty()) {
+          continue;
+        }
+        Dataset<Row> result = spark.sql(trimmed);
+        rowCount = result.count();
+      }
+      long elapsed = System.currentTimeMillis() - start;
+
+      QueryMetrics metrics = metricsListener != null ? metricsListener.getMetrics() : null;
+      return BenchmarkResult.success(queryName, format, iteration, elapsed, rowCount, metrics);
```
hamersaw: Compile error: no `BenchmarkResult.success(String, String, int, long, long, QueryMetrics)` overload exists. Only `success(name, fmt, iter, elapsed)` and `success(name, fmt, iter, elapsed, metrics)` are defined. Either add the overload (plus the `rowCount` field/getter) or revert to the existing 5-arg signature.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
```diff
@@ -0,0 +1,181 @@
...
+    long start = System.currentTimeMillis();
+    try {
+      long rowCount = 0;
+      for (String stmt : statements) {
+        String trimmed = stmt.trim();
+        if (trimmed.isEmpty()) {
+          continue;
+        }
+        Dataset<Row> result = spark.sql(trimmed);
+        rowCount = result.count();
```
hamersaw: TPC-DS uses `result.write().format("noop").mode("overwrite").save()` here (see `TpcdsQueryRunner.java:138`). `result.count()` triggers a different physical plan — Spark can short-circuit count aggregation, so the timings won't be apples-to-apples between TPC-H and TPC-DS. Suggest matching the TPC-DS sink so cross-suite numbers are comparable.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
```diff
@@ -0,0 +1,181 @@
...
+
+public class TpchQueryRunner {
+
+  private final SparkSession spark;
+  private final int iterations;
+  private final boolean explain;
+  private final QueryMetricsListener metricsListener;
+  private final Set<String> queryFilter;
+
+  public TpchQueryRunner(
```
hamersaw, line 43: Minor parity: `TpcdsQueryRunner` exposes a `(SparkSession spark, int iterations)` convenience constructor that delegates to the full one. Worth mirroring here for symmetry with the TPC-DS surface.

benchmark/src/main/java/org/lance/spark/benchmark/TpchDataGenerator.java
```diff
@@ -0,0 +1,184 @@
...
+        case "--scale-factor":
+          scaleFactor = Integer.parseInt(args[++i]);
+          break;
+        case "--formats":
+          formatsStr = args[++i];
+          break;
+        case "--use-double-for-decimal":
+          useDoubleForDecimal = true;
+          break;
+        default:
```
hamersaw, line 80: Interface drift from `TpcdsDataGenerator`: missing the `--file-format-version <version>` arg that TPC-DS accepts and threads through to the Lance writer (`writer.option("file_format_version", ...)`). Lance's file format version applies to TPC-H writes too — please add the same flag for parity.

benchmark/pom.xml
```diff
@@ -48,13 +55,6 @@
             <plugin>
                 <artifactId>maven-jar-plugin</artifactId>
```
hamersaw, line 56: Removing `<mainClass>org.lance.spark.benchmark.TpcdsBenchmarkRunner</mainClass>` from both `maven-jar-plugin` (this block) and `maven-shade-plugin` (line 70 area) breaks existing TPC-DS users who invoke `java -jar benchmark.jar` without `--class`. Two options:
- Keep `TpcdsBenchmarkRunner` as the default `mainClass` (least churn for existing users).
- Introduce a neutral dispatcher entrypoint (e.g. `BenchmarkMain` that routes by first arg) and set that as the default.

Either way, the no-`--class` invocation path should keep working.

benchmark/scripts/generate-tpch-data.sh
```diff
@@ -0,0 +1,92 @@
...
+  SPARK_SUBMIT="${SPARK_HOME}/bin/spark-submit"
+fi
+
+${SPARK_SUBMIT} \
+  --class org.lance.spark.benchmark.TpchDataGenerator \
+  --master "${SPARK_MASTER}" \
+  --driver-memory "${DRIVER_MEMORY:-4g}" \
+  --executor-memory "${EXECUTOR_MEMORY:-4g}" \
+  --jars "${BUNDLE_JAR}" \
+  --conf spark.sql.extensions=org.lance.spark.LanceSparkSessionExtension \
```
hamersaw: Wrong Spark extension class — `org.lance.spark.LanceSparkSessionExtension` does not exist. The real class is `org.lance.spark.extensions.LanceSparkSessionExtensions` (plural, `extensions.` package), which is what TPC-DS's `generate-data.sh:89` uses. As written, the Lance extension won't load.

benchmark/scripts/generate-tpch-data.sh
```diff
@@ -0,0 +1,92 @@
...
+  --conf spark.driver.extraJavaOptions="-XX:+IgnoreUnrecognizedVMOptions --add-opens=java.base/java.lang=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED --add-opens=java.base/sun.nio.ch=ALL-UNNAMED -Dio.netty.tryReflectionSetAccessible=true" \
+  "${BENCHMARK_JAR}" \
+  --data-dir "${DATA_DIR}" \
+  --scale-factor "${SCALE_FACTOR}" \
+  --formats "${FORMATS}"
```
hamersaw: Interface drift from `generate-data.sh`: TPC-DS honors `MAX_BYTES_PER_FILE` / `MAX_ROWS_PER_FILE` env vars and passes them as `spark.sql.catalog.lance_default.max_bytes_per_file` / `max_row_per_file`. Same knobs would be valuable for TPC-H — please mirror that block here.

benchmark/scripts/run-tpch-benchmark.sh
```diff
@@ -0,0 +1,109 @@
...
+  BENCHMARK_EXTRA_ARGS="${BENCHMARK_EXTRA_ARGS} --queries ${QUERIES}"
+fi
+
+${SPARK_SUBMIT} \
+  --class org.lance.spark.benchmark.TpchBenchmarkRunner \
+  --master "${SPARK_MASTER}" \
+  --driver-memory "${DRIVER_MEMORY:-4g}" \
+  --executor-memory "${EXECUTOR_MEMORY:-4g}" \
+  --jars "${BUNDLE_JAR}" \
+  --conf spark.sql.extensions=org.lance.spark.LanceSparkSessionExtension \
```
hamersaw: Same wrong extension class as `generate-tpch-data.sh:84` — should be `org.lance.spark.extensions.LanceSparkSessionExtensions`. Compare with TPC-DS's `run-benchmark.sh:95`.

benchmark/scripts/submit-tpch-datagen.sh
```diff
@@ -0,0 +1,323 @@
...
+            DRIVER_MEMORY="$2"; shift 2 ;;
+        --executor-memory)
+            EXECUTOR_MEMORY="$2"; shift 2 ;;
+        --executor-cores)
+            EXECUTOR_CORES="$2"; shift 2 ;;
+        --benchmark-jar)
+            BENCHMARK_JAR="$2"; shift 2 ;;
+        --app-name)
+            APP_NAME="$2"; shift 2 ;;
+        --use-double-for-decimal)
```
hamersaw, line 160: Missing `--file-format-version` arg parsing here (and the corresponding `SUBMIT_ARGS+=("--file-format-version" ...)` push around line 300, plus the `FILE_FORMAT_VERSION` var declaration and usage docs). TPC-DS's `submit-datagen.sh` exposes this; please mirror it once `TpchDataGenerator` accepts the flag.

hamersaw (commented):
Thanks for getting TPC-H wired up — the structure mirrors the TPC-DS suite cleanly. Two themes in this review: (1) a handful of issues that would change existing TPC-DS behavior, and (2) some interface drift between the TPC-H and TPC-DS surface area worth tightening before merge.

**Branch is behind `main` (3 commits).** As-is, merging would revert unrelated features:
- `2274c1e` — Date(MILLISECOND) Arrow roundtrip (`DateMilliUtils`, related tests, `LanceArrowUtils.scala`)
- `f2ca031` — Custom Lance read metrics (`LanceCustomMetrics`, `CustomNsTimeMetric`, `LanceReadMetricsTracker`, integration tests, `docs/src/performance.md`)
- `4332b60` — Lance dep bump to `6.0.0-rc.2` (plus the Dockerfile bump from `0.4.0-beta.3` → `0.4.0-beta.4`)

A rebase onto `main` should make the resulting diff a pure addition.

**Build is broken.** The benchmark module no longer compiles — see inline comments on `TpchQueryRunner.java`. This also blocks the existing TPC-DS jar build.

**Optional / informational:** the wrong `spark.sql.extensions` class name (`org.lance.spark.LanceSparkSessionExtension`) is already on `main` in `submit-datagen.sh` and `submit-benchmark.sh`. Not introduced here, but the new `submit-tpch-*.sh` scripts copy that bug and the new local-run scripts (`generate-tpch-data.sh`, `run-tpch-benchmark.sh`) introduce a fresh instance — flagged inline. Worth a follow-up to fix all four.

Leaving this as comments rather than a formal Request Changes so you can address inline before re-review.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
wombatu-kun: Fixed in ad371f9 — dropped `getRowCount()` from the log line.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
wombatu-kun: Fixed in ad371f9 — reverted to the existing 5-arg `success(name, fmt, iter, elapsed, metrics)` overload.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
wombatu-kun: Fixed in ad371f9 — switched to `result.write().format("noop").mode("overwrite").save()`, matching `TpcdsQueryRunner.java:138`.

benchmark/src/main/java/org/lance/spark/benchmark/TpchQueryRunner.java
wombatu-kun: Added in ad371f9 — convenience constructor `TpchQueryRunner(SparkSession spark, int iterations)` that delegates to the full one, mirroring `TpcdsQueryRunner.java:39`.

benchmark/src/main/java/org/lance/spark/benchmark/TpchDataGenerator.java
wombatu-kun: Added in ad371f9 — `--file-format-version <version>` arg parsing + threading through to `writer.option("file_format_version", ...)` for Lance writes, mirroring `TpcdsDataGenerator`.

benchmark/pom.xml
wombatu-kun: Restored in ad371f9 — `TpcdsBenchmarkRunner` as the default `mainClass` on both `maven-jar-plugin` and `maven-shade-plugin/ManifestResourceTransformer`. `java -jar benchmark.jar` keeps working for existing TPC-DS users; TPC-H users pass `--class` explicitly (scripts already do).

benchmark/scripts/generate-tpch-data.sh
wombatu-kun: Fixed in ad371f9 — corrected to `org.lance.spark.extensions.LanceSparkSessionExtensions` (matching `generate-data.sh:89`).

benchmark/scripts/generate-tpch-data.sh
wombatu-kun: Added in ad371f9 — mirrored the `MAX_BYTES_PER_FILE` / `MAX_ROWS_PER_FILE` / `FILE_FORMAT_VERSION` echo block + `--conf spark.sql.catalog.lance_default.max_bytes_per_file` / `max_row_per_file` passthroughs from `generate-data.sh`.

benchmark/scripts/run-tpch-benchmark.sh
wombatu-kun: Fixed in ad371f9 — corrected to `org.lance.spark.extensions.LanceSparkSessionExtensions` (matching `run-benchmark.sh:95`).

benchmark/scripts/submit-tpch-datagen.sh
wombatu-kun: Added in ad371f9 — `FILE_FORMAT_VERSION` declaration, `--file-format-version VER` help line + example, parser case, `SUBMIT_ARGS` push, and summary echo, all mirroring `submit-datagen.sh`.

hamersaw (approved):
Looks great! Thanks for the patience. I noticed we don't build any indexes for the TPC-H benchmarks, it would be really interesting to compare performance between indexed lookups and non.

PR #405, opened 2026-04-09
fix: add --fail flag to curl downloads in docker/Dockerfile

wombatu-kun:
## Summary

Adds `-fL` to every `curl` invocation in `docker/Dockerfile` and `docker/Dockerfile.test-base` so that the build fails loudly when an HTTP download returns a 4xx/5xx response, instead of silently saving the error page as the output file.

Without `-f`, curl exits with status 0 on HTTP 4xx/5xx and writes the response body (usually an HTML error page) into the output file. This caused `lance-spark-bundle_2.12-0.4.0-beta.1.jar` and `lance-namespace-glue-0.6.1.jar` to be silently created as 554-byte HTML 404 pages inside the spark-lance image, which then crashed Spark at runtime with `Failed to find the data source: lance`.

`-L` is added uniformly so that Maven Central and AWS redirects are followed (some curl calls already had `-L`, others didn't — this also standardizes on `curl -fL` everywhere).

Fixes #404

## What this PR does NOT do

This PR is intentionally minimal. It does not:
- Update `LANCE_SPARK_VERSION` (currently `0.4.0-beta.1`, not on Maven Central)
- Update `LANCE_NS_VERSION` (currently `0.6.1`, also not on Maven Central)
- Update opendal / AWS SDK versions
- Add hadolint or any other Dockerfile linter to CI

These are separate maintainer decisions; the goal here is just to make the existing pipeline fail explicitly when something is wrong, regardless of which version the maintainers eventually settle on.

## Test plan

- [ ] `docker build -f docker/Dockerfile .` still builds successfully when all upstream URLs are valid (or fails immediately with `curl: (22)` when they aren't)
- [ ] Manual negative test: temporarily set `LANCE_SPARK_VERSION=does-not-exist` and verify `make docker-build` now fails at the curl step with `curl: (22) The requested URL returned error: 404`, instead of silently producing a broken image
- [ ] No runtime behaviour change for valid downloads — `-f` only affects HTTP error responses, and `-L` is harmless on non-redirect URLs

## Precedent

Same class of bug as #183 (which addressed the Spark tarball download), now applied uniformly to all downloads in both Dockerfiles so the trap doesn't catch the next person who bumps a version to something temporarily missing.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -47,16 +47,16 @@ ENV LANCE_NS_VERSION=0.6.1
 
 # Download spark
 RUN mkdir -p ${SPARK_HOME} \
- && curl https://dlcdn.apache.org/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3${SPARK_SCALA_SUFFIX}.tgz -o spark.tgz \
+ && curl -fL https://dlcdn.apache.org/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop3${SPARK_SCALA_SUFFIX}.tgz -o spark.tgz \
  && tar xvzf spark.tgz --directory /opt/spark --strip-components 1 \
  && rm -rf spark.tgz
 
 # Download lance-spark bundle JAR from Maven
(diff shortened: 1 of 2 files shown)


hamersaw (approved):
Thanks for the fix!

PR #407, opened 2026-04-09
feat: implement native update path with stable row-id preservation

ivscheianu:
Fix for https://github.com/lance-format/lance-spark/issues/406

## Summary

Reworks `SparkPositionDeltaWrite` (Spark 3.5) to use Spark's native `DeltaWriter.update()` method instead of `representUpdateAsDeleteAndInsert()`, enabling stable row-ID preservation across updates.

### Problem

The previous implementation threw `UnsupportedOperationException` from `update()`, forcing Spark to decompose updates into separate delete + insert operations. This prevented the connector from knowing which `_rowid` maps to which newly written row, making it impossible to attach `RowIdMeta` to new
fragments. Without `RowIdMeta`, lance-core cannot trace updated rows back to their originals, breaking `_row_created_at_version` tracking and stable row-ID continuity.

Additionally, each worker opened its own `Dataset` to apply deletions, causing unnecessary I/O and potential race conditions.

### Changes

**Spark 3.5 (`SparkPositionDeltaWrite`):**
- Implement `update()`: capture `_rowid` from the id row, record deletion via `_rowaddr`, and write the data row — all in one call.
- On `commit()`: if stable row IDs are enabled, attach `RowIdMeta` to new fragments using `RowIdMeta.fromRowIds()` (from companion lance PR).
- Move deletion application from workers to driver: tasks send `Map<fragmentId, RoaringBitmap>` to the driver, which consolidates and applies deletions in a single `Dataset` session.
- Enable `useStableRowIds(true)` on `CommitBuilder` when the dataset has them.

**Spark 3.4 (`SparkPositionDeltaWrite`):**
- Same driver-side deletion consolidation refactor for consistency.
- No row-ID capture (3.4's DSv2 API doesn't support `update()`).

**CDF test expectation corrections:**
- `BaseCdfVersionTrackingTest`, `BaseCdfConfigTest`, `BaseCdfQueryPatternsTest`: updated `_row_created_at_version` expectations from `1` to the actual inser version (typically `2`). Previously tests passed only because the JNI gap in lance-core was silently dropping version metadata.

## Companion PR

This PR depends on https://github.com/lance-format/lance/pull/6465 which fixes the JNI version metadata serialization gap and the row-ID lookup bug in `Operation::Update`. Both PRs are needed for correct end-to-end `_row_created_at_version` tracking.


ivscheianu:
Hello @hamersaw , just a gentle nudge whenever you have a bit of time, I’d really appreciate your review. Just wanted to make sure it doesn’t get buried. It’s an older draft that was waiting on the lance-core release. No rush, thank you so much!

lance-spark-3.4_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
```diff
@@ -137,38 +136,65 @@ public DeltaWriterFactory createBatchWriterFactory(PhysicalWriteInfo info) {
...
+
+      for (WriterCommitMessage msg : messages) {
+        DeltaWriteTaskCommit m = (DeltaWriteTaskCommit) msg;
+        newFragments.addAll(m.newFragments());
+        m.deletionMap()
+            .forEach(
+                (fragId, bitmap) ->
+                    aggregatedDeletions.merge(
+                        fragId,
+                        bitmap,
```
hamersaw: I think this function modifies the fragment deletions in place, meaning we modify the first message to aggregate all of them. This could be problematic if the metadata is used, for example, on pipeline retries. We would want to copy the first entry to avoid this issue.

lance-spark-3.4_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
```diff
@@ -137,38 +136,65 @@ public DeltaWriterFactory createBatchWriterFactory(PhysicalWriteInfo info) {
...
+              List<Integer> rowIndexes = new ArrayList<>();
+              IntIterator it = bitmap.getIntIterator();
+              while (it.hasNext()) {
+                rowIndexes.add(it.next());
+              }
+              if (rowIndexes.isEmpty()) {
+                return;
+              }
+              FragmentMetadata updatedFragment =
+                  dataset.getFragment(fragmentId).deleteRows(ImmutableList.copyOf(rowIndexes));
```
hamersaw: This change makes `deleteRows` a serial call in the Spark driver rather than parallelized across executors. In scenarios where this is an expensive operation there could be performance degradation. You could fix this maybe by using `parallelStream()` over `aggregatedDeletions.entrySet()` or move the deletion logic back to executors and do manifest-level updates here.

hamersaw (commented):
Thanks for the fix! Apologies on the slow review. A few small things that would be good to iron out. Additionally, I think it makes sense to add an integration test for this so we can ensure functionality across Spark versions.

lance-spark-3.4_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
ivscheianu: Fixed by cloning the bitmap on first insertion into the aggregation map, thank you!

lance-spark-3.4_2.12/src/main/java/org/lance/spark/write/SparkPositionDeltaWrite.java
ivscheianu: Agree, refactored to `aggregatedDeletions.entrySet().parallelStream()` with `map()` + `Collectors.toList()`. Each `deleteRows()` operates on an independent FileFragment writing to a distinct object-store path, so parallel execution across fragments is safe. The collector handles thread safe accumulation internally

hamersaw (approved):
Thanks for the quick iteration! This looks great. Again, appreciate the patience.

Issue #168, opened 2026-01-06
Distributed index building for Zonemap index


jtuglu1:

westonpace:
> I recall there was some discussion about whether we should build or optimize the zonemap index on every commit (though I can’t remember the exact thread).

Perhaps you are recalling the discussion on statistics?

A PR was recently added: https://github.com/lance-format/lance/pull/5639

I haven't had a chance to look yet but am planning to soon.

PR #412, opened 2026-04-10
fix: intercept Spark 4.0+ native CREATE INDEX to prevent NPE

beinan:
## Summary

Fixes #400

Spark 4.0+ rejects `CREATE INDEX` with `ParseException`, which falls through to Lance's custom ANTLR parser. But the grammar only supported `ALTER TABLE ... CREATE INDEX` syntax, so `CREATE INDEX idx ON table (col) USING method` failed — Spark 3.4/3.5 fail similarly but with a clearer parse error, while Spark 4.0+ produces an NPE in `QueryExecution`.

## Changes

- **ANTLR grammar**: Added `CREATE INDEX idx ON table (cols) USING method [WITH (...)]` rule alongside the existing `ALTER TABLE ... CREATE INDEX` syntax
- **AST builders** (all 4 versions): Added `visitCreateIndexStandard` to handle the new grammar rule
- **Spark 4.0/4.1 parsers**: Added `CreateIndex` plan interception as defense in depth (for potential future Spark versions where the native parser may succeed)
- **Spark 4.0/4.1 pom.xml**: Added `src/test/java` as test source directory
- **Tests**: Added `CreateIndexStandardSyntaxTest` for Spark 4.0 and 4.1

## Notes

As @hamersaw noted, ZONEMAP is not supported by lance-core for distributed building (#168). With this fix, `CREATE INDEX ... USING ZONEMAP` gets a clear `UnsupportedOperationException` instead of NPE.

## Test plan

- [x] All 569 tests pass for Spark 4.0 (0 failures, 38 skipped)
- [x] All 569 tests pass for Spark 4.1 (0 failures, 38 skipped)  
- [x] Spark 3.5 AddIndexTest passes (11 tests)
- [x] New `CreateIndexStandardSyntaxTest` passes on both 4.0 and 4.1

### DIFF
--- a/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -94,6 +94,21 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
     AddIndex(table, indexName, method, columns, args)
   }
 
+  override def visitCreateIndexStandard(
+      ctx: LanceSqlExtensionsParser.CreateIndexStandardContext): AddIndex = {
+    val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
+    val indexName = cleanIdentifier(ctx.indexName.getText)
+    val method = cleanIdentifier(ctx.method.getText)
+    val columns = visitColumnList(ctx.columnList())
+    val args = ctx.namedArgument().asScala.map(a =>
+      LanceNamedArgument(
+        cleanIdentifier(a.identifier().getText),
+        a.constant().accept(this)))
+      .toSeq
+
+    AddIndex(table, indexName, method, columns, args)
+  }
+
   override def visitShowIndexes(ctx: LanceSqlExtensionsParser.ShowIndexesContext): LogicalPlan = {
     val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
     ShowIndexes(table)
--- a/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.4_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -94,21 +94,6 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
     AddIndex(table, indexName, method, columns, args)
   }
 
-  override def visitCreateIndexStandard(
-      ctx: LanceSqlExtensionsParser.CreateIndexStandardContext): AddIndex = {
-    val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
-    val indexName = cleanIdentifier(ctx.indexName.getText)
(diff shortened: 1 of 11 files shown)


hamersaw:
I don't necessarily love introducing multiple semantics for creating an index, IMO it unnecessarily increases surface area and makes maintainability more complex. Since `CREATE INDEX` is native in 4.0+, what do you think about scoping this to those versions? So rather than adding the `CREATE INDEX ..` grammer rule and parsing option in 3.4 / 3.5 to backport `CREATE INDEX` support, we just scope this to the `parsePlan` interception in the 4.0 and 4.1 `LanceSparkSqlExtensionsParser`?

beinan:
Good point on surface area. I investigated the `parsePlan` interception-only approach, but it won't work here.

Despite Spark 4.0/4.1 having a `CreateIndex` class and `AstBuilder.visitCreateIndex()`, the runtime parser routes `CREATE INDEX` through `visitFailNativeCommand` in `SparkSqlAstBuilder`, which throws `ParseException`. So `delegate.parsePlan("CREATE INDEX ...")` throws rather than returning a `CreateIndex` plan — there's nothing to intercept.

I confirmed this with debug logging: the delegate throws `ParseException`, Lance's ANTLR parser catches it and tries to parse, but fails because the grammar doesn't have the standard syntax.

So the grammar change is the necessary fix. That said, I'm happy to scope it more narrowly if you prefer — for example:
1. **Keep the grammar rule** but only in the shared grammar (as-is) — the `visitCreateIndexStandard` in 3.4/3.5 AST builders could be removed if we don't want to support it there, but it's harmless since those versions would never reach it (Spark 3.4/3.5 don't parse `CREATE INDEX` at all, so `ParseException` → Lance parser → grammar match)
2. Or if you'd rather not touch the grammar, we could do a string-level rewrite in the 4.0/4.1 `parsePlan`: detect `CREATE INDEX` pattern and rewrite it to `ALTER TABLE ... CREATE INDEX` before passing to the Lance parser

What's your preference?

hamersaw:
Ok, so IIUC then the interception logic on 4.0 / 4.1 that is currently proposed does not actually do anything? Is the overall goal to not NPE or to support `CREATE INDEX ...`? If it's the former why don't we add a small intercept in 4.0+ with something like:
```
if (sqlText.trim.toUpperCase.startsWith("CREATE INDEX")) {
      throw new ParseException(
        "Lance does not support standard CREATE INDEX syntax. " +
        "Use: ALTER TABLE <table> CREATE INDEX <name> USING <method> (<columns>)")
    }
```
Then we gracefully fail with a useful error message. This makes it a two-way door decision, where if we every have a compelling reason to support `CREATE INDEX` we remove the intercept and do it. Alternatively, if we decide to support it right now, it will be really difficult to pull back on that.

beinan:
Good call — agreed this should be a two-way door. I've updated the PR to:

1. **Remove** the `CREATE INDEX` grammar rule from the ANTLR grammar and all `visitCreateIndexStandard` visitor methods across 3.4/3.5/4.0/4.1
2. **Remove** the `CreateIndex` → `AddIndex` plan conversion in the 4.0/4.1 parsers
3. **Add** a simple `CREATE INDEX` string intercept in the 4.0/4.1 `parsePlan` that throws `UnsupportedOperationException` with a helpful error message directing users to `ALTER TABLE <table> CREATE INDEX <name> USING <method> (<columns>)`
4. **Update** the tests to verify the error is thrown with the correct message

Used `UnsupportedOperationException` instead of `ParseException` because Spark 4.0+ requires registered error class IDs for `ParseException` — a custom string message gets rejected at runtime. The UOE gives a clean, clear error.

All tests pass (12/12 for both 4.0 and 4.1).

hamersaw (approved):
Appreciate the patience on getting this one through, thanks!

PR #413, opened 2026-04-10
feat: add use_large_var_types option to avoid 2GB Arrow vector overflow

beinan:
## Summary
- Adds a `use_large_var_types` write option (default `false`) that switches string/binary columns from `VarCharVector`/`VarBinaryVector` (32-bit offsets, 2GB limit) to `LargeVarCharVector`/`LargeVarBinaryVector` (64-bit offsets, no practical limit)
- Fixes `OversizedAllocationException` when a single batch accumulates >2GB of variable-width data (e.g., large blobs or long strings)
- The flag propagates through nested types (arrays, structs, maps)

## Usage
```scala
df.write.format("lance")
  .option("use_large_var_types", "true")
  .save("/path/to/dataset")
```

## Test plan
- [x] All 52 existing unit tests pass with no regressions
- [ ] Integration test with large string/binary data exceeding 2GB per batch

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
@@ -165,13 +165,22 @@ object LanceArrowUtils {
       schema: StructType,
       timeZoneId: String,
       errorOnDuplicatedFieldNames: Boolean): Schema = {
+    toArrowSchema(schema, timeZoneId, errorOnDuplicatedFieldNames, largeVarTypes = false)
+  }
+
+  def toArrowSchema(
+      schema: StructType,
+      timeZoneId: String,
+      errorOnDuplicatedFieldNames: Boolean,
+      largeVarTypes: Boolean): Schema = {
     new Schema(schema.map { field =>
       toArrowField(
         field.name,
         deduplicateFieldNames(field.dataType, errorOnDuplicatedFieldNames),
         field.nullable,
         timeZoneId,
-        field.metadata)
+        field.metadata,
+        largeVarTypes)
     }.asJava)
   }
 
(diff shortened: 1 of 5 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -256,7 +281,8 @@ object LanceArrowUtils {
               .add(MapVector.KEY_NAME, keyType, nullable = false)
               .add(MapVector.VALUE_NAME, valueType, nullable = valueContainsNull),
             nullable = false,
-            timeZoneId)).asJava)
+            timeZoneId,
+            largeVarTypes = largeVarTypes)).asJava)
       case udt: UserDefinedType[_] =>
         toArrowField(name, udt.sqlType, nullable, timeZoneId)
```
hamersaw: Do we need to pass the largeVarTypes through this recursive function? Otherwise it defaults to false.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -234,15 +249,25 @@ object LanceArrowUtils {
...
           fieldType,
           fields.map { field =>
-            toArrowField(field.name, field.dataType, field.nullable, timeZoneId)
+            toArrowField(
+              field.name,
+              field.dataType,
+              field.nullable,
+              timeZoneId,
+              largeVarTypes = largeVarTypes)
           }.toSeq.asJava)
```
hamersaw, line 272: Can we include the `field.metadata` here as well. It's not related to this PR, but seems to be missing.

hamersaw (commented):
I think this is a great idea, a few small things then lets merge!

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
beinan: Good catch! Fixed — now passing `largeVarTypes` through the UDT recursive call as well.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
beinan: Good idea, added `field.metadata` to the struct children call so per-field metadata (e.g. `arrow:large-var-char`, blob encoding) is preserved for nested fields.

hamersaw (approved):
Looks great, thanks!

PR #415, opened 2026-04-10
fix(benchmark): accurately materialize tpcds query

### LINKED ISSUE #414: TPC-DS benchmark was not accurately measuring real query execution behavior
When execute benchmark q9, subqueries are not actually executed, only the main query are materialized
Image

**This means the benchmark was not accurately measuring real query execution behavior**

summaryzb:
Fix https://github.com/lance-format/lance-spark/issues/414

## Problem
The TPC-DS benchmark runner used `Dataset.count()` as the action to trigger query execution. However, `count()` activates Lance's count optimization, which scans with an empty schema (no columns). For queries containing scalar subqueries (e.g., TPC-DS q9), this optimization causes the subqueries to never be materialized

This means the benchmark was not accurately measuring real query execution behavior

## Approach
- The fix replaces `result.count()` with `result.write().format("noop").mode("overwrite").save()`, which forces Spark to fully materialize the query plan -- including all subqueries -- without triggering the count optimization shortcut. 
- The "noop" format is a built-in Spark sink that discards all output rows after computing them, making it an ideal replacement that still exercises the full query execution path.
- Since row counts are no longer available (the noop sink does not return a count), all row-count-related functionality is removed

## Test Coverage
No new tests. The benchmark module is a standalone harness rather than a tested library. To validate this change, one could:
Run TPC-DS q9 (trigger by df.show) against a Lance dataset and verify that the whole subqueries are materialized

### DIFF
--- a/benchmark/src/main/java/org/lance/spark/benchmark/TpcdsQueryRunner.java
+++ b/benchmark/src/main/java/org/lance/spark/benchmark/TpcdsQueryRunner.java
@@ -80,8 +80,8 @@ public class TpcdsQueryRunner {
 
         String status = result.isSuccess() ? "OK" : "FAIL";
         System.out.printf(
-            "  [%s] %s iter=%d time=%dms rows=%d%n",
-            status, queryName, i, result.getElapsedMs(), result.getRowCount());
+            "  [%s] %s iter=%d time=%dms%n",
+            status, queryName, i, result.getElapsedMs());
         if (result.getMetrics() != null) {
           System.out.println("       Metrics: " + result.getMetrics().toSummaryString());
         }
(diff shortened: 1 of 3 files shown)


benchmark/src/main/java/org/lance/spark/benchmark/TpcdsQueryRunner.java
```diff
@@ -129,19 +129,18 @@ private BenchmarkResult runQuery(String queryName, String format, String sql, in
...
     try {
-      long rowCount = 0;
       for (String stmt : statements) {
         String trimmed = stmt.trim();
         if (trimmed.isEmpty()) {
           continue;
         }
         Dataset<Row> result = spark.sql(trimmed);
-        rowCount = result.count();
+        result.write().format("noop").mode("overwrite").save();
```
summaryzb, line 138: Here is the main change to fix

hamersaw (approved):
This looks reasonable to me. Thanks! Care to share performance results?

PR #416, opened 2026-04-10
test: add unit tests for utility classes

LuciferYang:
## Summary
- Add unit tests for `BlobUtils`, `LargeVarCharUtils`, `Utils`, and `VectorUtils` covering all reachable branches in both Spark and Arrow field-checking methods
- Tests cover null inputs, missing/wrong metadata, case-insensitive value matching, type-check isolation (non-String for LargeVarChar, non-numeric for Vector), and edge cases like empty version lists and malformed metadata
- `UtilsTest` helper builds `ZonedDateTime` directly instead of re-implementing the production micros-to-Instant conversion, keeping test logic independent from source logic

## Test coverage details

| Class | Methods tested | Tests |
|-------|---------------|-------|
| `BlobUtils` | `isBlobSparkField`, `isBlobArrowField` | 12 |
| `LargeVarCharUtils` | `isLargeVarCharSparkField`, `isLargeVarCharArrowField`, `createPropertyKey` | 14 |
| `Utils` | `parseVersion`, `findVersion` | 11 |
| `VectorUtils` | `isVectorField`, `getVectorDimension`, `isVectorArrowField`, `getVectorArrowDimension`, `shouldBeFixedSizeList`, `createVectorSizePropertyKey` | 25 |

## Test plan
- [x] `mvn test -pl lance-spark-base_2.12 -Dtest="org.lance.spark.utils.*Test"` — 108 tests pass

### DIFF
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/utils/BlobUtilsTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/utils/BlobUtilsTest.java
@@ -31,25 +31,25 @@ import static org.junit.jupiter.api.Assertions.*;
 public class BlobUtilsTest {
 
   @Test
-  public void testIsBlobSparkField_nullField() {
+  public void testIsBlobSparkFieldNullField() {
     assertFalse(BlobUtils.isBlobSparkField(null));
   }
 
   @Test
-  public void testIsBlobSparkField_emptyMetadata() {
+  public void testIsBlobSparkFieldEmptyMetadata() {
     StructField field = DataTypes.createStructField("col", DataTypes.BinaryType, true);
     assertFalse(BlobUtils.isBlobSparkField(field));
   }
 
   @Test
-  public void testIsBlobSparkField_noKey() {
+  public void testIsBlobSparkFieldNoKey() {
     Metadata metadata = new MetadataBuilder().putString("other-key", "value").build();
     StructField field = new StructField("col", DataTypes.BinaryType, true, metadata);
     assertFalse(BlobUtils.isBlobSparkField(field));
   }
 
   @Test
-  public void testIsBlobSparkField_wrongValue() {
+  public void testIsBlobSparkFieldWrongValue() {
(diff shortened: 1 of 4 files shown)


hamersaw (commented):
It looks like the majority of these tests are essentially "x=10; assertTrue(isXTen(x))". It doesn't feel like there is a great deal of added value, the regressions would occur if upstream began setting `X` to something else, which we really can't test. I'm wondering if there's a better way to detect these kinds of issues.

LuciferYang:
@hamersaw Fair point — pushed 5b8bb14 trimming this:

- Dropped the `*Valid` tests from `BlobUtilsTest` and `LargeVarCharUtilsTest` — they were the literal "x=KEY; assertTrue(isKey(x))" tautology you called out. The `CaseInsensitive` cases already cover the positive path with non-tautological input (and would catch an accidental `.equals` regression).
- Replaced constant references in test inputs with string literals (`"lance-encoding:blob"`, `"arrow:large-var-char"`, `"arrow.fixed-size-list.size"`). A rename of the production constant now breaks the test — so the wire-format contract is pinned on the test side, which it wasn't before.
- Fixed `testCreateVectorSizePropertyKey` the same way (was concatenating the constant into the expected value).

58 tests total, all pass locally. Javadoc on each class now notes that upstream-convention drift is the integration layer's job — `BaseBlobCreateTableTest`, `SchemaConverterTest`, and the `LanceArrow*` suites already round-trip through real lance data and would catch that.

Let me know if this is the direction you had in mind, or if you'd rather drop the unit tests entirely and lean on the integration coverage.

hamersaw:
@LuciferYang really appreciate the PR. I'm thinking it probably makes sense to lean on the integration test coverage here.

LuciferYang:
Sounds good — I'll close this out and lean on the integration coverage. If a specific edge case ever slips past the integration suite, I'll add a targeted unit test for that gap rather than a broad sweep. Thanks for taking a look.

PR #420, opened 2026-04-10
feat: add byte-based batch flushing to prevent OOM on large rows

beinan:
## Summary

- Add `max_batch_bytes` write option (default 256MB) that flushes Arrow batches when byte size is reached, in addition to the existing row-count threshold (`batch_size`)
- Prevents OOM when writing tables with large rows (wide schemas, large binary/string columns, high-dimensional vector embeddings, etc.) where even a modest number of rows can exhaust memory
- Both `SemaphoreArrowBatchWriteBuffer` and `QueuedArrowBatchWriteBuffer` now use dedicated child allocators for accurate per-batch memory tracking

## Problem

The batch flush decision was purely row-count based (default `batch_size=8192`). For tables with large rows, a single batch could accumulate gigabytes of Arrow memory before flushing, causing OOM long before the row count threshold was reached.

## Solution

Batches now flush when **either** condition is met:
- Row count reaches `batch_size` (existing behavior, unchanged)
- Allocated memory reaches `max_batch_bytes` (new)

This is transparent — small-row workloads behave exactly as before since the row count limit is hit first. Large-row workloads get automatic memory protection.

### Usage

```python
df.write.format("lance") \
    .option("max_batch_bytes", "268435456") \  # 256MB (default)
    .save(path)
```

### Key implementation details

- **SemaphoreArrowBatchWriteBuffer**: now uses a child allocator (instead of the shared global one) so `getAllocatedMemory()` accurately reflects only this buffer's batch memory
- **QueuedArrowBatchWriteBuffer**: each batch gets its own child allocator via a `BatchEntry` wrapper, ensuring accurate byte tracking independent of concurrent consumer freeing of older batches
- **Backward compatible**: default `max_batch_bytes=256MB` with `Long.MAX_VALUE` meaning unlimited (matching previous behavior when not configured)

## Test plan

- [x] `testByteBasedFlush` — 100KB rows with 256KB byte limit → verifies batches flush with ~2-3 rows instead of waiting for row count
- [x] `testByteBasedFlushWithSmallRows` — small rows with high byte limit → verifies row-count flushing still works as before
- [x] `testByteBasedFlushSingleLargeRow` — each row exceeds byte limit → verifies 1 row per batch
- [x] All existing buffer tests pass (20 tests total)
- [x] `LanceDataWriterTest` end-to-end tests pass

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
@@ -56,6 +56,7 @@ public class LanceSparkWriteOptions implements Serializable {
   public static final String CONFIG_BATCH_SIZE = "batch_size";
   public static final String CONFIG_ENABLE_STABLE_ROW_IDS = "enable_stable_row_ids";
   public static final String CONFIG_USE_LARGE_VAR_TYPES = "use_large_var_types";
+  public static final String CONFIG_MAX_BATCH_BYTES = "max_batch_bytes";
 
   private static final WriteMode DEFAULT_WRITE_MODE = WriteMode.APPEND;
   private static final boolean DEFAULT_USE_QUEUED_WRITE_BUFFER = false;
@@ -63,6 +64,9 @@ public class LanceSparkWriteOptions implements Serializable {
   // Changed from 512 to 8192 for better write performance consistency with read path
   private static final int DEFAULT_BATCH_SIZE = 8192;
   private static final boolean DEFAULT_USE_LARGE_VAR_TYPES = false;
+  // Default max bytes per batch (256MB). Batches flush when either row count or byte size
+  // is reached, whichever comes first. This prevents OOM when rows are very large.
+  public static final long DEFAULT_MAX_BATCH_BYTES = 256L * 1024 * 1024;
 
   private final String datasetUri;
   private final WriteMode writeMode;
@@ -77,6 +81,7 @@ public class LanceSparkWriteOptions implements Serializable {
   // inherits from the manifest (e.g. append without re-specifying). Staged commit uses primitives.
   private final Boolean enableStableRowIds;
   private final boolean useLargeVarTypes;
+  private final long maxBatchBytes;
   private final Map<String, String> storageOptions;
 
   /** The namespace for credential vending. Transient as LanceNamespace is not serializable. */
@@ -100,6 +105,7 @@ public class LanceSparkWriteOptions implements Serializable {
     this.batchSize = builder.batchSize;
     this.enableStableRowIds = builder.enableStableRowIds;
     this.useLargeVarTypes = builder.useLargeVarTypes;
+    this.maxBatchBytes = builder.maxBatchBytes;
     this.storageOptions = new HashMap<>(builder.storageOptions);
     this.namespace = builder.namespace;
     this.tableId = builder.tableId;
@@ -179,6 +185,10 @@ public class LanceSparkWriteOptions implements Serializable {
     return useLargeVarTypes;
   }
 
+  public long getMaxBatchBytes() {
+    return maxBatchBytes;
+  }
+
   public Map<String, String> getStorageOptions() {
     return storageOptions;
   }
@@ -207,6 +217,9 @@ public class LanceSparkWriteOptions implements Serializable {
         .useQueuedWriteBuffer(useQueuedWriteBuffer)
         .queueDepth(queueDepth)
         .batchSize(batchSize)
+        .maxBatchBytes(maxBatchBytes)
+        .enableStableRowIds(enableStableRowIds)
+        .useLargeVarTypes(useLargeVarTypes)
         .storageOptions(storageOptions)
         .namespace(namespace)
         .tableId(tableId)
@@ -288,6 +301,7 @@ public class LanceSparkWriteOptions implements Serializable {
         && queueDepth == that.queueDepth
         && batchSize == that.batchSize
         && useLargeVarTypes == that.useLargeVarTypes
+        && maxBatchBytes == that.maxBatchBytes
         && Objects.equals(datasetUri, that.datasetUri)
         && writeMode == that.writeMode
         && Objects.equals(maxRowsPerFile, that.maxRowsPerFile)
@@ -314,6 +328,7 @@ public class LanceSparkWriteOptions implements Serializable {
         batchSize,
         enableStableRowIds,
         useLargeVarTypes,
+        maxBatchBytes,
         storageOptions,
         tableId,
         version);
@@ -332,6 +347,7 @@ public class LanceSparkWriteOptions implements Serializable {
     private int batchSize = DEFAULT_BATCH_SIZE;
     private Boolean enableStableRowIds;
     private boolean useLargeVarTypes = DEFAULT_USE_LARGE_VAR_TYPES;
+    private long maxBatchBytes = DEFAULT_MAX_BATCH_BYTES;
     private Map<String, String> storageOptions = new HashMap<>();
     private LanceNamespace namespace;
     private List<String> tableId;
@@ -391,6 +407,12 @@ public class LanceSparkWriteOptions implements Serializable {
 
     public Builder useLargeVarTypes(boolean useLargeVarTypes) {
       this.useLargeVarTypes = useLargeVarTypes;
+      return this;
+    }
+
+    public Builder maxBatchBytes(long maxBatchBytes) {
+      Preconditions.checkArgument(maxBatchBytes > 0, "maxBatchBytes must be positive");
+      this.maxBatchBytes = maxBatchBytes;
       return this;
     }
 
@@ -455,6 +477,11 @@ public class LanceSparkWriteOptions implements Serializable {
       }
       if (options.containsKey(CONFIG_USE_LARGE_VAR_TYPES)) {
         this.useLargeVarTypes = Boolean.parseBoolean(options.get(CONFIG_USE_LARGE_VAR_TYPES));
+      }
+      if (options.containsKey(CONFIG_MAX_BATCH_BYTES)) {
+        long parsedMaxBatchBytes = Long.parseLong(options.get(CONFIG_MAX_BATCH_BYTES));
+        Preconditions.checkArgument(parsedMaxBatchBytes > 0, "max_batch_bytes must be positive");
+        this.maxBatchBytes = parsedMaxBatchBytes;
       }
       return this;
     }
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
@@ -14,6 +14,7 @@
 package org.lance.spark.write;
 
 import org.lance.spark.LanceRuntime;
+import org.lance.spark.LanceSparkWriteOptions;
 
 import com.google.common.base.Preconditions;
 import org.apache.arrow.memory.BufferAllocator;
@@ -37,6 +38,16 @@ import java.util.concurrent.atomic.AtomicInteger;
  * simultaneously. This enables better pipelining between Spark row ingestion and Lance fragment
  * creation.
  *
+ * <p>Batches are flushed when either the row count reaches {@code batchSize} or the allocated
+ * memory for the current batch exceeds {@code maxBatchBytes}, whichever comes first. This limits
+ * the size of each individual batch when rows are very large (e.g., rows with large binary/string
+ * columns or embeddings).
+ *
+ * <p>Because this implementation is queue-based, multiple completed batches can be buffered at the
+ * same time. As a result, total in-flight Arrow memory can be roughly {@code queueDepth *
+ * maxBatchBytes}, plus the current producer batch and allocator overhead. Users may need to tune
+ * {@code queueDepth} and {@code maxBatchBytes} together to stay within memory limits.
+ *
  * <p>Architecture:
  *
  * <pre>
@@ -53,16 +64,23 @@ public class QueuedArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
   private final Schema schema;
   private final StructType sparkSchema;
   private final int batchSize;
+  private final long maxBatchBytes;
   private final int queueDepth;
 
-  /** Queue holding completed batches ready for consumption. */
-  private final BlockingQueue<VectorSchemaRoot> batchQueue;
+  /**
+   * Queue holding completed batches ready for consumption. Each entry pairs a VectorSchemaRoot with
+   * its dedicated child allocator, so the consumer can free both when done.
+   */
+  private final BlockingQueue<BatchEntry> batchQueue;
 
   /** Child allocator for producer batches (shares root with consumer for buffer transfer). */
   private final BufferAllocator producerAllocator;
 
   /** Current batch being filled by producer. */
   private VectorSchemaRoot currentBatch;
+
+  /** Child allocator dedicated to the current batch for accurate byte tracking. */
+  private BufferAllocator currentBatchAllocator;
 
   /** Arrow writer for current batch. */
   private org.lance.spark.arrow.LanceArrowWriter currentArrowWriter;
@@ -77,11 +95,36 @@ public class QueuedArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
   private volatile boolean consumerFinished = false;
 
   /** Current batch being read by consumer (for ArrowReader interface). */
-  private VectorSchemaRoot consumerCurrentBatch;
+  private BatchEntry consumerCurrentEntry;
+
+  /** Pairs a VectorSchemaRoot with its allocator for proper lifecycle management. */
+  private static class BatchEntry {
+    final VectorSchemaRoot batch;
+    final BufferAllocator allocator;
+
+    BatchEntry(VectorSchemaRoot batch, BufferAllocator allocator) {
+      this.batch = batch;
+      this.allocator = allocator;
+    }
+
+    void close() {
+      try {
+        batch.close();
+      } finally {
+        allocator.close();
+      }
+    }
+  }
 
   public QueuedArrowBatchWriteBuffer(
       BufferAllocator allocator, Schema schema, StructType sparkSchema, int batchSize) {
-    this(allocator, schema, sparkSchema, batchSize, DEFAULT_QUEUE_DEPTH);
+    this(
+        allocator,
+        schema,
+        sparkSchema,
+        batchSize,
+        DEFAULT_QUEUE_DEPTH,
+        LanceSparkWriteOptions.DEFAULT_MAX_BATCH_BYTES);
   }
 
   /** Simplified constructor that uses LanceRuntime allocator and converts Spark schema to Arrow. */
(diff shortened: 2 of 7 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -127,16 +173,27 @@ public QueuedArrowBatchWriteBuffer(
 
   /** Allocates a new batch for the producer to fill. */
   private void allocateNewBatch() {
-    currentBatch = VectorSchemaRoot.create(schema, producerAllocator);
+    // Each batch gets its own child allocator so getAllocatedMemory() accurately reflects
+    // only this batch's memory, unaffected by concurrent consumer freeing of older batches.
+    currentBatchAllocator = producerAllocator.newChildAllocator("batch", 0, Long.MAX_VALUE);
+    currentBatch = VectorSchemaRoot.create(schema, currentBatchAllocator);
     currentBatch.allocateNew();
```
hamersaw: If either of these error the `currentBatchAllocator` resource is leaked, can we close on failure?

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -77,26 +88,49 @@ public class QueuedArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
...
+
+    BatchEntry(VectorSchemaRoot batch, BufferAllocator allocator) {
+      this.batch = batch;
+      this.allocator = allocator;
+    }
+
+    void close() {
+      batch.close();
+      allocator.close();
+    }
```
hamersaw, line 116: If `batch.close()` throws an Exception then `allocator.close()` is never called.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
```diff
@@ -87,6 +123,19 @@ public void onTaskComplete() {
     }
   }
 
+  /** Returns whether the current batch should be flushed based on byte size. */
+  private boolean isBatchFullByBytes() {
+    if (maxBatchBytes == Long.MAX_VALUE) {
+      return false;
+    }
+    return batchAllocator.getAllocatedMemory() >= maxBatchBytes;
```
hamersaw: I think the SemaphoreArrowBatchWriteBuffer uses a single allocator between producer / consumer. This means that flushing batches may not be precisely what the user set. I'm wondering if it makes sense to move to a per-batch allocator like the QueuedArrowBatchWriter approach?

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -77,26 +88,49 @@ public class QueuedArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
...
   public QueuedArrowBatchWriteBuffer(
       BufferAllocator allocator, Schema schema, StructType sparkSchema, int batchSize) {
-    this(allocator, schema, sparkSchema, batchSize, DEFAULT_QUEUE_DEPTH);
+    this(allocator, schema, sparkSchema, batchSize, DEFAULT_QUEUE_DEPTH, Long.MAX_VALUE);
   }
 
   /** Simplified constructor that uses LanceRuntime allocator and converts Spark schema to Arrow. */
   public QueuedArrowBatchWriteBuffer(StructType sparkSchema, int batchSize) {
-    this(sparkSchema, batchSize, DEFAULT_QUEUE_DEPTH);
+    this(sparkSchema, batchSize, DEFAULT_QUEUE_DEPTH, Long.MAX_VALUE);
```
hamersaw: I think the default `max_batch_size` story needs to be ironed out a bit. The PR indicates the default is 256MB, but if we build the QueuedArrowBatchWriteBuffer using this constructor, then it is Long.MAX_VALUE and consequently disabled. I think we should:
(1) Use the default const here instead of Long.MAX_VALUE
(2) Decide if we want default to be 256MB, or if default is Long.MAX_VALUE to disable

hamersaw (commented):
Overall looks great! Think we might need to tighten up the UX so this setting does not surprise anyone.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
beinan: Good observation. The SemaphoreBuffer is different from the QueuedBuffer in that it reuses a single `VectorSchemaRoot` across batches (created once by `ArrowReader.getVectorSchemaRoot()`, then `reset()` between batches). Since `reset()` doesn't free the underlying `ArrowBuf` allocations, the child allocator retains peak capacity from the largest batch seen.

This means the byte tracking is **conservative** — after a large batch, subsequent batches may flush earlier than the configured limit because the allocator still reports the old capacity. This is safe (prevents memory growth), but not as precise as the QueuedBuffer's per-batch allocator approach.

In practice this is fine for the Semaphore buffer since it processes one batch at a time — there's at most one batch worth of memory at any point. The conservatism just means smaller batches after a big one, which is acceptable.

If we wanted precise per-batch tracking, we'd need to change the ArrowReader contract to create a new VectorSchemaRoot per batch (with its own allocator), which would be a larger refactor. Happy to do it if you think it's worth it, but I'd lean toward keeping the current approach since the memory bound is correct even if batch sizes aren't perfectly uniform.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
beinan: Fixed — added try/catch in `allocateNewBatch()` to close the child allocator if `VectorSchemaRoot.create()` or `allocateNew()` fails.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
beinan: Fixed — `BatchEntry.close()` now uses try/finally so `allocator.close()` runs even if `batch.close()` throws.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
beinan: Good catch. Made `DEFAULT_MAX_BATCH_BYTES` public in `LanceSparkWriteOptions` and updated all buffer constructors to reference it instead of `Long.MAX_VALUE`. Default is 256MB everywhere now.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
```diff
@@ -378,6 +393,11 @@ public Builder enableStableRowIds(Boolean enableStableRowIds) {
       return this;
     }
 
+    public Builder maxBatchBytes(long maxBatchBytes) {
```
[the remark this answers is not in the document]
beinan: Done — added `Preconditions.checkArgument(maxBatchBytes > 0)` in the builder method.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/SemaphoreArrowBatchWriteBufferTest.java
```diff
@@ -254,4 +256,185 @@ public void testWriteErrorPropagation() throws Exception {
...
+      Thread writerThread =
+          new Thread(
+              () -> {
+                for (int i = 0; i < totalRows; i++) {
+                  UTF8String largeValue = generateLargeString(rowSizeBytes);
+                  InternalRow row = new GenericInternalRow(new Object[] {largeValue});
+                  writeBuffer.write(row);
+                  rowsWritten.incrementAndGet();
+                }
+                writeBuffer.setFinished();
```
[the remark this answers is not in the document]
beinan: Fixed — moved `setFinished()` to `finally` blocks in all test writer threads (both Semaphore and Queued buffer tests).

lance-spark-base_2.12/src/test/java/org/lance/spark/write/SemaphoreArrowBatchWriteBufferTest.java
```diff
@@ -254,4 +256,185 @@ public void testWriteErrorPropagation() throws Exception {
...
+              });
```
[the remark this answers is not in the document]
beinan: Fixed in same commit.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/QueuedArrowBatchWriteBufferTest.java
```diff
@@ -580,4 +581,249 @@ public void testWriteErrorPropagation() throws Exception {
...
+              () -> {
+                try {
+                  for (int i = 0; i < totalRows; i++) {
+                    UTF8String largeValue = generateLargeString(rowSizeBytes);
+                    InternalRow row = new GenericInternalRow(new Object[] {largeValue});
+                    writeBuffer.write(row);
+                    rowsWritten.incrementAndGet();
+                  }
+                  writeBuffer.setFinished();
+                } catch (Throwable e) {
```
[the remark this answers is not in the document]
beinan: Fixed in same commit.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/QueuedArrowBatchWriteBufferTest.java
```diff
@@ -580,4 +581,249 @@ public void testWriteErrorPropagation() throws Exception {
...
+                try {
+                  for (int i = 0; i < totalRows; i++) {
+                    InternalRow row =
+                        new GenericInternalRow(new Object[] {rowsWritten.incrementAndGet()});
+                    writeBuffer.write(row);
+                  }
+                  writeBuffer.setFinished();
+                } catch (Throwable e) {
+                  writerError.set(e);
+                  e.printStackTrace();
```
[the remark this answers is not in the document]
beinan: Fixed in same commit.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/QueuedArrowBatchWriteBufferTest.java
```diff
@@ -580,4 +581,249 @@ public void testWriteErrorPropagation() throws Exception {
...
+                  for (int i = 0; i < totalRows; i++) {
+                    UTF8String largeValue = generateLargeString(rowSizeBytes);
+                    InternalRow row = new GenericInternalRow(new Object[] {largeValue});
+                    writeBuffer.write(row);
+                    rowsWritten.incrementAndGet();
+                  }
+                  writeBuffer.setFinished();
+                } catch (Throwable e) {
+                  writerError.set(e);
+                  e.printStackTrace();
```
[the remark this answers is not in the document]
beinan: Fixed in same commit.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
```diff
@@ -55,25 +67,49 @@ public class SemaphoreArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
...
-      BufferAllocator allocator, Schema schema, StructType sparkSchema, int batchSize) {
-    super(allocator);
+      BufferAllocator allocator,
+      Schema schema,
+      StructType sparkSchema,
+      int batchSize,
+      long maxBatchBytes) {
+    // Pass a child allocator to ArrowReader so VectorSchemaRoot allocation is tracked
+    super(allocator.newChildAllocator("semaphore-buffer", 0, Long.MAX_VALUE));
     Preconditions.checkNotNull(schema);
```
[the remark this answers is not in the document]
beinan: Fixed — `closeReadSource()` now closes `batchAllocator`.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -37,6 +37,10 @@
  * simultaneously. This enables better pipelining between Spark row ingestion and Lance fragment
  * creation.
  *
+ * <p>Batches are flushed when either the row count reaches {@code batchSize} or the allocated
+ * memory exceeds {@code maxBatchBytes}, whichever comes first. This prevents OOM when individual
+ * rows are very large (e.g., rows with large binary/string columns or embeddings).
+ *
```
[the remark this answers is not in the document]
beinan: Done — updated the javadoc to document that total in-flight memory can be roughly `queueDepth * maxBatchBytes`.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
```diff
@@ -55,18 +73,31 @@ public class SemaphoreArrowBatchWriteBuffer extends ArrowBatchWriteBuffer {
...
     Preconditions.checkArgument(batchSize > 0);
+    Preconditions.checkArgument(maxBatchBytes > 0, "maxBatchBytes must be positive");
     this.schema = schema;
     this.sparkSchema = sparkSchema;
     this.batchSize = batchSize;
+    this.maxBatchBytes = maxBatchBytes;
+    this.batchAllocator = this.allocator;
```
hamersaw: This seems like a redundant alias?

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -91,18 +134,34 @@ public QueuedArrowBatchWriteBuffer(StructType sparkSchema, int batchSize) {
...
+    this(
+        sparkSchema,
+        batchSize,
+        queueDepth,
+        useLargeVarTypes,
+        LanceSparkWriteOptions.DEFAULT_MAX_BATCH_BYTES);
+  }
+
+  /** Simplified constructor with all tuning parameters. */
+  public QueuedArrowBatchWriteBuffer(
```
hamersaw, line 147: There are now like 6 constructors for QueuedArrowBatchWriter (similar for SemaphoreArrowBatchWriter). Does it make sense to consolidate these? Or update to a builder pattern?

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
```diff
@@ -56,13 +56,17 @@ public class LanceSparkWriteOptions implements Serializable {
   public static final String CONFIG_BATCH_SIZE = "batch_size";
   public static final String CONFIG_ENABLE_STABLE_ROW_IDS = "enable_stable_row_ids";
   public static final String CONFIG_USE_LARGE_VAR_TYPES = "use_large_var_types";
+  public static final String CONFIG_MAX_BATCH_BYTES = "max_batch_bytes";
```
hamersaw, line 59: Can we add this configuration option to the docs?

lance-spark-base_2.12/src/test/java/org/lance/spark/write/SemaphoreArrowBatchWriteBufferTest.java
```diff
@@ -254,4 +264,56 @@ public void testWriteErrorPropagation() throws Exception {
...
+
+  /** Generate a string of approximately the given size in bytes. */
+  private UTF8String generateLargeString(int sizeBytes) {
+    byte[] data = new byte[sizeBytes];
+    java.util.Arrays.fill(data, (byte) 'A');
+    return UTF8String.fromBytes(data);
+  }
+
+  @Test
+  public void testByteBasedFlushWithSmallRows() throws Exception {
```
hamersaw, line 292: It looks like testing coverage on QueuedArrowBatchWriteBuffer contains a few more functions than this one. Can we get parity?

hamersaw (commented):
Appreciate the iteration. Just a few small concerns and then I think we'll be good to merge!

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
beinan: Done — removed the `batchAllocator` field, using `this.allocator` (inherited from ArrowReader) directly.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
beinan: Consolidated — removed unused intermediate constructors from both buffer classes. Each now has: a canonical full-args constructor, a test-friendly 4-arg variant, and the minimal constructors needed by SparkPositionDeltaWrite and LanceDataWriter.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
beinan: Done — added `max_batch_bytes` documentation to `docs/src/performance.md` under the Write Performance section, including the note about `queue_depth * max_batch_bytes` interaction for queued buffers.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/SemaphoreArrowBatchWriteBufferTest.java
beinan: I investigated adding the large-row byte-based flush tests (`testByteBasedFlush` and `testByteBasedFlushSingleLargeRow`) to the SemaphoreBuffer. Unfortunately, they're inherently flaky (~30% failure rate) due to the shared-allocator architecture: when the allocator retains peak memory after `reset()`, every row triggers a flush (1 row/batch), and the tight producer-consumer loop races with `createTrackedTask`'s `onTaskComplete()` callback. The QueuedBuffer avoids this because each batch gets its own allocator.

The existing `testByteBasedFlushWithSmallRows` verifies that `maxBatchBytes` doesn't interfere with normal row-count flushing. The large-row scenarios are thoroughly covered by the QueuedBuffer tests. I also added a defensive `VectorSchemaRoot.clear()` in `loadNextBatch()`'s early-return path to prevent memory leaks.

hamersaw: This seems like a bug right? If the allocator retaining memory means that every additional row insert will trigger a single row flush, that kind of breaks everything here.

hamersaw: cc @beinan has this been resolved?

beinan: Done — added `testByteBasedFlush` and `testByteBasedFlushSingleLargeRow` to match QueuedArrowBatchWriteBuffer coverage.

Also fixed a real bug this exposed: the SemaphoreBuffer was using absolute allocator memory for byte tracking, but Arrow's `allocateNew()` pre-allocates capacity based on previous batch sizes. After the first large batch, all subsequent writes fit within pre-allocated capacity (delta = 0), so byte-based flushing never triggered again — every row after the first batch accumulated without flushing.

Fix: track per-batch bytes via before/after allocator snapshots around each `write()` call, and reset vectors with `setInitialCapacity(1)` between batches to prevent Arrow from pre-allocating based on historical sizes.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
beinan: Done — added `max_batch_bytes` documentation to `docs/src/performance.md` under the Write Performance section, including the note about `queue_depth * max_batch_bytes` interaction for queued buffers.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/SemaphoreArrowBatchWriteBuffer.java
beinan: Update: this has been fixed in the latest commit. The SemaphoreBuffer now tracks per-batch bytes accurately by snapshotting allocator memory before/after each write() call and accumulating the delta. Between batches, vectors are reset with setInitialCapacity(1) to prevent Arrow from pre-allocating based on previous batch sizes. This eliminates the conservative behavior — batch sizes are now consistent. Tests testByteBasedFlush and testByteBasedFlushSingleLargeRow verify this.

beinan:
> Appreciate the iteration. Just a few small concerns and then I think we'll be good to merge!

all comments addressed, could you take a look again?  many thanks!

hamersaw (approved):
Thanks for the iteration! My only concern is on the large number of allocations required for both the Semaphore and Queued variants. I ran a benchmark with 1m rows, 10 columns, and batch size of 1024. The semaphore required 107,470 allocations and the queued implementation 9,770.

I don't think we need to gate merging this on the potential issue, but I wrote a different "pooled" approach that reduces this to just 40 allocations. I'm going to do some more testing and open up a PR shortly.

PR #425, opened 2026-04-12
fix: report post-pruning statistics to enable BroadcastHashJoin with SPJ

beinan:
## Summary
- **Fix SPJ + BroadcastHashJoin incompatibility**: When SPJ is active and one side of a join becomes tiny after filter pushdown, Spark's AQE cannot fall back to BroadcastHashJoin because SPJ removes all shuffle Exchange nodes (no materialization points for AQE to re-evaluate). The root cause is that `LanceStatistics` always reported full-table size from `ManifestSummary`, so `JoinSelection` always picked SortMergeJoin for large tables — which then got locked into SPJ.
- **Estimate post-pruning statistics at build time**: In `LanceScanBuilder.build()`, run `ZonemapFragmentPruner.pruneFragments()` against pushed filters and zonemap stats to get the surviving fragment count, then scale full-table stats by `survivingFragments / totalFragments`. When the pruned size falls below Spark's `autoBroadcastJoinThreshold` (default 10MB), `JoinSelection` picks BroadcastHashJoin *before* `EnsureRequirements` considers SPJ.

### How Spark's planning order makes this work
1. `JoinSelection` checks `estimateStatistics().sizeInBytes()` → picks BroadcastHashJoin (if small) or SortMergeJoin (if large)
2. `EnsureRequirements` checks `outputPartitioning()` → only applies SPJ to SortMergeJoin
3. AQE cannot override SPJ→broadcast because SPJ removes all Exchange nodes

So accurate post-pruning statistics let JoinSelection pick broadcast before SPJ is ever considered. SPJ still activates for large joins where both sides are big.

### Expected benchmark impact
From the PR #396 benchmark: filtered join (`region = 'east'`) was 31.4s with SPJ vs 16.2s without. With this fix, the filtered join should use BroadcastHashJoin (matching the 16.2s baseline) while the unfiltered full inner join continues to use SPJ (27.3s, 12.1x speedup).

## Test plan
- [x] 7 new unit tests for `LanceStatistics.estimatePostPruning()` covering: scaling arithmetic, all-survive, none-survive, zero-total-fragments, more-surviving-than-total
- [x] All 258 existing tests pass
- [x] Spotless clean

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
@@ -138,7 +138,6 @@ public class LanceScanBuilder
 
     // Get statistics from manifest summary before closing dataset
     ManifestSummary summary = getOrOpenDataset().getVersion().getManifestSummary();
-    LanceStatistics statistics = new LanceStatistics(summary);
 
     // Collect all columns that need zonemap stats: filter columns + partition column (if declared).
     Set<String> columnsToLoad = extractReferencedColumns(pushedFilters);
@@ -176,6 +175,37 @@ public class LanceScanBuilder
               partValues.get().size());
         }
       }
+    }
+
+    // Estimate post-pruning statistics so Spark's JoinSelection can pick BroadcastHashJoin
+    // when the pruned size is below the broadcast threshold. Without this, the full-table size
+    // causes JoinSelection to pick SortMergeJoin, which then gets locked into SPJ — preventing
+    // AQE from switching to broadcast even when one side is tiny after filter pushdown.
+    LanceStatistics statistics;
+    if (pushedFilters.length > 0 && !zonemapStats.isEmpty()) {
+      java.util.Optional<Set<Integer>> survivingIds =
+          ZonemapFragmentPruner.pruneFragments(pushedFilters, zonemapStats);
+      if (survivingIds.isPresent()) {
+        statistics =
+            LanceStatistics.estimatePostPruning(
+                summary.getTotalRows(),
+                summary.getTotalFilesSize(),
+                summary.getTotalFragments(),
+                survivingIds.get().size());
+        LOG.info(
+            "Estimated post-pruning statistics: {} of {} fragments survive,"
+                + " estimatedSize={}, estimatedRows={} (full: size={}, rows={})",
+            survivingIds.get().size(),
+            summary.getTotalFragments(),
+            statistics.sizeInBytes(),
+            statistics.numRows(),
+            summary.getTotalFilesSize(),
+            summary.getTotalRows());
+      } else {
+        statistics = new LanceStatistics(summary);
+      }
+    } else {
+      statistics = new LanceStatistics(summary);
     }
 
     // Close the lazily opened dataset - it's no longer needed after build
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
@@ -164,46 +164,45 @@ public class LanceScanBuilder
             partitionColumn,
             TABLE_OPT_PARTITION_COLUMNS);
       } else {
-        java.util.Optional<Map<Integer, Comparable<?>>> partValues =
-            ZonemapFragmentPruner.computeFragmentPartitionValues(zonemapStats.get(partitionColumn));
-        if (partValues.isPresent()) {
-          partitionInfo =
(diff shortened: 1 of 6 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -178,6 +177,37 @@ public Scan build() {
       }
     }
 
+    // Estimate post-pruning statistics so Spark's JoinSelection can pick BroadcastHashJoin
+    // when the pruned size is below the broadcast threshold. Without this, the full-table size
+    // causes JoinSelection to pick SortMergeJoin, which then gets locked into SPJ — preventing
+    // AQE from switching to broadcast even when one side is tiny after filter pushdown.
+    LanceStatistics statistics;
+    if (pushedFilters.length > 0 && !zonemapStats.isEmpty()) {
+      java.util.Optional<Set<Integer>> survivingIds =
```
hamersaw: Can we move `java.util.Optional` as a top-level import.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -178,6 +177,37 @@ public Scan build() {
...
+          ZonemapFragmentPruner.pruneFragments(pushedFilters, zonemapStats);
+      if (survivingIds.isPresent()) {
+        statistics =
+            LanceStatistics.estimatePostPruning(
+                summary.getTotalRows(),
+                summary.getTotalFilesSize(),
+                summary.getTotalFragments(),
+                survivingIds.get().size());
+        LOG.info(
```
hamersaw: `debug` level here? I don't think we need to info log every query that has filters.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -178,6 +177,37 @@ public Scan build() {
...
     }
 
+    // Estimate post-pruning statistics so Spark's JoinSelection can pick BroadcastHashJoin
+    // when the pruned size is below the broadcast threshold. Without this, the full-table size
+    // causes JoinSelection to pick SortMergeJoin, which then gets locked into SPJ — preventing
+    // AQE from switching to broadcast even when one side is tiny after filter pushdown.
+    LanceStatistics statistics;
+    if (pushedFilters.length > 0 && !zonemapStats.isEmpty()) {
+      java.util.Optional<Set<Integer>> survivingIds =
+          ZonemapFragmentPruner.pruneFragments(pushedFilters, zonemapStats);
```
hamersaw: We would now be calling this twice within the same query execution (here + LanceScan). I don't think it's necessarily expensive, but wondering if it makes sense to cache the results and pass to the `LanceScan`?

hamersaw (commented):
Great solution! A few nits for repo, wondering if we need to modify the AGENTS.md file so your AI coder picks it up?

LuciferYang:
Compilation failed
```
[INFO] compiling 2 Scala sources and 101 Java sources to /home/runner/work/lance-spark/lance-spark/lance-spark-4.1_2.13/target/test-classes ...
Error:  /home/runner/work/lance-spark/lance-spark/lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseSpjBroadcastTest.java:81:17:  error: unreported exception IOException; must be caught or declared to be thrown
```

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Done — eliminated all inline `java.util.Optional` FQN usages by unwrapping with `.orElse(null)` and using nullable types. This avoids the naming conflict with `org.lance.spark.utils.Optional`.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Done — changed to `LOG.debug` for both the post-pruning stats log and the "loaded zonemap stats" log.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Good call — now pre-computing `survivingFragmentIds` in `LanceScanBuilder.build()` and passing it to `LanceScan` via a new constructor parameter. `pruneByZonemapStats()` uses the cached result when available.

hamersaw (approved):
Looks great! Excited to see perf analysis.

PR #426, opened 2026-04-12
fix: add clean-bundle target for reliable source change detection

ivscheianu:
Fix for https://github.com/lance-format/lance-spark/issues/374

`make bundle` uses incremental compilation which can silently serve stale artifacts when source changes occur across modules. The `scala-maven-plugin` upgrade (https://github.com/lance-format/lance-spark/pull/388) improved within-module Zinc compilation but cross-module staleness via the Maven `.m2` cache still occurs.


hamersaw (approved):
Looks great! Thanks.

PR #427, opened 2026-04-13
fix: implement equals/hashCode on LanceScan to enable ReusedExchange

LuciferYang:
## Summary

Implement `equals()`/`hashCode()` on `LanceScan` so that Spark's `ReusedExchange` optimization takes effect when the same Lance table is scanned multiple times in a query (e.g., self-join).

## Problem

`BatchScanExec.equals()` compares `batch` objects to determine if two scans are equivalent. Since `LanceScan` implements `Batch` and did not override `equals()`, it used `Object` reference equality — two scans of the same table were never equal, and exchanges were never reused.

## Solution

Override `equals()`/`hashCode()` on `LanceScan`, comparing the fields that define scan identity: `schema`, `readOptions`, `whereConditions`, `limit`, `offset`, `topNSortOrders`, `pushedAggregation`, and `pushedFilters`.

Comparison strategies follow Spark and Iceberg conventions:
- `pushedFilters`: order-independent via sort-by-hashCode, per `FileScan.equivalentFilters()`
- `pushedAggregation`: sort array components by hashCode, per `AggregatePushDownUtils.equivalentAggregations()`
- `topNSortOrders`: `toString()` comparison since `ColumnOrdering` lacks `equals()`, per Iceberg's filter expression approach

## Test plan

- [x] Unit: identical scans are equal; different filters/schema are not; hashCode consistent
- [x] E2E (Spark 3.4 & 3.5): self-join and 3-way self-join produce `ReusedExchange` referencing existing Exchange nodes
- [x] E2E: self-join with different filters does NOT produce `ReusedExchange`
- [x] All 255 existing base module tests pass

### DIFF
--- a/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/ReusedExchangeTest.java
+++ b/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/ReusedExchangeTest.java
@@ -1,0 +1,16 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


LuciferYang:
cc @hamersaw

hamersaw (approved):
Looks great! Thanks for the boost.

LuciferYang:
Thank you @hamersaw

Issue #36, opened 2025-07-07
Support config compression through Spark


majin1102:
Typically should be like:

create table
xxxxx

properties(
xxx.xxxx.xxxx= ''
)

or
through FieldConfig

hamersaw:
Completed in https://github.com/lance-format/lance-spark/pull/428

PR #428, opened 2026-04-13
feat: support compression config via Spark TBLPROPERTIES

ivscheianu:
## Summary
Fix for https://github.com/lance-format/lance-spark/issues/36

Lance's Rust encoder reads `lance-encoding:<key>` Arrow field metadata to configure per-column compression and encoding at write time. Until now, there was no way to set these from Spark: users could not control whether columns were compressed with zstd, lz4, or fsst, nor tune structural encoding or byte stream split behavior.

This PR closes that gap by extending the existing `SchemaConverter` pipeline to map `<column>.lance.<key>` TBLPROPERTIES to Arrow field metadata, following the same pattern already used for blob encoding, float16, and large varchar.

## What's supported

Five keys are wired in this release:

| TBLPROPERTIES key | Arrow field metadata key | Valid values |
|---|---|---|
| `<col>.lance.compression` | `lance-encoding:compression` | `zstd`, `lz4`, `fsst`, `none` |
| `<col>.lance.compression-level` | `lance-encoding:compression-level` | integer string |
| `<col>.lance.structural-encoding` | `lance-encoding:structural-encoding` | `miniblock`, `fullzip` |
| `<col>.lance.rle-threshold` | `lance-encoding:rle-threshold` | float in `(0.0, 1.0]` |
| `<col>.lance.bss` | `lance-encoding:bss` | `off`, `on`, `auto` |

Invalid values fail fast at `CREATE TABLE` time with a clear `IllegalArgumentException`. Unknown keys and unmatched column names are silently ignored, consistent with existing `SchemaConverter` behavior.

## Why only these five keys

The remaining lance-core encoding keys (`dict-divisor`, `dict-size-ratio`, `dict-values-compression`, `dict-values-compression-level`, `minichunk-size`) are deferred deliberately:

- **Dictionary keys**: dictionary encoding in lance-core is currently auto-selected based on data characteristics. Exposing manual overrides before the defaults are stable risks locking users into configurations that become suboptimal as the encoder evolves.
- **minichunk-size**: a low-level tuning parameter with non-obvious interactions with block size  and compression. Needs more usage data before surfacing in the connector API.

These keys are silently ignored if present, so they can be wired in a follow-up without a breaking change.

## Note on ALTER TABLE

`ALTER TABLE ... SET TBLPROPERTIES` does **not** change the encoding of previously written data. Compression properties take effect only when the table is created.

### DIFF
--- a/docker/Dockerfile.test-base
+++ b/docker/Dockerfile.test-base
@@ -22,7 +22,7 @@ RUN curl -fL -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linu
 RUN python3 -m venv /opt/venv && \
     . /opt/venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install pytest pytest-timeout packaging azure-storage-blob boto3 lance
+    pip install pytest pytest-timeout packaging azure-storage-blob boto3 pylance
 
 ENV PATH="/opt/venv/bin:$PATH"
 ENV SPARK_HOME="/opt/spark"
--- a/docker/Dockerfile.test-base
+++ b/docker/Dockerfile.test-base
@@ -22,7 +22,7 @@ RUN curl -fL -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linu
 RUN python3 -m venv /opt/venv && \
     . /opt/venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install pytest pytest-timeout packaging azure-storage-blob boto3 pylance
+    pip install pytest pytest-timeout packaging azure-storage-blob boto3
 
 ENV PATH="/opt/venv/bin:$PATH"
 ENV SPARK_HOME="/opt/spark"
--- a/docker/Dockerfile.test-base
+++ b/docker/Dockerfile.test-base
@@ -22,7 +22,7 @@ RUN curl -fL -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linu
 RUN python3 -m venv /opt/venv && \
     . /opt/venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install pytest pytest-timeout packaging azure-storage-blob boto3
+    pip install pytest pytest-timeout packaging azure-storage-blob boto3 lance
 
 ENV PATH="/opt/venv/bin:$PATH"
 ENV SPARK_HOME="/opt/spark"
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
@@ -100,6 +100,7 @@ public final class LanceEncodingUtils {
     }
   }
 
+  // codec-specific upper-bound enforcement is left to the Rust encoder
   private static void validateCompressionLevel(String columnName, String value) {
     int level;
     try {
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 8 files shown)


hamersaw (commented):
Thanks for the PR! Overall I think this looks great. Kind of a more meta question I was interested in your thoughts on. It feels like the current column encoding scheme, putting the column name first, can have couple of issues. First, it may introduce ambiguity if a column is named the same as a different keywork, and second, how do we set these fields on nested datatypes (ex. struct fields, etc). I'm wonder if rather than prefixing the column, we should postfix it? So in this case, rather then:
```
'payload.lance.compression'       = 'zstd'
'payload.lance.compression-level' = '3'
'ts.lance.structural-encoding'    = 'miniblock'
'ts.lance.rle-threshold'          = '0.5'
'ts.lance.bss'                    = 'auto'
```
we would do something like:
```
'lance.compression.column.payload'       = 'zstd'
'lance.compression-level.column.payload' = '3'
'lance.structural-encoding.column.ts'    = 'miniblock'
'lance.rle-threshold.column.ts'          = '0.5'
'lance.bss.column.ts'                    = 'auto'
```
This would remove the potential ambiguity and allow us to set specifically on nested fields by using mulitple identifiers after the `column` naming.

If this is the direction we decide to go, we can make the change backward compatible. For example, the existing options will respect both formats, but moving forward we should only support the new format.

ivscheianu:
Thanks for the feedback, @hamersaw! Good points, I actually think we might need to consider a wider redesign here.

A few trade-offs I noticed with the postfix format though:
-  Column-first keeps all config for a column grouped together (payload.lance.*), which is nice when you're scanning a multi-property declaration. Key-first scatters them across different prefixes.
- All existing column properties already use column-first (embeddings.arrow.fixed-size-list.size, content.lance.encoding, content.arrow.large_var_char, etc). Switching formats just for compression might create a split in the public API. If we go this route, it should probably apply to everything.

I'd suggest to keep this PR on the established column-first format so it stays consistent with what's already shipped, and I can open a follow-up issue to design a connector-wide namespace migration (I'll also try to take care of the impl). The nested-field problem affects vectors, blobs, and large varchar just as much as compression too, so it feels like it deserves its own scoped discussion rather than being solved here. Maybe something like `lance.column.<col_path>.<key>` with a backward-compat transition window could work well across the board?

I don’t have any hard feelings toward any of them, to be honest. I'm happy to go with your suggestion if you think it works for now.

hamersaw:
@ivscheianu sounds great to me. I created a [new ticket here](https://github.com/lance-format/lance-spark/issues/434) to track the update.

docker/Dockerfile.test-base
```diff
@@ -22,7 +22,7 @@ RUN curl -fL -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linu
 RUN python3 -m venv /opt/venv && \
     . /opt/venv/bin/activate && \
     pip install --upgrade pip && \
-    pip install pytest pytest-timeout packaging azure-storage-blob boto3
+    pip install pytest pytest-timeout packaging azure-storage-blob boto3 pylance
```
hamersaw: We shouldn't need to install `pylance` here I think.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
```diff
@@ -0,0 +1,198 @@
...
+              + "'. Valid values: "
+              + VALID_COMPRESSION_SCHEMES);
+    }
+  }
+
+  // Only integer parseability is checked; codec-specific range enforcement is left to the Rust
+  // encoder.
+  private static void validateCompressionLevel(String columnName, String value) {
+    try {
+      Integer.parseInt(value);
```
hamersaw: Should we be validating this further? ex. non-negative

hamersaw (commented):
Thanks for the iteration on this. I think a few remaining nits and then we should be good!

docker/Dockerfile.test-base
ivscheianu: good catch, I missed removing it after some troubleshooting

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LanceEncodingUtils.java
ivscheianu: added non-negative validation, the rust core already handles the rest

ivscheianu:
> Thanks for the iteration on this. I think a few remaining nits and then we should be good!

thanks @hamersaw, addressed the comments. Ready for review when you get a chance. Thanks a lot!

hamersaw (approved):
Looks great! Thanks for the PR.

PR #439, opened 2026-04-16
feat: add param `rows_per_range` for range-based btree index built

fangbo:
# Background

Currently the partition(range) number is configured by spark parameter like: `spark.sql.adaptive.coalescePartitions.initialPartitionNum` or `spark.sql.shuffle.partitions` when building btree index using range-mode.

The current approach cannot dynamically adjust the number of ranges based on changes in the total row count of the Dataset. This becomes quite inconvenient when the total row count of the Dataset continues to grow.

# Design

So, we add a new parameter `rows_per_range` for range-mode. This param specifies the row number for each range. The spark partition number is calculated by `Dataset.total_rows/rows_per_range`. This method can dynamically adjust the number of Ranges based on the Dataset's row count.

@hamersaw @puchengy Could you please take a look and see if this makes sense? Thank you.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -85,10 +85,11 @@ case class AddIndexExec(
     val uuid = UUID.randomUUID()
     val indexType = IndexUtils.buildIndexType(method)
 
-    val indexBuildResult =
-      createIndexJob(lanceDataset, readOptions, uuid.toString, fragmentIds).run()
-
     val dataset = Utils.openDatasetBuilder(readOptions).build()
+
+    val indexBuildResult =
+      createIndexJob(dataset, lanceDataset, readOptions, uuid.toString, fragmentIds).run()
+
     try {
       // Merge index metadata after all fragments are indexed
       dataset.mergeIndexMetadata(uuid.toString, indexType, Optional.empty())
@@ -148,6 +149,7 @@ case class AddIndexExec(
   }
 
   private def createIndexJob(
+      dataset: Dataset,
       lanceDataset: LanceDataset,
       readOptions: LanceSparkReadOptions,
       uuid: String,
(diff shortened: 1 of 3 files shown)


puchengy:
@fangbo Hi, I will try to take a look but not sure if I can review it becasue I am still ramping up. Please ping @hamersaw if I don't provide comment in time, thanks!

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -84,10 +84,11 @@ case class AddIndexExec(
     val uuid = UUID.randomUUID()
     val indexType = IndexUtils.buildIndexType(method)
 
+    val dataset = Utils.openDatasetBuilder(readOptions).build()
+
     // Create distributed index job and run it
-    createIndexJob(lanceDataset, readOptions, uuid.toString, fragmentIds).run()
+    createIndexJob(dataset, lanceDataset, readOptions, uuid.toString, fragmentIds).run()
```
hamersaw: This change means if createIndexJob throws an exception then `dataset` is never close because it's outside of the try-finally.

Also, I think there may be a bigger issue where reading the manifest before `createIndexJob` can potentially miss index entries because it runs on a stale handle. We need to make sure this results in correct, performant execution.

hamersaw (commented):
Appreciate the add! I think overall this looks good, we just need to understand the implications of moving the dataset initialization.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
fangbo: > This change means if createIndexJob throws an exception then `dataset` is never close because it's outside of the try-finally.

Thank you for pointing this, I have fixed it.

> Also, I think there may be a bigger issue where reading the manifest before createIndexJob can potentially miss index entries because it runs on a stale handle. We need to make sure this results in correct, performant execution.

I think the readOptions is not changed when index building job is created and running. Logically, the whole process is based the same version. In my opinion, it is no problem. If I've misunderstood, please point it out for me. Thank you.

hamersaw (approved):
Looks great! Thanks for the iteration on this.

PR #441, opened 2026-04-16
fix: remove Array<Struct> filter pushdown workaround (upstream lance#…

summaryzb:
- Remove the workaround in LanceScanBuilder.pushFilters() that disabled all filter pushdown when the schema contained any Array<Struct> column
 - The upstream bug (lance#3578) was fixed in lance-core PR #4838 (commit ed6947f6, 2025-10-02), and lance-spark's current dependency on lance-core 5.0.0-beta.6 includes that fix
 - Update the corresponding unit test to verify filters are now correctly pushed down for schemas with nested Array<Struct> columns


hamersaw (approved):
Thanks for completing the loop here! Can we remove all comments about the upstream PR fix? Since this will make everything work e2e as expected, I'm not sure there is value in maintaining the historical context in code comments, it should be available in the git log.

summaryzb:
Appreciate the feedback. I'll clean up the legacy issue comments

hamersaw (approved):
Great follow up! Thanks.

PR #444, opened 2026-04-16
feat: support reading non-microsecond Arrow timestamp columns

summaryzb:
Lance datasets created by external tools (Python PyArrow, Rust) may contain Timestamp(SECOND), Timestamp(MILLISECOND), or Timestamp(NANOSECOND) columns. Spark only natively supports Timestamp(MICROSECOND). This change adds:

- Schema conversion in LanceArrowUtils.fromArrowField() to map all timestamp units to TimestampType (with TZ) or TimestampNTZType (without TZ)
- TimestampUnitAccessor for vectorized reads, converting non-microsecond Arrow timestamp vectors to microsecond longs at read time
- End-to-end read tests covering all units, TZ/NTZ, and null handling


lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseTimestampUnitReadTest.java
```diff
@@ -0,0 +1,344 @@
...
+    assertEquals(2, rows.size());
+    // TimestampNTZ values surface as LocalDateTime via getAs
+    java.time.LocalDateTime expected1 =
+        java.time.LocalDateTime.ofInstant(
+            Instant.ofEpochMilli(TS1_MILLIS), java.time.ZoneOffset.UTC);
+    java.time.LocalDateTime expected2 =
+        java.time.LocalDateTime.ofInstant(
+            Instant.ofEpochMilli(TS2_MILLIS), java.time.ZoneOffset.UTC);
+    assertEquals(expected1, rows.get(0).<java.time.LocalDateTime>getAs(1));
+    assertEquals(expected2, rows.get(1).<java.time.LocalDateTime>getAs(1));
```
hamersaw: Can we move the `java.time.*` to a top-level import?

hamersaw (commented):
Thanks for the fix! This will be really powerful. One small nit (imports) and then we also need to add a test stub at `lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/TimestampUnitReadTest.java` so this is covered across all versions.

hamersaw (approved):
Awesome, thanks!

PR #445, opened 2026-04-16
feat: require clustered distribution on write for SPJ

beinan:
## Summary

Adds `RequiresDistributionAndOrdering` to `SparkWrite` so that when `lance.partition.columns` is set, Spark clusters and sorts the write data by the partition column(s). This ensures each Lance fragment contains exactly one distinct partition value — the prerequisite for Storage-Partitioned Joins (SPJ) on the read path.

### Changes
- `SparkWrite` implements `RequiresDistributionAndOrdering` — returns `Distributions.clustered()` and ascending `SortOrder` for partition columns
- `LanceConstant.TABLE_OPT_PARTITION_COLUMNS` centralizes the table property key
- `LanceScanBuilder` uses `LanceConstant` directly (no duplicate constant)
- Thread `tableProperties` through `SparkWriteBuilder` → `SparkWrite`

### Tests
- Unit tests for `requiredDistribution()` / `requiredOrdering()` with and without partition columns
- `BasePartitionedWriteTest` integration test verifying that fragments written with partition columns each contain exactly one distinct partition value

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
@@ -92,8 +92,6 @@ public class LanceScanBuilder
 
   private final java.util.Map<String, String> tableProperties;
 
-  static final String TABLE_OPT_PARTITION_COLUMNS = LanceConstant.TABLE_OPT_PARTITION_COLUMNS;
-
   public LanceScanBuilder(
       StructType schema,
       LanceSparkReadOptions readOptions,
@@ -142,7 +140,7 @@ public class LanceScanBuilder
 
     // Collect all columns that need zonemap stats: filter columns + partition column (if declared).
     Set<String> columnsToLoad = extractReferencedColumns(pushedFilters);
-    String partitionColumn = tableProperties.get(TABLE_OPT_PARTITION_COLUMNS);
+    String partitionColumn = tableProperties.get(LanceConstant.TABLE_OPT_PARTITION_COLUMNS);
     if (partitionColumn != null && !partitionColumn.trim().isEmpty()) {
       partitionColumn = partitionColumn.trim();
       columnsToLoad.add(partitionColumn);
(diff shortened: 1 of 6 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -92,7 +93,7 @@ public class LanceScanBuilder
 
   private final java.util.Map<String, String> tableProperties;
 
-  static final String TABLE_OPT_PARTITION_COLUMNS = "lance.partition.columns";
+  static final String TABLE_OPT_PARTITION_COLUMNS = LanceConstant.TABLE_OPT_PARTITION_COLUMNS;
```
hamersaw: I don't think we need to define this constant twice. Lets just use `LanceConstant.TABLE_OPT_PARTITION_COLUMNS` directly everywhere.

hamersaw (commented):
@beinan thanks for the PR! I think this is in the right direction, Can we add some tests (unit + integration) to verify? And then just one concern (I'm hoping the tests will validate):

I don't think this does exactly what you're expecting. This PR imposes global sort and hash-based distribution of data on partition columns during write. Presumably this is so we have a dataset that adheres to the partitioned read path, namely one partition column variant per fragment.

However, this can have multiple partition key combinations sent to the same Spark partition (just a hash distribution) and without the writer being told to create a new fragment per partition column combination we would just get partition keys sorted within the same Lance fragment.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Done — removed the duplicate constant. Now using `LanceConstant.TABLE_OPT_PARTITION_COLUMNS` directly in LanceScanBuilder.

beinan:
Thanks for the detailed review! You're right that `Distributions.clustered()` uses hash partitioning, so multiple distinct values *can* land in the same Spark task if they hash to the same bucket.

However, the combination of **clustered distribution + sort ordering** ensures that within each task, rows are grouped contiguously by partition value. Since each `LanceDataWriter` (one per task) creates fragments via `Fragment.create()`, and lance-core respects the sorted order when writing, the result is that each fragment contains exactly one partition value — as long as fragment size limits don't split mid-value.

I've added an integration test (`BasePartitionedWriteTest.testWriteWithPartitionColumnClustersData`) that verifies this end-to-end: it writes 50 rows across 5 regions in a single INSERT and asserts that each resulting fragment has exactly 1 distinct region value using `_fragid`.

The key insight: even though hash collisions can send multiple partition values to the same task, the sort ordering within the task means the writer sees values in order (e.g., all "east" rows, then all "north" rows), and lance-core's fragment creation naturally produces one fragment per contiguous block.

jackye1995 (approved):
looks like there is a merge conflict, once resolved I can merge it!

PR #448, opened 2026-04-18
refactor: migrate filter pushdown from V1 Filter to V2 Predicate API

wombatu-kun:
## Summary

- Replace `SupportsPushDownFilters` with `SupportsPushDownV2Filters` in `LanceScanBuilder`, and migrate the downstream filter-analysis utilities (`FilterPushDown`, `RowAddressFilterAnalyzer`, `ZonemapFragmentPruner`) from the sealed `org.apache.spark.sql.sources.Filter` hierarchy to the extensible `org.apache.spark.sql.connector.expressions.filter.Predicate`.
- V1 `Filter` is legacy in Spark 3.3+; V2 `Predicate` is the current recommended API and supports arbitrary named predicates, making future extensions (e.g. custom functions) straightforward.
- No external behavior changes — SQL filter semantics, zonemap pruning, `_rowaddr` fragment pruning, and the compiled SQL `WHERE` clause are preserved.

## Notable details

- V2 `Literal.value()` returns `UTF8String` for strings; normalized to `String` in `ZonemapFragmentPruner` before `compareTo` against zone min/max stats (which store native `String`).
- Column access via `NamedReference.fieldNames()` joined with `"."` for nested paths (drop-in for V1 `Filter.references()` flat names).
- Test helper `TestPredicates` builds canonical V2 predicates with the same shape Spark produces via `SupportsPushDownV2Filters`, including UTF8String/Date-epoch-days/Timestamp-epoch-micros conversion matching Spark's V2 literal representation.

## Test plan

- [x] `make install` — builds successfully
- [x] `make test` — 619 tests pass (0 failures, 0 errors)
- [x] `make lint` — checkstyle + spotless pass
- [x] No remaining references to V1 Filter API anywhere in the codebase

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
@@ -209,12 +209,18 @@ public class FilterPushDown {
     if (value instanceof String) {
       return "'" + ((String) value).replace("'", "''") + "'";
     }
+    if (value instanceof org.apache.spark.sql.types.Decimal) {
+      return formatDecimalCast(((org.apache.spark.sql.types.Decimal) value).toJavaBigDecimal());
+    }
     if (value instanceof BigDecimal) {
-      BigDecimal bd = (BigDecimal) value;
-      int scale = bd.scale();
-      int precision = Math.max(bd.precision(), scale);
-      return "CAST(" + bd.toPlainString() + " AS DECIMAL(" + precision + ", " + scale + "))";
+      return formatDecimalCast((BigDecimal) value);
     }
     return value.toString();
+  }
+
+  private static String formatDecimalCast(BigDecimal bd) {
+    int scale = bd.scale();
+    int precision = Math.max(bd.precision(), scale);
+    return "CAST(" + bd.toPlainString() + " AS DECIMAL(" + precision + ", " + scale + "))";
   }
 }
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
@@ -15,195 +15,206 @@ package org.lance.spark.read;
 
 import org.lance.spark.utils.Optional;
 
-import org.apache.spark.sql.sources.And;
-import org.apache.spark.sql.sources.EqualNullSafe;
-import org.apache.spark.sql.sources.EqualTo;
-import org.apache.spark.sql.sources.Filter;
-import org.apache.spark.sql.sources.GreaterThan;
-import org.apache.spark.sql.sources.GreaterThanOrEqual;
-import org.apache.spark.sql.sources.In;
-import org.apache.spark.sql.sources.IsNotNull;
-import org.apache.spark.sql.sources.IsNull;
-import org.apache.spark.sql.sources.LessThan;
-import org.apache.spark.sql.sources.LessThanOrEqual;
-import org.apache.spark.sql.sources.Not;
-import org.apache.spark.sql.sources.Or;
-import org.apache.spark.sql.sources.StringContains;
-import org.apache.spark.sql.sources.StringEndsWith;
-import org.apache.spark.sql.sources.StringStartsWith;
+import org.apache.spark.sql.connector.expressions.Expression;
+import org.apache.spark.sql.connector.expressions.Literal;
+import org.apache.spark.sql.connector.expressions.NamedReference;
+import org.apache.spark.sql.connector.expressions.filter.And;
+import org.apache.spark.sql.connector.expressions.filter.Not;
+import org.apache.spark.sql.connector.expressions.filter.Or;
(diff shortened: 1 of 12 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
```diff
@@ -15,195 +15,206 @@
...
-        if (sb.length() > 0) {
-          sb.append(", ");
-        }
-        sb.append(compileValue(obj));
-      }
-      return sb.toString();
-    } else {
-      return value.toString();
     }
+    return value.toString();
```
hamersaw, line 218: Decimal literal CAST wrapper lost in V2 migration: Spark V2 `LiteralValue` stores decimal values using Catalyst's internal `org.apache.spark.sql.types.Decimal`, not `java.math.BigDecimal` — the same reason this file already had to add a `UTF8String` branch alongside `String` (lines 206-210). On a real query like `WHERE net_profit >= 100.00`, `literal.value()` returns a `Decimal`, no `instanceof` check matches, and execution falls through to `value.toString()` at line 218, yielding `net_profit >= 100.00`. Lance's DataFusion parser then infers Float64 for the bare numeric literal and fails type resolution against the Decimal128 column — the exact bug the CAST wrapper was added to prevent (see comment in the V1 version of this method and the `testDecimalFilterPushDown` rationale). The unit tests don't catch this because `TestPredicates.literalOf` (TestPredicates.java:118-122) passes the raw `BigDecimal` straight into `LiteralValue` without converting to the internal `Decimal` representation that Spark actually uses, so the test exercises a code path that production never hits.

**Fix:** Add a branch handling `org.apache.spark.sql.types.Decimal` (call `toJavaBigDecimal()` then reuse the existing precision/scale CAST logic). Also update `TestPredicates.literalOf` to normalize `BigDecimal → Decimal.apply(bd)` so the test mirrors production.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
```diff
@@ -217,27 +200,34 @@ private static boolean zoneMatchesComparison(
...
     }
 
     Set<Integer> matchingFragments = new HashSet<>();
     for (ZoneStats zone : stats) {
-      for (Object value : values) {
+      for (int i = 1; i < children.length; i++) {
+        if (!(children[i] instanceof Literal)) {
+          continue;
+        }
+        Object value = normalizeLiteral(((Literal<?>) children[i]).value());
```
hamersaw: IN literal extraction repeated per zone: The previous V1 implementation received `Object[] values` pre-extracted; the new V2 form re-evaluates `children[i] instanceof Literal`, casts to `Literal<?>`, calls `.value()`, and runs `normalizeLiteral` (another instanceof + possible UTF8String→String allocation) inside the outer `for (ZoneStats zone : stats)` loop. The literal values are invariant across zones, so this is O(zones × inValues) work where O(inValues) suffices. On wide tables this loop fires for every zone in every column with zonemap stats — e.g. an `IN (...)` with 50 values against 1000 zones now performs 50000 cast+normalize ops instead of 50, including up to 50000 redundant `UTF8String.toString()` allocations for string IN lists.

**Fix:** Hoist the value extraction above the `for (ZoneStats zone : stats)` loop: build a `Object[] normalizedValues` once (skipping non-Literal children), then iterate the cached array inside the zone loop.

hamersaw (commented):
Thanks for the PR! Overall I think this looks great, just a few small regressions then I think we should be good.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
wombatu-kun: Confirmed and fixed in 5e00eea. Added an `org.apache.spark.sql.types.Decimal` branch in `compileLiteral` (calls `toJavaBigDecimal()` and reuses the precision/scale CAST via a small `formatDecimalCast` helper), and normalized `TestPredicates.literalOf` to wrap `BigDecimal` as `Decimal.apply(bd)` so the existing decimal pushdown tests now exercise the production path. Also added `testDecimalFilterPushDownEndToEnd` in `BaseSparkDataTypeRoundtripTest` that drives a real Spark `WHERE v >= 100.00` query against a `Decimal(10, 2)` Lance table — a regression would surface as a DataFusion type-resolution failure, not just a unit-test miss.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
wombatu-kun: fixed

hamersaw (approved):
Looks great! I added a bug fix so that we don't accidentally prune fragments using ZoneMaps where the filter has an expression. Also added test for this to ensure we don't regress in the future.

PR #452, opened 2026-04-18
refactor: centralize Lance metadata column definitions

wombatu-kun:
## Summary

- Extract the five `MetadataColumn` instances (`_rowid`, `_rowaddr`, `_row_created_at_version`, `_row_last_updated_at_version`, `_fragid`) from `LanceDataset` into a new `LanceMetadataColumns` registry class
- Replace the hand-maintained exclusion filter and if-ladder in `LanceFragmentScanner.getColumnNames` with iteration over `ALL` / `PROJECTABLE`
- Migrate callers that only needed the column name string (`LanceDataset.X_COLUMN.name()` → `LanceConstant.X`)

## Motivation

Before this change, adding a virtual column took edits in three coupled sites: a new anonymous `MetadataColumn` constant in `LanceDataset`, an entry in the `METADATA_COLUMNS` array, a clause in the "exclude from regular columns" filter, and a stanza in the "append to scanner projection" if-ladder. All had to stay in lock-step; forgetting one produced silent bugs (column in schema but not in scanner output, or Lance rejecting an unknown name). After, adding a column is a one-line registry entry.

This refactor unblocks the `_score` column PR (stacked in the FTS PR series); future virtual columns such as `_score_explain` benefit from the same one-line shape.

Pure refactor — no user-visible behavior change. The asymmetry between `ALL` (exclusion + Spark registration) and `PROJECTABLE` (scanner projection) pins the invariant that `_fragid` is computed per-fragment outside the scanner, preserving the original behavior.

## Test plan

- [x] `make test SPARK_VERSION=3.5 SCALA_VERSION=2.12` — full regression passes
- [x] `make lint` — checkstyle + spotless clean
- [ ] CI green across all 5 version modules (3.4_2.12, 3.5_2.12, 3.5_2.13, 4.0_2.13, 4.1_2.13)

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java
@@ -13,6 +13,7 @@
  */
 package org.lance.spark;
 
+import org.lance.spark.read.LanceMetadataColumns;
 import org.lance.spark.read.LanceScanBuilder;
 import org.lance.spark.utils.BlobUtils;
 import org.lance.spark.write.AddColumnsBackfillWrite;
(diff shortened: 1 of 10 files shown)


wombatu-kun:
Closing this - on reflection the central registry isn't the right fit. Since this PR was opened, main moved metadata-column handling in a different direction (e.g. _rowid / _rowaddr are now requested through dedicated withRowId / withRowAddress scanner flags rather than the projection list), so a uniform ALL / PROJECTABLE registry no longer matches how the scanner actually projects columns. The only PR that depended on it (#455, the _score column) has been reworked to define _score directly on LanceDataset following the existing metadata-column pattern, so it is now self-contained and this refactor is no longer needed.

PR #456, opened 2026-04-18
fix: support CHAR/VARCHAR types in SQL DDL write path

LuciferYang:
## What

`CREATE TABLE ... (v CHAR(n))` or `VARCHAR(n)` via a V2 catalog blows up on the write path. Two separate match misses, different blast radius.

## Why

Two Scala pattern styles with different semantics:

```
case _: StringType   →  isInstanceOf (type test)
case StringType      →  equals()     (value comparison)
```

**`toArrowType` uses `case _: StringType` (type test)**

On Spark 4.0+, `CharType`/`VarcharType` extend `StringType`, so they match fine.
On 3.4/3.5 they extend `AtomicType` instead — no match, throws `unsupportedDataTypeError`.

**`createFieldWriter` uses `case (StringType, ...)` (value comparison)**

`CharType(10).equals(StringType)` is `false` on every Spark version:
- 3.4/3.5: completely different type hierarchy
- 4.0+: same hierarchy, but `CharType` carries a `FixedLength` constraint while the `StringType` singleton has `NoConstraint`

So even when `toArrowType` succeeds, the writer still fails.

Summary:

| Location | 3.4/3.5 | 4.0+ |
|---|---|---|
| `toArrowType` (`_: StringType`) | broken | fine |
| `createFieldWriter` (`StringType`) | broken | broken |

## How

- `LanceArrowUtils.toArrowType`: add `case _: CharType | _: VarcharType` branches mapping to Utf8 / LargeUtf8 (respecting `largeVarTypes`). On 4.0+ these are dead code.
- `LanceArrowWriter.createFieldWriter`: add `case (_: CharType | _: VarcharType, vector)` branches dispatching to StringWriter / LargeStringWriter.

Length constraints are discarded — Arrow has no varchar concept.

## Tests

- `LanceArrowUtilsSuite`: 4 unit tests covering CharType/VarcharType → Arrow type mapping with `largeVarTypes` on/off
- `BaseCharVarcharRoundtripTest`: E2E via V2 catalog — DDL create + insert + read back, covering CHAR, VARCHAR, mixed columns, and NULLs
- `CharVarcharRoundtripTest` subclass in Spark 3.4 and 3.5 modules; Spark 4.0 and 4.1 inherit it automatically via shared test-source configuration


LuciferYang:
cc @hamersaw and @fangbo

hamersaw (approved):
Looks great! Thanks for the patience.

PR #457, opened 2026-04-19
feat: support Spark 4.1 TimeType (time-of-day)

LuciferYang:
## Summary

Spark 4.1 introduces `TimeType`, representing time-of-day (`00:00:00.000000` to `23:59:59.999999`) stored internally as nanoseconds since midnight. Lance already supports Arrow `Time32`/`Time64` types at the core level; this PR adds read, write, and filter pushdown support in the lance-spark connector layer.

Since `TimeType` only exists in Spark 4.1+, and the base module must compile against older Spark versions, all `TimeType` references use reflection. This follows the same cross-version compatibility pattern already established in the project — e.g., `Float2Vector` (Arrow 18+) uses class name checks rather than direct imports to avoid compile-time dependencies.

### Divergence from Spark's default Arrow mapper

Spark's own `ArrowUtils.fromArrowType` accepts only `Time(NANOSECOND, 64)` and throws `unsupportedArrowTypeError` on every other Arrow Time variant. This PR deliberately accepts all four (`Time(SEC,32)`, `Time(MILLI,32)`, `Time(MICRO,64)`, `Time(NANO,64)`) and rescales to nanoseconds in `TimeUnitAccessor`, so Lance datasets written by non-Spark producers (Arrow, DataFusion, Polars) round-trip cleanly. The write path stays strict — Spark `TimeType` always serializes as `Time(NANOSECOND, 64)`, matching Spark's default.

### New files

- `TimeUnitAccessor.java` — read-path accessor for Arrow Time vectors
- `TimeTypeRoundtripTest.java` — E2E roundtrip test in the Spark 4.1 module

### Write path

- `LanceArrowUtils.toArrowType` maps `TimeType` to `ArrowType.Time(NANOSECOND, 64)` via class name check to avoid compile-time dependency
- `LanceArrowWriter` adds `TimeNanoWriter` — Spark stores nanos internally, Arrow `Time(NANOSECOND)` expects nanos, so the value is passed through directly with no conversion

### Read path

- `LanceArrowUtils.fromArrowField` adds an `ArrowType.Time` branch that resolves the Spark type via `TimeUtils.resolveSparkTimeType()` — returns `TimeType` on Spark 4.1+, falls back to `LongType` on older versions
- New `TimeUnitAccessor` handles all four Arrow Time vectors (TimeSec / TimeMilli / TimeMicro / TimeNano) and normalizes them to nanoseconds for Spark. Time32 vectors return `int` and Time64 return `long` with no common base class, so a `TypedGetter` functional interface is used for typed access to avoid per-row boxing overhead
- `LanceArrowColumnVector` dispatches the four Time vector types to `TimeUnitAccessor`

### Filter pushdown

- `FilterPushDown.isColumnLiteralShape` (V2 Predicate API) rejects literals whose `dataType()` class name is `org.apache.spark.sql.types.TimeType`. Lance's DataFusion planner (`lance-datafusion/src/planner.rs`) does not yet support `SQLDataType::Time` in `parse_type()`, so time predicates are evaluated post-scan by Spark.

### Tests

- `FilterPushDownTest.testTimeTypeFilterRejected`: constructs a `TimeType` literal via reflection (skipped on Spark < 4.1) and verifies rejection plus `processPredicates` bucketing.
- `LanceArrowUtilsSuite`: bidirectional mapping for all four Arrow Time variants and nested forms (struct / list / map). On Spark < 4.1 these assert the `LongType` fallback; on Spark 4.1+ they assert `TimeType`.
- `LanceArrowColumnVectorSuite`: read-path assertions for all four Time vector types, verifying normalization to nanos.
- `TimeTypeRoundtripTest` (4.1 module): end-to-end write-read roundtrip.


LuciferYang:
cc @hamersaw @fangbo Is this one acceptable?

hamersaw (approved):
Looks great! Thanks for the fix and patience on the review. I'm going to have some time opening up here to get back on top of things after this week.

LuciferYang:
Thank you @hamersaw ~

PR #464, opened 2026-04-21
feat: preserve Arrow Date(MILLISECOND) columns through Spark roundtrip

summaryzb:
Preserve Arrow `Date(MILLISECOND)` columns through Spark read-write roundtrips by carrying the original Arrow date unit in Spark field metadata. Without this change, Lance datasets created by PyArrow or other non-Spark sources that use `Date(MILLISECOND)` would silently downgrade to `Date(DAY)` when written back through Spark, corrupting interoperability with downstream consumers that expect millisecond-unit dates.


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -104,6 +116,9 @@ object LanceArrowUtils {
       case ts: ArrowType.Timestamp =>
         if (ts.getTimezone != null && ts.getTimezone.nonEmpty) TimestampType
         else TimestampNTZType
+      // TODO: Date(MILLISECOND), FixedSizeList, and LargeUtf8 metadata is lost for array
+      // elements because fromArrowField returns DataType without metadata. Track as a
+      // known limitation; address in a follow-up.
```
hamersaw: Can you make sure there is a ticket for this?

hamersaw: And when it exists, lets remove the TODO comments. The captured ticket should be enough to remind us this needs to be fixed.

hamersaw (commented):
Looks great! Lets just turn the TODO's into github issues and fix the integration tests then it will be ready to merge!

hamersaw (approved):

PR #470, opened 2026-04-22
feat: preserve FixedSizeBinary(n) byte width across read-write roundtrips

LuciferYang:
## Summary

Arrow's `FixedSizeBinary(n)` has no Spark equivalent, so on the way in it gets mapped to `BinaryType` and the byte width is lost. On the way back out, the connector then writes it as variable-length `Binary` — the schema silently degrades across a roundtrip. This PR preserves the byte width in Spark field metadata on read and uses it on write, so a column that comes in as `FixedSizeBinary(16)` goes back out as `FixedSizeBinary(16)`. It also adds a way for users to declare the width on `CREATE TABLE` via a TBLPROPERTY, and fails loudly on obviously-invalid widths instead of handing garbage to Arrow.

## Changes

A new `FixedSizeBinaryUtils` carries the metadata key (`arrow.fixed-size-binary.byte-width`), plus a helper to check whether `Metadata` carries the byte-width key and one to build the TBLPROPERTY key. Parallel to `LargeVarCharUtils` and `VectorUtils`, just for a different Arrow type.

`LanceArrowUtils.fromArrowSchema` gets a new case in its per-field metadata switch: when an Arrow field's type is `FixedSizeBinary`, stash `fsb.getByteWidth` into Spark metadata. `LanceArrowUtils.toArrowField` gets the mirror: a new `BinaryType` case guarded by `FixedSizeBinaryUtils.hasFixedSizeBinaryMetadata(metadata)` that reads the width back out and builds an Arrow `FixedSizeBinary(n)` field. Plain `BinaryType` without metadata falls through to the existing catch-all and still produces variable-length Binary, so nothing changes for existing tables.

`LanceArrowWriter` grows a `FixedSizeBinaryWriter`. It caches `byteWidth` once at construction time, validates `bytes.length == byteWidth` on each `setValue` call, and throws `IllegalArgumentException` naming the column with the expected-vs-actual counts when they disagree. Without the check, `FixedSizeBinaryVector.setSafe` silently truncates or corrupts the vector depending on Arrow version, so this is the difference between "clear user error" and "debugging a mystery later."

`SchemaConverter.processSchemaWithProperties` picks up `addFixedSizeBinaryMetadata`, slotting into the existing chain after `addLargeVarCharMetadata`. A TBLPROPERTY of the form `<column>.arrow.fixed-size-binary.byte-width` is parsed through `parseFixedSizeBinaryByteWidth`, which rejects non-numeric strings, zero, negatives, and values above `Integer.MAX_VALUE` with column-named error messages. Matches the type-mismatch guard used by the sibling methods: BINARY column → metadata attached; non-BINARY column with the property → `IllegalArgumentException`; column that doesn't exist → silently ignored (consistent with `addCompressionMetadata` et al.).

## Tests

`SchemaConverterTest` adds tests for the happy path (metadata attached to a `BinaryType` column), type-mismatch rejection, non-numeric width, zero/negative widths, and int-overflow width — each verifying the error message names the offending column so users can find it.

`LanceArrowUtilsSuite` adds two schema-level tests: one feeds a `StructType` with the metadata through `toArrowSchema` and asserts the emitted Arrow type is `FixedSizeBinary(16)`; the other does the full round-trip (Arrow `FixedSizeBinary(32)` → Spark `BinaryType` + metadata → Arrow `FixedSizeBinary(32)`) and asserts the width survives at every hop.

`LanceArrowColumnVectorSuite` adds a read-path test that allocates a `FixedSizeBinaryVector` with 10 non-null entries and 1 null, wraps it in `LanceArrowColumnVector`, and asserts the values round-trip through `getBinary` / `isNullAt`.

`mvn test -pl lance-spark-base_2.12` passes 308 tests. `lance-spark-3.4_2.12` cross-compile is green. Spotless and Checkstyle clean.

## Usage

A table whose physical schema contains `FixedSizeBinary(16)` columns now roundtrips automatically — no configuration needed. To create a new table that emits `FixedSizeBinary(n)` instead of variable-length Binary on write, declare the width in `TBLPROPERTIES`:

```sql
CREATE TABLE t (
  id BIGINT,
  hash BINARY
) USING lance TBLPROPERTIES (
  'hash.arrow.fixed-size-binary.byte-width' = '16'
);
```

Every row written to `hash` must then be exactly 16 bytes; a mismatch throws `IllegalArgumentException` naming the column at write time rather than silently corrupting the vector.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
@@ -58,17 +58,15 @@ public class FixedSizeBinaryUtils {
    * Get the byte width from a Spark field's metadata.
    *
    * @param field the Spark struct field
-   * @return the byte width, or -1 if not a FixedSizeBinary field
+   * @return the byte width, or -1 if the field has no FixedSizeBinary metadata
    */
   public static int getByteWidth(StructField field) {
     if (!isFixedSizeBinaryField(field)) {
       return -1;
     }
-    try {
-      return (int) field.metadata().getLong(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY);
-    } catch (Exception e) {
-      return -1;
-    }
+    // isFixedSizeBinaryField confirmed the key is present; getLong throwing here would indicate
+    // internal corruption and should surface rather than be masked as "-1 means no metadata".
+    return (int) field.metadata().getLong(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY);
   }
 
   /**
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
@@ -13,9 +13,7 @@
  */
 package org.lance.spark.utils;
 
-import org.apache.spark.sql.types.BinaryType;
 import org.apache.spark.sql.types.Metadata;
-import org.apache.spark.sql.types.StructField;
 
 /**
  * Utility class for FixedSizeBinary Arrow type metadata handling. When reading a Lance file that
@@ -29,22 +27,6 @@ public class FixedSizeBinaryUtils {
       "arrow.fixed-size-binary.byte-width";
 
   /**
-   * Check if a Spark field carries FixedSizeBinary metadata.
-   *
-   * @param field the Spark struct field to check
-   * @return true if the field is BinaryType with FixedSizeBinary byte-width metadata
-   */
-  public static boolean isFixedSizeBinaryField(StructField field) {
-    if (field == null) {
-      return false;
-    }
-    if (!(field.dataType() instanceof BinaryType)) {
-      return false;
-    }
-    return hasFixedSizeBinaryMetadata(field.metadata());
-  }
-
-  /**
    * Check if metadata contains the FixedSizeBinary byte-width key.
    *
    * @param metadata the Spark field metadata
@@ -52,21 +34,6 @@ public class FixedSizeBinaryUtils {
    */
   public static boolean hasFixedSizeBinaryMetadata(Metadata metadata) {
     return metadata != null && metadata.contains(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY);
-  }
-
-  /**
-   * Get the byte width from a Spark field's metadata.
-   *
-   * @param field the Spark struct field
-   * @return the byte width, or -1 if the field has no FixedSizeBinary metadata
-   */
-  public static int getByteWidth(StructField field) {
-    if (!isFixedSizeBinaryField(field)) {
-      return -1;
-    }
-    // isFixedSizeBinaryField confirmed the key is present; getLong throwing here would indicate
-    // internal corruption and should surface rather than be masked as "-1 means no metadata".
-    return (int) field.metadata().getLong(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY);
   }
 
   /**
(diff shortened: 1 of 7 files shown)


LuciferYang:
cc @hamersaw @fangbo

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -362,6 +363,24 @@ private[arrow] class BinaryWriter(val valueVector: VarBinaryVector) extends Lanc
...
+  override def setNull(): Unit = {
+    valueVector.setNull(count)
+  }
+  override def setValue(input: SpecializedGetters, ordinal: Int): Unit = {
+    val bytes = input.getBinary(ordinal)
+    // FixedSizeBinaryVector.setSafe silently truncates/corrupts when the input length does not
+    // match byteWidth; validate up-front so the user gets a clear, column-aware error instead.
+    if (bytes.length != byteWidth) {
+      throw new IllegalArgumentException(
+        s"FixedSizeBinary column expects $byteWidth bytes per row, got ${bytes.length}")
```
hamersaw: Is there any way to output which column this error occurred on?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
```diff
@@ -0,0 +1,82 @@
...
+   */
+  public static boolean isFixedSizeBinaryField(StructField field) {
+    if (field == null) {
+      return false;
+    }
+    if (!(field.dataType() instanceof BinaryType)) {
+      return false;
+    }
+    return hasFixedSizeBinaryMetadata(field.metadata());
+  }
```
hamersaw: I don't think this is called anywhere, is it still useful?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
```diff
@@ -0,0 +1,82 @@
...
+   * @return the byte width, or -1 if the field has no FixedSizeBinary metadata
+   */
+  public static int getByteWidth(StructField field) {
+    if (!isFixedSizeBinaryField(field)) {
+      return -1;
+    }
+    // isFixedSizeBinaryField confirmed the key is present; getLong throwing here would indicate
+    // internal corruption and should surface rather than be masked as "-1 means no metadata".
+    return (int) field.metadata().getLong(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY);
+  }
```
hamersaw: Similarly, not called anywhere.

hamersaw (commented):
Thanks for the PR! A few small nits and then lets get this merged.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
LuciferYang: Good catch — added the column name via the base class's `name`. The error now reads: `FixedSizeBinary column '<name>' expects N bytes per row, got M`. Fixed in 77b27c8.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
LuciferYang: You're right, leftover from an earlier draft. Removed.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/FixedSizeBinaryUtils.java
LuciferYang: Same — removed.

hamersaw (approved):
Thanks for the PR!

LuciferYang:
Thank you @hamersaw

PR #471, opened 2026-04-22
feat: support VectorUDT writes via UDT unwrapping in LanceArrowWriter

LuciferYang:
## Summary

`LanceArrowWriter.createFieldWriter` had no case for `UserDefinedType`, so writing a DataFrame with a UDT column (most commonly MLlib's `VectorUDT` — the type produced by `VectorAssembler`, returned by many ML transformers) threw `UnsupportedOperationException`. This PR adds a single case that unwraps the UDT to its `sqlType` and recurses, matching the pattern Spark's native `ArrowWriter` uses. Read-back already worked because Arrow has no UDT concept — the column comes back as the underlying struct sqlType — so the write path is the only gap, and that's what this fixes.

## Changes

One new case in `LanceArrowWriter.createFieldWriter`:

```scala
case (udt: UserDefinedType[_], _) =>
  createFieldWriter(vector, udt.sqlType, metadata)
```

Placed just before the catch-all `case (dt, _) => throw new UnsupportedOperationException(...)` so specific matchers run first and anything that's still a UDT at that point gets unwrapped. Metadata is threaded through so nested cases that need it — e.g. `FixedSizeListWriter` when a UDT's sqlType resolves to `ArrayType` with embedding metadata — still receive the right field metadata.

## Tests

`BaseSparkDataTypeRoundtripTest` picks up `testVectorUDTRoundtrip`: creates a DataFrame with an `id INT` column and a `vec VectorUDT` column, writes two rows (a `DenseVector` and a `SparseVector`) as Lance, reads back, and asserts two things:

1. **Schema contract.** Arrow has no UDT, so the read-back schema must NOT carry `VectorUDT` — it must come back as the underlying sqlType (`StructType`). Asserts both halves: `!instanceof VectorUDT` and `instanceof StructType`. Locks this behavior so a future change that accidentally preserves UDT in field metadata is caught loudly.

2. **Value fidelity.** Reconstructs each row's vector from the read-back struct and checks `equals` against the original. A small `reconstructVector(Row)` helper does the manual reconstruction because `VectorUDT.deserialize` expects `InternalRow`, not the public `Row` that Spark hands back in test assertions. The helper switches on the type byte — 0 = sparse, 1 = dense — and throws `IllegalArgumentException` for anything else rather than silently misinterpreting bogus data as sparse.

A null-VectorUDT row is intentionally omitted and documented in the test's Javadoc: `VectorUDT.sqlType` marks its inner `type` field as non-nullable, and writing a parent-null struct produces null placeholders in the non-nullable children that Lance's native writer rejects. That's a Lance-side limitation with struct-of-non-nullable-children, not specific to UDTs; out of scope here.

`spark-mllib` is added as a **test-scope** dependency across all eight non-bundle modules (`-base_2.12`, `-base_2.13`, `-3.4_2.12/2.13`, `-3.5_2.12/2.13`, `-4.0_2.13`, `-4.1_2.13`). Each module pins its own `spark<version>.version` so cross-compile stays version-accurate. Bundle modules don't run tests so they don't need it. The 4.0/4.1 modules inherit the concrete `SparkDataTypeRoundtripTest` subclass via cross-compile from `lance-spark-3.5_2.12/src/test/java`, so the VectorUDT test executes on 4.0 and 4.1 too — `spark-mllib` is needed there for both compile and runtime.

### DIFF
--- a/lance-spark-3.4_2.12/pom.xml
+++ b/lance-spark-3.4_2.12/pom.xml
@@ -26,6 +26,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-mllib_${scala.compat.version}</artifactId>
+            <version>${spark34.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.lance</groupId>
             <artifactId>lance-spark-base_${scala.compat.version}</artifactId>
             <version>${lance-spark.version}</version>
--- a/lance-spark-3.4_2.13/pom.xml
+++ b/lance-spark-3.4_2.13/pom.xml
@@ -28,6 +28,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-mllib_${scala.compat.version}</artifactId>
+            <version>${spark34.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.lance</groupId>
             <artifactId>lance-spark-base_${scala.compat.version}</artifactId>
             <version>${lance-spark.version}</version>
--- a/lance-spark-3.5_2.12/pom.xml
+++ b/lance-spark-3.5_2.12/pom.xml
@@ -23,6 +23,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-mllib_${scala.compat.version}</artifactId>
+            <version>${spark35.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>org.lance</groupId>
             <artifactId>lance-spark-base_${scala.compat.version}</artifactId>
             <version>${lance-spark.version}</version>
--- a/lance-spark-3.5_2.13/pom.xml
+++ b/lance-spark-3.5_2.13/pom.xml
@@ -27,6 +27,12 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-mllib_${scala.compat.version}</artifactId>
+            <version>${spark35.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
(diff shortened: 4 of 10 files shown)


jiaoew1991 (approved):

LuciferYang:
Thanks for your review. @jiaoew1991 ,can we merge this PR? I don't have merge permission.

LuciferYang:
cc @hamersaw and @fangbo

hamersaw (approved):
Thanks for the PR! I don't think there's anything wrong with this, but I'm wondering what the use-case is? IIUC reading fields written by a VectorUDT will just return a struct without any metadata information. So the write will succeed, but the read seems to be useless. I would feel a little more comfortable merging if I understood the plan moving forward.

LuciferYang:
Thanks for the review! @hamersaw 

The use case is removing the hard failure. Today, anything coming out of `VectorAssembler` or most MLlib transformers — i.e. the standard shape of feature data in a Spark ML pipeline — can't be written to a Lance table at all. The write throws `UnsupportedOperationException` before you get a chance to deal with the type mismatch. After this PR, it just writes.

You're right that the read side comes back as struct without UDT metadata. That's unavoidable: UDT is a Spark-only concept, so any persistence format has to unwrap to its `sqlType` — Spark does exactly this in `ArrowUtils.toArrowField` for Arrow, and in Parquet/ORC/JSON/CSV/Avro for their respective formats. This PR adds the same unwrap pattern to `LanceArrowWriter`. Anything we did to preserve UDT info in field metadata would only round-trip back to Spark; non-Spark consumers (Python, Rust) still see the struct, because that's the actual on-disk shape.

Reconstruction on read is mechanical — ~15 lines in the test, switching on the type byte for sparse vs. dense. (`VectorUDT.deserialize` does the same logic but takes `InternalRow`, which test assertions don't expose, so the test inlines the switch.) If it'd close the gap for you, I can do a follow-up that auto-rewraps on read via an opt-in option (e.g. `lance.udt.restore.<column> = <udt-fqn>`) — a schema-level rebrand at scan-build time, no partition-reader changes. But I'd rather land the write fix first — right now the write doesn't reach the point where "useless on read" is even relevant.

hamersaw:
@LuciferYang , sounds great! I'll get this merged and be awaiting the follow up PR.

LuciferYang:
Thank you @hamersaw

PR #474, opened 2026-04-23
feat(dfp): implement dynamic file pruning via SupportsRuntimeV2Filtering

LuciferYang:


LuciferYang:
It’s still in the draft stage.

PR #475, opened 2026-04-23
fix: propagate index_details from distributed index creation

LuciferYang:
## Background

After distributed index creation via `AddIndexExec`, querying zonemap indexes on the read side fails with:

```
WARN LanceScanBuilder: Failed to query zonemap indexes:
LanceError(Index): Index details are required for index description.
This index must be retrained to support this method.
```

The root cause: when executors call `createIndex`, the returned `Index` object carries `index_details` (protobuf bytes), but those bytes were discarded when merging results back on the driver. The committed `IndexMetadata` ends up with empty `index_details`. The no-arg `describeIndices()` on the Rust side doesn't filter these out and blows up.

## Fix

**Write side (AddIndexExec)**

- `FragmentIndexTask` and `RangeBTreeIndexBuilder` now capture `index_details` bytes after `createIndex` and ship them back to the driver via Kryo
- Driver picks the first non-empty result with `IndexUtils.collectFirstIndexDetails()` and sets it on `Index.Builder.indexDetails()` before committing the transaction
- Extracted `IndexUtils.extractIndexDetails()` for Java Optional → Scala Option conversion; both the fragment and range paths use the same serialization format

**Read side (LanceScanBuilder)**

- `findZonemapIndexedColumns()` now calls `describeIndices(criteria)` with an empty `IndexCriteria` instead of the no-arg overload, so the Rust side silently skips indexes missing `index_details` — backward compatible with datasets created before this fix

## Tests

- `testBTreeIndexHasIndexDetails` — fragment-based btree path
- `testRangeBTreeIndexHasIndexDetails` — range-based btree path
- `testFtsIndexHasIndexDetails` — FTS (inverted) path

All three verify that `index_details` is populated, `describeIndices(criteria)` returns the correct index type, and the no-arg `describeIndices()` no longer throws.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
@@ -16,6 +16,7 @@ package org.lance.spark.read;
 import org.lance.Dataset;
 import org.lance.Fragment;
 import org.lance.ManifestSummary;
+import org.lance.index.IndexCriteria;
 import org.lance.index.IndexDescription;
 import org.lance.index.scalar.ZoneStats;
 import org.lance.ipc.ColumnOrdering;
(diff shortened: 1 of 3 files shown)


fangbo:
Great fix!

hamersaw (approved):
Great fix! Thanks.

LuciferYang:
Thank you @hamersaw and @fangbo

PR #481, opened 2026-04-27
feat: support Lance index metadata in Spark indexing

jackye1995:
## Summary
- preserve Lance index details, version, type, and created-at metadata when committing Spark-built indexes
- reconnect namespace-backed dataset opens from Spark tasks when options were deserialized, without creating runtime namespaces unless configured
- document and test FTS v2 via LANCE_FTS_FORMAT_VERSION=2 on driver and executors
- add Docker integration assertions for index details while keeping Docker test targets generic

Refs lance-format/lance#6624

Local validation covered targeted AddIndexTest coverage on Spark 3.5 / Scala 2.12, the same test with LANCE_FTS_FORMAT_VERSION=2, LanceSparkWriteOptionsTest, clean compile for Spark 3.5 / Scala 2.13 plus Spark 4.0 / Scala 2.13, make lint, and Makefile dry-run coverage for Docker env propagation. CI covers the Dockerized integration matrix.


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -518,24 +543,80 @@ object IndexUtils {
...
+    }
   }
 
-  /** Returns the first non-empty index_details from serialized worker results. */
-  def collectFirstIndexDetails(encodedResults: Array[String]): Option[Array[Byte]] = {
-    encodedResults.iterator
-      .map(encoded => decode[Option[Array[Byte]]](encoded))
-      .collectFirst { case Some(details) => details }
+  /** Extracts the commit metadata from a newly created Index. */
+  def extractIndexBuildResult(index: Index): IndexBuildResult = {
```
hamersaw, line 568: IIUC this now retrieves the index details from every single worker rather than just accepting the first. For a dataset with a highly segmented index this could introduce latency. I'm not sure it will be significant, but worth calling out here IMO.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
```diff
@@ -139,6 +151,17 @@ public Dataset build() {
             .readOptions(roBuilder.build())
             .build();
       }
+      LanceNamespace runtimeNamespace =
+          LanceRuntime.getOrCreateNamespace(runtimeNamespaceImpl, runtimeNamespaceProperties);
```
hamersaw: This is calling `getOrCreateNamespace` on every single dataset open, even when no runtime namespace is configured. Does it make sense to gate this behind a `runtimeNamespaceImpl != null` check (that already exists)?

Makefile
```diff
@@ -188,8 +189,14 @@ docker-test:
...
 		$(if $(TEST_BACKENDS),-e TEST_BACKENDS=$(TEST_BACKENDS)) \
+		$(if $(LANCE_FTS_FORMAT_VERSION),-e LANCE_FTS_FORMAT_VERSION=$(LANCE_FTS_FORMAT_VERSION)) \
 		lance-spark-test:$(SPARK_VERSION)_$(SCALA_VERSION) \
-		"pytest /home/lance/tests/ -v --timeout=180"
+		"pytest /home/lance/tests/ -v --timeout=180 $(PYTEST_ARGS)"
+
+.PHONY: docker-test-fts-v2
+docker-test-fts-v2:
+	$(MAKE) docker-test LANCE_FTS_FORMAT_VERSION=2 \
+		PYTEST_ARGS="-k test_create_fts_index_honors_format_version_v2"
```
hamersaw: Adding a new command just to set an environment variable on tests doesn't feel like it will scale well as we need to enhance our test coverage. I think there's a ton of ways around this:
(1) monkeypatch the env var into the pytest (maybe won't work?)
(2) add this to the matrix in github workflows and set the env var for FTS v2 tests (while restricting the test scope)
(3) etc

But I'm wondering if there's a reason we require the env var to create a FTS V2 index? Why can we not include this in the index creation metadata? Then we could simply use it in our pytest.

I would rather remove this and miss the manual test coverage rather than set precedent for bloating this unnecessarily.

hamersaw (approved):
I think there's a few rough edges, but nothing blocking a merge.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
jackye1995: Addressed in 3a87165: runtime namespace lookup is now gated by runtimeNamespaceImpl != null, so normal URI opens do not call getOrCreateNamespace.

Makefile
jackye1995: Addressed in 3a87165: removed docker-test-fts-v2 and the PYTEST_ARGS Makefile plumbing. The integration assertion remains available when LANCE_FTS_FORMAT_VERSION=2 is set, without adding a dedicated one-off make target.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
jackye1995: Addressed in 3a87165: fragment-based index jobs now designate a single task to return the IndexBuildResult, and the driver accepts the first returned metadata instead of validating/transporting details from every fragment task.

PR #487, opened 2026-04-28
test: add Glue Docker integration coverage

jackye1995:
## Summary
- align Spark bundles with Lance Namespace 0.7.2 core and Apache client artifacts
- load the published lance-namespace-glue 0.3.0 bundle in Docker tests instead of staging local impl jars
- add an AWS Glue backend for integration tests against real Glue and S3
- run the Glue/S3 Docker integration in CI on the latest supported Spark/Scala pair using AWS secrets
- verify the AWS job reaches real STS, S3, and Glue APIs before Docker tests, and clean up the S3 test root afterward
- avoid reconnecting Glue namespace clients from Spark worker scan tasks

PR #488, opened 2026-04-28
test: add Glue Docker integration coverage

jackye1995:
## Summary
- align Spark bundles with Lance Namespace 0.7.2 core and Apache client artifacts
- load the published lance-namespace-glue 0.3.0 bundle in Docker tests instead of staging local impl jars
- add an AWS Glue backend for integration tests against real Glue and S3
- run the Glue/S3 Docker integration in same-repo PR CI on the latest supported Spark/Scala pair using AWS secrets
- verify the AWS job reaches real STS, S3, and Glue APIs before Docker tests, and clean up the S3 test root afterward
- avoid reconnecting Glue namespace clients from Spark worker scan tasks


jackye1995:
Closing this duplicate. It was opened only to run the AWS Glue/S3 workflow as same-repo PR CI; that run passed and is now reflected on the merged #487 check list.

PR #492, opened 2026-04-29
fix: reject DECIMAL256 columns at schema resolution time with actiona…

summaryzb:
## Problem
When a Lance dataset contains a DECIMAL256 column (256-bit decimal, precision > 38), reading it through the Spark connector previously fell through to Spark's upstream `ArrowUtils.fromArrowField`, which would throw a cryptic, non-actionable `UnsupportedOperationException` deep inside the data-reading path. 

## Approach
This PR intercepts that case at schema resolution time in `LanceArrowUtils.fromArrowField` and raises a clear `SparkUnsupportedOperationException` with the column name, bit-width, and a concrete remediation hint.

## Test Coverage
- DECIMAL256 field (precision=38, scale=5, bitWidth=256) passed to `fromArrowField` throws `SparkUnsupportedOperationException`.
- Exception message contains the column name (`amount`), the unsupported bit-width (`256`), the supported bit-width hint (`128`), and the precision limit (`38`).


hamersaw (approved):
Actionable failures will make this a much better experience, thanks for the PR!

Issue #419, opened 2026-04-10
ADD COLUMNS FROM should advance _row_last_updated_at_version for materialized rows


jerryjch:
## Summary

ADD COLUMNS FROM commits a new dataset version but does not advance
_row_last_updated_at_version on rows in fragments that were rewritten to materialize
the new columns.

ALTER TABLE ... ADD COLUMNS ... FROM (AddColumnsBackfill in lance-spark) backfills
from a source view keyed by row identity. The table version increases, but per-row
change-data metadata _row_last_updated_at_version can stay the same as before the
commit even though every row in those fragments received new column values.

## Expected behavior

For stable row IDs and CDF columns enabled, when ADD COLUMNS FROM rewrites a
fragment so that all physical rows receive the new column values,
_row_last_updated_at_version for those rows should reflect the dataset version of
that commit. _row_created_at_version should remain the version at which each row
first appeared.

## Actual behavior

Merged fragments from executors can lose or fail to update version sequences;
_row_last_updated_at_version may not move forward consistently for ADD COLUMNS
rewrites. (See update_version_issue.txt ADD COLUMNS section.)

## Reproduction

1. Create a Lance table with stable row IDs enabled (enable_stable_row_ids).
2. Insert rows (single fragment is enough); record _row_last_updated_at_version and dataset version.
3. Run ALTER TABLE ... ADD COLUMNS ... FROM (backfill new columns from a view keyed by _rowaddr/_fragid).
4. After commit, scan _row_last_updated_at_version for rows in rewritten fragments: values may not
   advance to the new dataset version even though new column data appeared.

## Note
  This issue is on top of and depends on
https://github.com/lance-format/lance/issues/6464 — JNI loses version metadata; incorrect stable row ID
lookup in Operation::Update (see fix PRs such as https://github.com/lance-format/lance/pull/6465).

jerryjch:
Fixed. Refer to https://github.com/lance-format/lance-spark/pull/503

PR #503, opened 2026-05-05
test: add columns from version metadata end-to-end validation

jerryjch:
Add a test to verify https://github.com/lance-format/lance-spark/issues/419 after https://github.com/lance-format/lance/pull/6640 and https://github.com/lance-format/lance/pull/6465
The test will fail before we bump the lance version to include the above two lance PRs, and should pass after we bump the lance version.


hamersaw (approved):
I updated lance version to 7.0.0-beta.7 so tests here pass! In that PR we updated quite a few others versions in the tests to adhere to the new (correct) behavior. Two questions:
(1) Is these tests still necessary? Or are we just duplicating other test logic.
(2) Do we need to be more extensive? I see this tests the backfill add column, what about backfill update column?

jerryjch:
@hamersaw Thanks for the review.
The added test is specifically for lance-spark backfill add columns test, which is not covered in other repo or PRs. This test is the only thing we need here, no code fix needed here in lance-spark since the fix in the lance.
For backfill update columns, we need some code fix here in lance-spark to hook-up to and use the lance fix. I plan to have a separate PR to contain a code fix and test.

hamersaw:
> @hamersaw Thanks for the review. The added test is specifically for lance-spark backfill add columns test, which is not covered in other repo or PRs. This test is the only thing we need here, no code fix needed here in lance-spark since the fix in the lance. For backfill update columns, we need some code fix here in lance-spark to hook-up to and use the lance fix. I plan to have a separate PR to contain a code fix and test.

@jerryjch sounds great! I think this is a great opportunity to improve the test coverage of this feature-space across the repo. I added some more tests in https://github.com/lance-format/lance-spark/pull/523 and found two bugs, namely https://github.com/lance-format/lance/issues/6734 and https://github.com/lance-format/lance/issues/6735. I'll file followup spark issues when these are fixed in the main repo.

jerryjch:
@hamersaw https://github.com/lance-format/lance/issues/6734https://github.com/lance-format/lance/issues/6734 is dup of https://github.com/lance-format/lance/issues/6505?

PR #511, opened 2026-05-08
feat(cbo): expose ColumnStatistics with conservative NDV from zonemap

### LINKED ISSUE #510: Enable Spark CBO with column statistics from zonemap
## Background

Spark's Cost-Based Optimizer (CBO) consumes per-column `ColumnStatistics` (min, max, nullCount, distinctCount, avgLen, maxLen) via the DSv2 `Statistics.columnStats()` API to drive join reordering, BroadcastHashJoin vs SortMergeJoin selection, AQE coalesce decisions, and filter selectivity estimation. Today, lance-spark exposes `numRows` and `sizeInBytes` only — no per-column stats — so Spark CBO falls back to row-count heuristics on every query against a Lance table.

Lance already has zonemap indexes that store per-zone min/max/nullCount per column. We can aggregate those into a `ColumnStatistics` map at scan-build time without any new Lance-core SDK surface and without any new I/O on top of what filter pushdown already loads.

## Proposal

Two complementary additions:

**ColumnStatistics from zonemap aggregation**
- New `ColumnStatsAggregator` reduces a list of per-zone `ZoneStats` to a single DSv2 `ColumnStatistics` (min / max / nullCount).
- `LanceStatistics.columnStats()` exposes the aggregated map to Spark's CBO.
- `LanceScanBuilder` loads zonemap stats for projected zonemap-indexed columns up to a cap (`max.columns`, default 8). Filter columns are always loaded (they already drive fragment pruning), so the cap only bounds extra projection-driven I/O.
- Master kill-switch + per-scan option for safe rollback.

**Conservative NDV from zonemap pattern**
- Small helper inside `ColumnStatsAggregator`: when every non-null zone has `min == max` (single distinct value per zone), `distinctCount = number of distinct zone-min values`. When the precondition fails, returns `null` so CBO falls back to row-count heuristics — no guessing.
- Most useful for sorted/clustered low-cardinality columns (`d_year`, `ca_state`, `cd_marital_status`). Zero extra I/O on top of what zonemap stats already loaded.

The NDV addition is a small helper inside the aggregator that shares the kill-switch, config keys, and I/O path. Splitting it out would create a strictly worse intermediate state where the aggregator always returns `null` for `distinctCount` despite already loading the data needed to compute it.

## Out of scope (follow-ups)

- `avgLen` / `maxLen` for variable-length types — separate issue.
- Sketch-based NDV via T-Digest or HLL on unsorted columns — requires a Lance-core SDK extension; separate issue.
- `ANALYZE TABLE` and persistent column stats in TBLPROPERTIES — separate issue, builds on this.
- Histograms — depend on sketch-based NDV; separate issue.

## Configuration

| SparkConf | Per-scan option | Default | Purpose |
|---|---|---|---|
| `spark.lance.cbo.column.stats.enabled` | `lance.cbo.column.stats.enabled` | `true` | Master kill-switch. |
| `spark.lance.cbo.column.stats.max.columns` | `lance.cbo.column.stats.max.columns` | `8` | Cap on columns whose zonemap stats are loaded at plan time. |

Per-scan options win over SparkConf. Master switch off restores prior behavior (no `columnStats()`).

## Risks and mitigations

- **Zonemap I/O at plan time** — bounded by `max.columns` cap. Filter columns are always loaded (they drive pruning regardless). Tunable per workload.
- **Stats only available for zonemap-indexed columns** — documented limitation. A future `ANALYZE TABLE` follow-up provides a fallback path for un-indexed columns.
- **NDV accuracy** — the NDV path returns `null` rather than guess on multi-valued zones. CBO degrades gracefully to current behavior on those columns.
- **Format/serialization stability** — no on-disk format added.

## PR

#511.

LuciferYang:
Closes #510.

## Summary

- New `ColumnStatsAggregator` reduces per-zone `ZoneStats` into DSv2 `ColumnStatistics` with min / max / nullCount + a conservative NDV when every non-null zone has `min == max`.
- `LanceStatistics.columnStats()` exposes the aggregated map to Spark's CBO via the standard `Statistics.columnStats()` API.
- `LanceScanBuilder` loads zonemap stats for projected zonemap-indexed columns up to `lance.cbo.column.stats.max.columns` (default 8). Filter and partition columns are always loaded — they already drive fragment pruning.
- Master kill-switch (`spark.lance.cbo.column.stats.enabled`, default `true`) + per-scan option for safe rollback. Both flow through `parseTypedFlags` so catalog-level configs propagate the same way as #476.

## Configuration

| SparkConf | Per-scan option | Default | Purpose |
|---|---|---|---|
| `spark.lance.cbo.column.stats.enabled` | `lance.cbo.column.stats.enabled` | `true` | Master kill-switch. Disables `columnStats()` reporting entirely. |
| `spark.lance.cbo.column.stats.max.columns` | `lance.cbo.column.stats.max.columns` | `8` | Cap on zonemap-indexed columns whose stats are loaded at plan time. Filter and partition columns are always loaded — the cap only bounds extra projection-driven I/O. |

Per-scan options win over SparkConf. Master switch off restores prior behavior (no `columnStats()` reported).

## Related

- #513 — adds `USING zonemap` as a recognized `CREATE INDEX` method. Not a strict prerequisite (this PR's tests don't depend on it), but pairing them lets users opt into a zonemap-only index for CBO column stats without paying btree's index-build cost.

## What's NOT in this PR (follow-ups)

- `avgLen` / `maxLen` for variable-length types — separate issue.
- Sketch-based NDV on unsorted columns — needs a Lance-core RFC.
- `ANALYZE TABLE` and persistent column statistics — separate follow-up PR; depends on this one.

## Test plan

- [ ] `make test SPARK_VERSION=3.5 SCALA_VERSION=2.12` — `ColumnStatsAggregatorTest` (17), `LanceStatisticsTest` (19), `LanceSparkReadOptionsSerializationTest` (8) all pass.
- [ ] `make test SPARK_VERSION=3.4 SCALA_VERSION=2.12` — same suite passes.
- [ ] `make test SPARK_VERSION=4.0 SCALA_VERSION=2.13` — same suite passes.
- [ ] `make test SPARK_VERSION=4.1 SCALA_VERSION=2.13` — same suite passes.
- [ ] `make lint` — Checkstyle + Spotless clean.
- [ ] Manual: query a Lance table with `EXPLAIN COST` and confirm `columnStats` lines appear under the `Statistics` block when zonemap indexes exist on filter columns.
- [ ] Manual: set `spark.lance.cbo.column.stats.enabled=false` and confirm `columnStats` disappears (kill-switch verification).

## Risks

- Adds a plan-time per-column zonemap-stats fetch on top of the filter-column loading that already happens. Bounded by `max.columns` cap (default 8). On wide tables with no filter columns, this is the only added I/O.
- NDV path is precondition-gated and returns `null` on failure — never poisons CBO with a guessed value.

### DIFF
--- a/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/ZonemapColumnStatsTest.java
+++ b/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/ZonemapColumnStatsTest.java
@@ -1,0 +1,17 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.read;
+
+/** Spark 3.4 runner for {@link BaseZonemapColumnStatsTest}. */
+public class ZonemapColumnStatsTest extends BaseZonemapColumnStatsTest {}
--- a/lance-spark-3.5_2.12/src/test/java/org/lance/spark/read/ZonemapColumnStatsTest.java
+++ b/lance-spark-3.5_2.12/src/test/java/org/lance/spark/read/ZonemapColumnStatsTest.java
@@ -1,0 +1,17 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.read;
+
+/** Spark 3.5 runner for {@link BaseZonemapColumnStatsTest}. */
+public class ZonemapColumnStatsTest extends BaseZonemapColumnStatsTest {}
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
@@ -58,6 +58,16 @@ public class LanceSparkReadOptions implements Serializable {
   public static final String CONFIG_BATCH_SIZE = "batch_size";
   public static final String CONFIG_TOP_N_PUSH_DOWN = "topN_push_down";
 
+  /** Per-scan kill-switch for CBO column-stats reporting. */
+  public static final String CONFIG_CBO_COLUMN_STATS_ENABLED = "cbo.column.stats.enabled";
+
+  /**
+   * Per-scan cap on the number of projected columns for which we load and aggregate zonemap stats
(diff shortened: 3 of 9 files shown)


LuciferYang:
Still a draft.

PR #513, opened 2026-05-08
feat(sql): support USING zonemap with distributed and consolidated build paths

### LINKED ISSUE #512: Recognize USING zonemap as a CREATE INDEX method
## Background

`ALTER TABLE ... CREATE INDEX ... USING <method>` currently accepts `btree` and `fts` only (see `IndexUtils.buildIndexType` in `AddIndexExec.scala`). Lance's underlying `IndexType` enum already has a `ZONEMAP` value, but there is no way to request a zonemap index from SQL today.

This matters because **btree does NOT produce a zonemap as a side-effect.** Verified in lance-core:

- `BTreeIndexPlugin::index_type()` → `IndexType::BTree` (`rust/lance-index/src/scalar/btree.rs:1474`).
- BTree training writes only `page_lookup.lance` + `page_data.lance` (`btree.rs:71-72`, all `new_index_file(...)` call sites at `btree.rs:1893,1942,1947,2206`); it never writes `zonemap.lance`.
- `Dataset.getZonemapStats(col)` filters by `desc.index_type().to_lowercase().contains("zonemap")` (`java/lance-jni/src/blocking_dataset.rs:3437–3439`) and then opens `zonemap.lance` directly — so a btree-typed entry does not satisfy the filter, and even if it did, the file lookup would fail.

Net: until this issue is resolved, the only ways to get a zonemap on a Lance dataset are (a) call the lower-level Lance Java SDK directly, or (b) skip zonemap entirely and accept that lance-spark's CBO column-statistics feature has no source for per-column min/max/nullCount on tables created via SQL.

## Why expose it

A zonemap-only index is meaningfully cheaper than btree:

- Smaller on-disk footprint (one `zonemap.lance` file with per-zone summaries; no per-page lookup tree).
- Faster build time.
- Sufficient for: range / equality fragment pruning at scan time, runtime / dynamic-file-pruning filters, and column-statistics reporting to Spark's CBO.

When the workload doesn't need fast single-row point lookups, btree's overhead is wasted — and even then, btree alone does not enable CBO column statistics, which require a zonemap.

## Proposal

Recognize `zonemap` as a method name in both directions of the `IndexUtils` mapping:

```scala
def buildIndexType(method: String): IndexType = method.toLowerCase match {
  case "btree"   => IndexType.BTREE
  case "fts"     => IndexType.INVERTED
  case "zonemap" => IndexType.ZONEMAP   // <-- added
  case other     => throw new UnsupportedOperationException(...)
}

def buildScalarIndexParamType(method: String): String = method.toLowerCase match {
  case "btree"   => "btree"
  case "fts"     => "inverted"
  case "zonemap" => "zonemap"           // <-- added
  case other     => throw new UnsupportedOperationException(...)
}
```

The PR also needs to skip `dataset.mergeIndexMetadata(...)` for `IndexType.ZONEMAP`, because lance-core's `merge_index_metadata` has no `ZoneMap` arm and would throw `Unsupported index type` at runtime; per-fragment `zonemap.lance` files are independently sufficient for `getZonemapStats`.

## Test plan

- New `IndexUtilsTest` in `lance-spark-base_2.12/src/test/java/...`:
  - All three method names map to the correct `IndexType`.
  - Lookup is case-insensitive (already the existing contract).
  - Reverse lookup `buildScalarIndexParamType` round-trips for all three.
  - Unknown methods still throw `UnsupportedOperationException`.
- Integration tests in `BaseAddIndexTest` covering: zonemap creation succeeds end-to-end; `getZonemapStats` is empty after a btree-only build; same-name `USING btree` then `USING zonemap` is last-write-wins; reverse direction; different names coexist.

## Out of scope

This issue is scope-limited to recognizing the keyword and making the build pipeline tolerant of `IndexType.ZONEMAP`. Tooling discussion (Spark UI / `SHOW INDEXES` rendering, default index choice for CBO column statistics) belongs in separate issues.

## PR

#513.

### LINKED ISSUE #514: Distributed scalar-index build races on per-fragment zonemap.lance
## Background

`AddIndexExec`'s `FragmentBasedIndexJob` spawns one Spark task per fragment, all sharing a single index UUID via `IndexOptions.withIndexUUID(uuid)`. On the lance-core side this maps each task's per-fragment build to the same `indices_dir().join(uuid)` directory (`rust/lance/src/index/scalar.rs:287` — `LanceIndexStore::from_dataset_for_new(dataset, uuid)`).

For index types that emit a **single artifact per build**, all N executors race to write the same object-store path. For ZONEMAP that artifact is `zonemap.lance` (`rust/lance-index/src/scalar/zonemap.rs:50,745`); whichever task commits last wins, and the read path (`getZonemapStats`) returns only one fragment's worth of zone stats.

## Reproduction

`AddIndexTest.testCreateZonemapIndex` on a 20-fragment table (Spark `local[10]`, 20 single-row inserts):

```java
Set<Integer> fragmentIdsWithStats =
    zonemapStats("id").stream()
        .map(ZoneStats::getFragmentId)
        .collect(Collectors.toSet());
// expected: 20, actual: 1
```

This was the empirical confirmation that surfaced during review of #513. The current PR keeps the loose `assertFalse(zonemapStats("id").isEmpty())` and documents the limitation.

## Why this is also a BTree concern

The same code path is used for `USING btree` (default fragment build_mode). The race is functionally invisible there because btree's user-visible behavior is correctness-preserving: a fragment without an index entry simply gets a full scan — slower, never wrong. That's why the existing `testCreateIndexDistributed` has not caught it. The race is real, just silent.

## Fix options

**(A) Single-call driver-side build for affected types.** Have AddIndexExec call `dataset.createIndex(opts)` once on the driver with all fragment IDs. `build_scalar_index` then loads training data across every fragment and produces a single consistent artifact. Loses per-fragment parallelism (acceptable for ZONEMAP — small payload — possibly not for large btree builds).

**(B) Scalar-index segment commit path.** Mirror lance-core's vector segment-commit pattern (`commit_existing_index_segments`, `rust/lance/src/dataset/index.rs:230`) for scalar indexes. Each task gets a fresh UUID and writes to its own directory; the driver collects N IndexMetadata segments and commits them as separate entries with the same name. `getZonemapStats` already supports multi-segment reads (`java/lance-jni/src/blocking_dataset.rs:3437–3470`). Requires a lance-core API extension.

**(C) Per-task UUID with manifest-side join.** Have lance-spark commit N IndexMetadata entries (different UUIDs, same name) directly via `AddIndexOperation.withNewIndices(List)`. The lance-core `describe_indices` already groups by name (`rust/lance/src/index.rs:957–962`), so consumers see one logical index. Most code-local to lance-spark.

I attempted (A) first but ran into manifest-visibility issues that need deeper Lance-runtime debugging (driver-side `createIndex` returns successfully but `getIndexes` reports zero entries — possibly session-cache opacity). (C) is the most attractive given existing lance-core infrastructure.

## Scope of this issue

- Investigate (A) vs (C); pick the smaller-blast-radius fix.
- Apply to ZONEMAP first (visible symptom), then BTree fragment-mode (silent symptom).
- Add a strict per-fragment-coverage test once the fix lands: `assertEquals(fragmentsIndexed, fragmentIdsWithStats.size())`.

## Related

- #513 — adds `USING zonemap` SQL support; documents this limitation in the PR but does not fix.
- #511 — uses `getZonemapStats` to feed Spark CBO column statistics; the race directly bounds the achievable coverage on un-ANALYZEd tables.

LuciferYang:
Closes #512. Closes #514.

Adds `ALTER TABLE ... CREATE INDEX ... USING zonemap (col)` with two build paths:

**Distributed** (default): one Lance segment per fragment, built in parallel. Each task uses a fresh per-task UUID so the per-task `<uuid>/zonemap.lance` files don't collide; the driver commits N IndexMetadata entries under the shared name in one transaction. Works against current upstream lance-core, no new APIs needed.

**Consolidated** (opt-in via `spark.lance.zonemap.consolidate.enabled=true`): workers compute per-fragment zone batches via `dataset.computeZonemapBatch` and return them to the driver; the driver merges them via `dataset.writeZonemapIndexFromBatches` and commits a single IndexMetadata entry covering every fragment.

## Creation time and footprint (sf=100 store_sales, `ss_sold_date_sk`)

Same dataset (234 fragments, 287,997,024 rows), same column, single zonemap index:

|                       | Distributed | Consolidated |
|-----------------------|-------------|--------------|
| Wall-clock            | **15.0 s**  | **28.1 s**   |
| Index segments        | 234         | 1            |
| `_indices/` bytes     | 1,099,920   | 137,835      |
| Per-segment file size | 4.7 KB avg  | 137 KB total |

The consolidated path is ~2× slower in wall-clock (driver-side write replaces parallel executor writes) but produces ~8× less on-disk index data because Lance's per-file header overhead is paid once instead of 234 times. Manifest grows by 1 IndexMetadata entry instead of 234.

Bulk-build of all 15 fact-table zonemaps on the four clustered tables (~903M rows across the four facts) takes **56 s** end-to-end via the consolidated path (default per-zone is `rows_per_zone=8192`).

## Dependency on lance-core

The consolidated path uses two new lance-core APIs that are in review upstream:

- lance-format/lance#6779 — Rust `compute_zonemap_batch` + `write_consolidated_zonemap_segment`
- lance-format/lance#6780 — Java JNI bindings, stacked on #6779

**Not mergeable until both land and a lance-core release with those APIs ships.** The distributed path alone works against current `<lance.version>` (no pom change in this PR).

## What changed

- `AddIndexExec.scala`: SQL `zonemap` method, `runZonemapDistributed`, `runZonemapConsolidated`, two per-fragment task classes, and a pre-commit check that the consolidated file's `fragment_bitmap` matches the dispatcher's fragment set.
- `LanceSparkReadOptions.java`: `spark.lance.zonemap.consolidate.enabled` (lenient parse: case-insensitive `"true"` after trim, anything else falls through to distributed).
- `docs/src/operations/ddl/create-index.md`: zonemap method, `rows_per_zone` option, consolidated SparkConf flag, single-column constraint.
- `integration-tests/test_lance_spark.py`: two pytest cases (default path + consolidated mode).

## Tests

`AddIndexTest`: 38 cases, green on Spark 3.5/2.12 and 4.0/2.13.

- Distributed: multi-segment commit shape (per-fragment IndexMetadata, distinct UUIDs, fragment-bitmap union equals the indexed set), `rows_per_zone` forwarding, N=1 degenerate case, idempotent re-create, two coexisting indexes on different columns.
- Consolidated: single-segment shape, cross-path replace (distributed→consolidated and the reverse), flag parsing edge cases (default-off, `"false"` routes to distributed, `"TRUE"` with whitespace), type coverage (int / long / string / nullable-string), `rows_per_zone` min/max correctness, end-to-end SELECT after CREATE INDEX.

`IndexUtilsTest`: 14 cases for the SQL → `IndexType` mapping and `toJson` wire-shape edge cases.

`integration-tests/test_lance_spark.py`: `test_create_zonemap_index` (default distributed path with `rows_per_zone=16`) and `test_create_zonemap_index_consolidated` (asserts exactly one IndexMetadata segment under the consolidated mode).

`mergeIndexMetadata` is skipped on both paths: nothing to merge for distributed (Lance's read path serves multi-segment natively), already merged in-memory for consolidated.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -16,8 +16,9 @@ package org.apache.spark.sql.execution.datasources.v2
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.node.ObjectNode
 import org.apache.arrow.c.{ArrowArrayStream, Data}
+import org.apache.arrow.memory.RootAllocator
 import org.apache.arrow.vector.VectorSchemaRoot
-import org.apache.arrow.vector.ipc.ArrowReader
+import org.apache.arrow.vector.ipc.{ArrowReader, ArrowStreamReader, ArrowStreamWriter}
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Attribute, GenericInternalRow}
 import org.apache.spark.sql.catalyst.plans.logical.{AddIndexOutputType, LanceNamedArgument}
@@ -35,6 +36,8 @@ import org.lance.spark.arrow.LanceArrowWriter
 import org.lance.spark.utils.{CloseableUtil, Utils}
 import org.lance.spark.write.SingleBatchArrowReader
 
+import java.io.{ByteArrayInputStream, ByteArrayOutputStream}
+import java.nio.channels.Channels
 import java.time.Instant
 import java.util.{Collections, Optional, UUID}
 
@@ -83,9 +86,28 @@ case class AddIndexExec(
     }
 
     val indexType = IndexUtils.buildIndexType(method)
+    // Parse the consolidate flag BEFORE opening the dataset handle. If the parse threw
+    // (currently can't, but defense-in-depth against future evolution of getOption/trim
+    // behaviour), opening the dataset first would leak the JNI handle past the unwind.
+    // SparkConf-only flag for now. Default false preserves the multi-segment distributed
+    // shape that has been the only path for the entire history of this code path; flipping
+    // the default would silently change the commit footprint of every existing pipeline.
+    // Lenient parse: `_.toBoolean` rejects "yes"/"on"/"1" with IllegalArgumentException —
+    // surfacing a plan-time crash for a config typo is worse than degrading to the safe
+    // default, since the failure mode obscures the only corrective action (fix the value).
+    // Accept the JVM's canonical "true" (case-insensitive); everything else routes to the
+    // distributed path.
+    val consolidateZonemap =
+      indexType == IndexType.ZONEMAP &&
+        session.conf.getOption("spark.lance.zonemap.consolidate.enabled")
+          .exists(v => v != null && "true".equalsIgnoreCase(v.trim))
+
     val dataset = Utils.openDatasetBuilder(readOptions).build()
 
     if (indexType == IndexType.ZONEMAP) {
+      if (consolidateZonemap) {
+        return runZonemapConsolidated(dataset, lanceDataset, readOptions, fragmentIds)
+      }
       return runZonemapDistributed(dataset, lanceDataset, readOptions, fragmentIds)
     }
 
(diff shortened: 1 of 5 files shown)


LuciferYang:
It's still a draft, please don't review it for now.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -82,26 +85,56 @@ case class AddIndexExec(
...
+    val consolidateZonemap =
+      indexType == IndexType.ZONEMAP &&
+        session.conf.getOption("spark.lance.zonemap.consolidate.enabled")
+          .exists(v => v != null && "true".equalsIgnoreCase(v.trim))
 
     val dataset = Utils.openDatasetBuilder(readOptions).build()
 
+    if (indexType == IndexType.ZONEMAP) {
+      if (consolidateZonemap) {
+        return runZonemapConsolidated(dataset, lanceDataset, readOptions, fragmentIds)
```
LuciferYang, line 109: The consolidated path uses two new Lance APIs that are in review upstream:

- https://github.com/lance-format/lance/pull/6779 — Rust compute_zonemap_batch + write_consolidated_zonemap_segment
- https://github.com/lance-format/lance/pull/6780 — Java JNI bindings, stacked on #6779

LuciferYang:
cc @hamersaw this PR depends on

- https://github.com/lance-format/lance/pull/6779 — Rust compute_zonemap_batch + write_consolidated_zonemap_segment
- https://github.com/lance-format/lance/pull/6780 — Java JNI bindings, stacked on #6779

to build successfully.

PR #466, opened 2026-04-21
feat: support zonemap indexes in ALTER TABLE CREATE INDEX

beinan:
## Summary
- add single-column zonemap support for `ALTER TABLE ... CREATE INDEX` via Lance direct `createIndex` instead of fragment training
- surface zonemap indexes in `SHOW INDEXES` and scan planning, and fix numeric zonemap pruning across mixed numeric types
- clarify and test that multi-column zonemap is not yet supported, with a clearer Spark-side error and updated docs

## Testing
- `./mvnw -pl lance-spark-4.0_2.13,lance-spark-4.1_2.13 -Dtest=AddIndexTest,ShowIndexesTest,ZonemapFragmentPrunerTest,CreateIndexStandardSyntaxTest -Dsurefire.failIfNoSpecifiedTests=false test`
- `./mvnw -pl lance-spark-4.0_2.13,lance-spark-4.1_2.13 -Dtest=AddIndexTest,ShowIndexesTest -Dsurefire.failIfNoSpecifiedTests=false test`

## Notes
- generic Spark `CREATE INDEX` syntax accepts a column list, but current Lance core rejects multi-column zonemap creation with `LanceError(Index): Only support building index on 1 column at the moment`
- Spark now fails earlier with `Zonemap index currently supports a single column only`
- Python integration tests were updated but not run locally because this environment does not have `pytest` or the expected `/home/lance/data` fixture path

PR #473, opened 2026-04-23
feat: add distributed zonemap index build

beinan:
## Summary

This PR adds a distributed zonemap index build path for  in .

It keeps the SQL surface unchanged, but changes the execution strategy from the driver-only/single-process flow to Spark-side fragment processing and logical segment commit.

This is intended as the distributed alternative to #466, which stays open as the simpler single-process solution.

## What Changed

- build zonemap index segments on Spark tasks and commit them through 
- support Spark-side zonemap pruning for segmented zonemap indices
- batch multiple fragments into a single zonemap segment so task and segment counts stay bounded
- keep zonemap restricted to single-column indexing
- update docs and tests for distributed zonemap behavior

## Validation

Tested with Spark 4.0 / Scala 2.13:

[INFO] Scanning for projects...
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-3.5_2.12:jar:0.4.0-beta.3
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-shade-plugin is missing. @ line 73, column 21
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-3.4_2.12:jar:0.4.0-beta.3
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-shade-plugin is missing. @ line 76, column 21
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-3.5_2.13:jar:0.4.0-beta.3
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-shade-plugin is missing. @ line 77, column 21
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-3.4_2.13:jar:0.4.0-beta.3
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-shade-plugin is missing. @ line 78, column 21
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-4.0_2.13:jar:0.4.0-beta.3
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-shade-plugin is missing. @ line 79, column 21
[WARNING] 
[WARNING] Some problems were encountered while building the effective model for org.lance:lance-spark-bundle-4.1_2.13:jar:0.4.0-beta.3
[...]

PR #516, opened 2026-05-11
feat: add distributed zonemap index build with configurable segments

beinan:
## Summary

- Add zonemap as a new index type in `CREATE INDEX` DDL with distributed build support
- Batch fragments into configurable segments via `num_segments` option (defaults to `spark.default.parallelism`)
- Each segment is built in parallel on Spark executors and committed as a logical index on the driver
- Zonemap indexes currently support single column only

## What Changed

- `AddIndexExec.scala`: Zonemap-specific path with `ZonemapIndexJob`/`ZonemapIndexTask` and `commitIndexSegments`
- `create-index.md`: Document zonemap index type, options, and usage
- Tests: unit tests for segment creation/validation and integration test

## Notes

- Rebased cleanly onto current `main`
- Depends on lance-core `7.0.0-beta.10` or newer which includes zonemap segment support
- Supersedes PR #473 and closed PR #466

## Test plan

- [x] CI passes (lint, unit tests, integration tests across all Spark/Scala versions)
- [x] Zonemap index creation with default segment count
- [x] Zonemap index creation with explicit `num_segments`
- [x] Repeated zonemap index creation replaces existing segments
- [x] Query correctness after zonemap index creation

### DIFF
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -7,7 +7,7 @@ Creates a scalar index on a Lance table to accelerate queries.
 
 ## Overview
 
-The `CREATE INDEX` command builds an index on one or more columns of a Lance table. Indexing can improve the performance of queries that filter on the indexed columns. This operation is performed in a distributed manner, building indexes for each data fragment in parallel.
+The `CREATE INDEX` command builds an index on one or more columns of a Lance table. Indexing can improve the performance of queries that filter on the indexed columns. Depending on the index method, Lance Spark either uses a fragment-parallel build path or a driver-coordinated commit flow after parallel executor builds.
 
 ## Basic Usage
 
@@ -24,12 +24,22 @@ The following index methods are supported:
 
 | Method  | Description                                                                 |
 |---------|-----------------------------------------------------------------------------|
+| `zonemap` | Lightweight min/max index for fragment pruning on a scalar column. |
 | `btree` | B-tree index for efficient range queries and point lookups on scalar columns. |
 | `fts`   | Full-text search (inverted) index for text search on string columns.        |
 
 ## Options
 
 The `CREATE INDEX` command supports options via the `WITH` clause to control index creation. These options are specific to the chosen index method.
+
+### ZoneMap Options
+
+For the `zonemap` method, the following options are supported:
+
+| Option          | Type | Description                                  |
+|-----------------|------|----------------------------------------------|
+| `rows_per_zone` | Long    | The approximate number of rows per zonemap zone. |
+| `num_segments`  | Integer | Target number of index segments (upper bound; clamped to fragment count when larger). Each segment covers a batch of fragments. Defaults to `min(fragment_count, spark.default.parallelism)`. |
 
 ### BTree Options
 
@@ -92,6 +102,15 @@ Create a composite index on multiple columns.
     ALTER TABLE lance.db.logs CREATE INDEX idx_ts_level USING btree (timestamp, level);
     ```
 
+### Lightweight Fragment Pruning
+
+Create a zonemap index when you want lightweight min/max-based fragment pruning:
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE INDEX idx_id_zonemap USING zonemap (id);
+    ```
+
 ### Indexing with Options
 
 Create an index and specify the `zone_size` for the B-tree:
@@ -99,6 +118,15 @@ Create an index and specify the `zone_size` for the B-tree:
 === "SQL"
     ```sql
     ALTER TABLE lance.db.users CREATE INDEX idx_id_zoned USING btree (id) WITH (zone_size = 2048);
+    ```
+
+### Zonemap with Options
+
+Create a zonemap index and specify the approximate number of rows per zone:
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE INDEX idx_id_zonemap USING zonemap (id) WITH (rows_per_zone = 2048);
     ```
 
 ### Full-Text Search Index
@@ -133,17 +161,19 @@ The `CREATE INDEX` command returns the following information about the operation
 Consider creating an index when:
 
 - You frequently filter a large table on a specific column.
+- You want lightweight fragment pruning based on per-zone min/max statistics.
 - Your queries involve point lookups or small range scans.
 
 ## How It Works
 
 The `CREATE INDEX` command operates as follows:
 
-1.  **Distributed Index Building**: For each fragment in the Lance dataset, a separate task is launched to build an index on the specified column(s).
-2.  **Metadata Merging**: Once all per-fragment indexes are built, their metadata is collected and merged.
+1.  **Index Build Execution**: Lance Spark chooses an execution path based on the index method. Methods such as `btree`, `fts`, and `zonemap` can build physical index segments in parallel across fragments. `zonemap` publishes those segments directly as one logical index. Range-mode `btree` uses Spark repartitioning and sorted preprocessed data.
+2.  **Metadata Finalization**: Lance Spark merges or commits the resulting index metadata on the driver so the new logical index becomes visible atomically.
 3.  **Transactional Commit**: A new table version is committed with the new index information. The operation is atomic and ensures that concurrent reads are not affected.
 
 ## Notes and Limitations
 
-- **Index Methods**: The `btree` and `fts` methods are supported for scalar index creation.
-- **Index Replacement**: If you create an index with the same name as an existing one, the old index will be replaced by the new one. This is because the underlying implementation uses `replace(true)`.
+- **Index Methods**: The `zonemap`, `btree`, and `fts` methods are supported for scalar index creation.
+- **Zonemap Column Count**: Zonemap indexes currently support a single column only. The generic `CREATE INDEX` grammar accepts a column list, but Lance rejects multi-column zonemap creation.
+- **Index Replacement**: If you create an index with the same name as an existing one, the old index will be replaced by the new one.
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -18,6 +18,7 @@ import com.fasterxml.jackson.databind.node.ObjectNode
 import org.apache.arrow.c.{ArrowArrayStream, Data}
 import org.apache.arrow.vector.VectorSchemaRoot
 import org.apache.arrow.vector.ipc.ArrowReader
+import org.apache.spark.internal.Logging
 import org.apache.spark.sql.catalyst.InternalRow
 import org.apache.spark.sql.catalyst.expressions.{Attribute, GenericInternalRow}
 import org.apache.spark.sql.catalyst.plans.logical.{AddIndexOutputType, LanceNamedArgument}
@@ -32,7 +33,7 @@ import org.lance.index.scalar.{BTreeIndexParams, ScalarIndexParams}
 import org.lance.operation.{CreateIndex => AddIndexOperation}
 import org.lance.spark.{BaseLanceNamespaceSparkCatalog, LanceDataset, LanceRuntime, LanceSparkReadOptions}
 import org.lance.spark.arrow.LanceArrowWriter
-import org.lance.spark.utils.{CloseableUtil, ParserUtils, Utils}
+import org.lance.spark.utils.{CloseableUtil, Utils}
 import org.lance.spark.write.SingleBatchArrowReader
 
 import java.time.Instant
@@ -44,9 +45,9 @@ import scala.collection.JavaConverters._
  * Physical execution of distributed CREATE INDEX (ALTER TABLE ... CREATE INDEX ...) for Lance datasets.
  *
  * <ul>
- * <li>For BTREE index, it uses a range-based approach that redistributes and sorts data across partitions, creates indexes for each range in parallel, and finally merges them into a global index structure.
- * <li>For other index types, it processes each fragment independently in parallel, merges index metadata
- * and commits an index-creation transaction.
+ * <li>For zonemap index, it builds uncommitted index segments in parallel and commits the logical index on the driver.
+ * <li>For BTREE index, range mode redistributes and sorts data across partitions, creates indexes for each range in parallel, and finally merges them into a global index structure.
+ * <li>For fragment-trainable scalar index types (BTREE fragment mode, FTS, etc.), it processes each fragment independently in parallel, merges index metadata and commits an index-creation transaction.
  * </ul>
  */
 case class AddIndexExec(
@@ -82,92 +83,119 @@ case class AddIndexExec(
       return Seq(new GenericInternalRow(Array[Any](0L, UTF8String.fromString(indexName))))
     }
 
-    val uuid = UUID.randomUUID()
     val indexType = IndexUtils.buildIndexType(method)
 
+    if (indexType == IndexType.ZONEMAP && columns.size != 1) {
+      throw new UnsupportedOperationException(
+        "Zonemap index currently supports a single column only")
+    }
+
+    val btreeBuildMode = IndexUtils.btreeBuildMode(indexType, args)
+    val useLogicalSegmentCommit = IndexUtils.useLogicalSegmentCommit(indexType)
+
+    val numSegmentsOpt = args.find(_.name == "num_segments")
+    if (numSegmentsOpt.isDefined && !useLogicalSegmentCommit) {
+      throw new IllegalArgumentException(
+        "num_segments option is only supported for index types that use segmented builds (e.g., zonemap)")
+    }
+    val validatedNumSegments: Option[Int] = numSegmentsOpt.map { arg =>
+      arg.value match {
+        case null =>
+          throw new IllegalArgumentException(
+            "num_segments must be a positive integer, got: null")
+        case n: Number =>
+          val asLong = n.longValue()
+          if (asLong < 1L || asLong > Int.MaxValue)
+            throw new IllegalArgumentException(
+              s"num_segments must be a positive integer that fits in Int, got: $asLong")
+          asLong.toInt
+        case other =>
+          throw new IllegalArgumentException(
+            s"num_segments must be a positive integer, got: $other")
+      }
+    }
+
+    // Zonemap uses logical segment commit path
+    if (useLogicalSegmentCommit) {
+      val (nsImpl, nsProps, tableId, initialStorageOpts) =
+        extractNamespaceInfo(lanceDataset, readOptions)
+      val zonemapJob = new ZonemapIndexJob(
+        this,
+        readOptions,
+        fragmentIds,
+        validatedNumSegments,
+        nsImpl,
+        nsProps,
+        tableId,
+        initialStorageOpts)
+      val segments = zonemapJob.run()
+      // Atomic add+remove via Lance core; see commitIndexSegments
+      commitIndexSegments(readOptions, columns.head, segments)
+      return Seq(new GenericInternalRow(Array[Any](
+        fragmentIds.size.toLong,
+        UTF8String.fromString(indexName))))
+    }
+
+    val uuid = UUID.randomUUID()
     val dataset = Utils.openDatasetBuilder(readOptions).build()
 
+    val indexBuildResult =
+      createIndexJob(
+        dataset,
+        lanceDataset,
+        readOptions,
+        uuid.toString,
+        fragmentIds,
+        btreeBuildMode).run()
+
     try {
-      if (indexType == IndexType.ZONEMAP) {
-        // ZoneMap indexes are lightweight (metadata-only).
-        // Build per-fragment segments on the driver and commit
-        // via the segment-based path, avoiding the merge path
-        // which doesn't support ZoneMap.
-        val argsJson = IndexUtils.toJson(args)
-        val segments = fragmentIds.map { fragId =>
-          val params = IndexParams.builder()
-            .setScalarIndexParams(
-              ScalarIndexParams.create("zonemap", argsJson))
-            .build()
-          val opts = IndexOptions
-            .builder(columns.asJava, IndexType.ZONEMAP, params)
-            .replace(true)
-            .withIndexName(indexName)
-            .withFragmentIds(Collections.singletonList(fragId))
-            .build()
-          dataset.createIndex(opts)
-        }
-        dataset.commitExistingIndexSegments(
-          indexName,
-          columns.head,
-          segments.asJava)
-      } else {
-        val indexBuildResult =
-          createIndexJob(
-            dataset,
-            lanceDataset,
-            readOptions,
-            uuid.toString,
-            fragmentIds).run()
-        // BTree/Inverted use the merge path
-        dataset.mergeIndexMetadata(uuid.toString, indexType, Optional.empty())
+      // Merge index metadata after all fragments are indexed
+      dataset.mergeIndexMetadata(uuid.toString, indexType, Optional.empty())
 
-        val fieldIdByName = dataset.getLanceSchema.fields().asScala
-          .map(f => f.getName -> f.getId)
-          .toMap
-        val fieldIds = columns.map { column =>
-          fieldIdByName.getOrElse(
-            column,
-            throw new IllegalArgumentException(
-              s"Cannot find index column in Lance schema: $column"))
-        }.toList
+      val fieldIdByName = dataset.getLanceSchema.fields().asScala
+        .map(f => f.getName -> f.getId)
+        .toMap
+      val fieldIds = columns.map { column =>
+        fieldIdByName.getOrElse(
+          column,
+          throw new IllegalArgumentException(s"Cannot find index column in Lance schema: $column"))
+      }.toList
 
-        val datasetVersion = dataset.version()
+      val datasetVersion = dataset.version()
 
-        val indexBuilder = Index
-          .builder()
-          .uuid(uuid)
-          .name(indexName)
-          .fields(fieldIds.map(java.lang.Integer.valueOf).asJava)
-          .datasetVersion(datasetVersion)
-          .indexDetails(indexBuildResult.indexDetails)
-          .indexVersion(indexBuildResult.indexVersion)
-          .indexType(indexBuildResult.indexType)
-          .fragments(fragmentIds.asJava)
-        indexBuildResult.createdAt.foreach(indexBuilder.createdAt)
-        val index = indexBuilder.build()
+      val indexBuilder = Index
+        .builder()
+        .uuid(uuid)
+        .name(indexName)
+        .fields(fieldIds.map(java.lang.Integer.valueOf).asJava)
+        .datasetVersion(datasetVersion)
+        .indexDetails(indexBuildResult.indexDetails)
+        .indexVersion(indexBuildResult.indexVersion)
+        .indexType(indexBuildResult.indexType)
+        .fragments(fragmentIds.asJava)
+      indexBuildResult.createdAt.foreach(indexBuilder.createdAt)
+      val index = indexBuilder.build()
 
-        // Find existing indices with the same name to mark as removed (for replace)
-        val removedIndices = dataset.getIndexes.asScala
-          .filter(_.name() == indexName)
-          .toList.asJava
+      // Find existing indices with the same name to mark as removed (for replace)
+      val removedIndices = dataset.getIndexes.asScala
+        .filter(_.name() == indexName)
+        .toList.asJava
 
-        val op = AddIndexOperation.builder()
-          .withNewIndices(Collections.singletonList(index))
-          .withRemovedIndices(removedIndices)
-          .build()
-        val txn = new Transaction.Builder()
-          .readVersion(dataset.version())
-          .operation(op)
-          .build()
-        try {
-          val newDataset = new CommitBuilder(dataset)
-            .writeParams(readOptions.getStorageOptions)
-            .execute(txn)
-          newDataset.close()
-        } finally {
-          txn.close()
-        }
+      val op = AddIndexOperation.builder()
+        .withNewIndices(Collections.singletonList(index))
+        .withRemovedIndices(removedIndices)
+        .build()
+      val txn = new Transaction.Builder()
+        .readVersion(dataset.version())
+        .operation(op)
+        .build()
+      try {
+        val newDataset = new CommitBuilder(dataset)
+          .writeParams(readOptions.getStorageOptions)
+          .execute(txn)
+        newDataset.close()
+      } finally {
+        txn.close()
       }
     } finally {
       dataset.close()
@@ -178,18 +206,32 @@ case class AddIndexExec(
       UTF8String.fromString(indexName))))
   }
 
-  private def createIndexJob(
-      dataset: Dataset,
-      lanceDataset: LanceDataset,
+  // Lance core's commitExistingIndexSegments handles atomic replacement:
+  // it finds existing segments whose fragments overlap with incoming ones
+  // and removes them in the same CreateIndex transaction.
+  private def commitIndexSegments(
       readOptions: LanceSparkReadOptions,
-      uuid: String,
-      fragmentIds: List[Integer]): IndexJob = {
-    // Get namespace info from catalog if available (for credential vending on workers)
-    val (nsImpl, nsProps, tableId, initialStorageOpts): (
-        Option[String],
-        Option[Map[String, String]],
-        Option[List[String]],
-        Option[Map[String, String]]) = catalog match {
+      column: String,
+      segments: Seq[Index]): Unit = {
+    val dataset = Utils.openDatasetBuilder(readOptions).build()
+    try {
+      dataset.commitExistingIndexSegments(
+        indexName,
+        column,
+        segments.toList.asJava)
+    } finally {
+      dataset.close()
+    }
+  }
+
+  private def extractNamespaceInfo(
+      lanceDataset: LanceDataset,
+      readOptions: LanceSparkReadOptions): (
+      Option[String],
+      Option[Map[String, String]],
+      Option[List[String]],
+      Option[Map[String, String]]) = {
+    catalog match {
       case nsCatalog: BaseLanceNamespaceSparkCatalog =>
         (
           Option(nsCatalog.getNamespaceImpl),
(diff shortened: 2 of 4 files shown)


beinan (commented):
Overall, the distributed architecture (segmenting fragments across Spark executors) is solid. However, there are several critical issues in the implementation that need to be addressed before merging:

### Critical Issues (P0 & P1)

1. **Dataset opened repeatedly in `pushFilters` (P0)**
   In `LanceScanBuilder.java`, opening a full dataset connection just to check index metadata during `pushFilters` is a significant performance bottleneck. `pushFilters` can be called multiple times during query planning. This check should be deferred to `build()` or cached.

2. **`useScalarIndex` logic disables ALL scalar indexes (P0)**
   In `LanceScanBuilder.java`, when `forcePostScanFiltering` is true, it sets `useScalarIndex` to false for the entire scan. This disables the usage of *all* scalar indexes (including BTREEs on other columns), not just the zonemap. The flag needs to be scoped strictly to the zonemap-indexed columns.

3. **No error handling/cleanup for segment failures (P1)**
   In `AddIndexExec.scala` (`executeZonemapIndex`), if `segmentRdd.mapPartitionsWithIndex(...).collect()` throws an exception because one partition fails, there is no cleanup for the segments that *did* succeed. This leaves orphaned segment artifacts. Consider adding a `try/catch` around `collect()` to clean up partial segments on failure (or document if Lance core handles this natively).

4. **Missing validation for `num_segments` (P1)**
   In `AddIndexExec.scala`, `val numSegments = options.getOrElse("num_segments", ...).toInt` has no bounds checks for `num_segments <= 0`, and `toInt` will throw a raw `NumberFormatException` on non-numeric input. This needs proper validation and a user-friendly error message.

### Moderate & Minor Issues (P2 & P3)

5. **Thread-local `SparkSession.active`:** `AddIndexExec` inherits `session` from `LeafExecNode`. Use `session` instead of `SparkSession.active` to avoid potential thread-local context issues.
6. **Serialization of `LanceConfig`:** Ensure that `LanceConfig` is strictly `Serializable` and doesn't wrap non-serializable Spark configuration objects, or the `ZonemapIndexTask` closure will fail at runtime.
7. **Empty partition overhead:** `FragmentUtils.batchFragments` might produce fewer batches than requested due to `math.ceil` rounding. `fragmentIter.flatten` handles this, but it wastes scheduler overhead.
8. **Unused Imports:** Remove `TaskContext` from `AddIndexExec.scala`.
9. **Scala Style:** `FragmentUtils.batchFragments` uses explicit `return` statements, which isn't idiomatic Scala.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -213,10 +216,20 @@ public Scan build() {
...
     }
 
+    // Detect segmented zonemap indexes and force post-scan filtering if needed.
+    // Done here (not in pushFilters) to reuse the already-open dataset.
+    forcePostScanFiltering = shouldForcePostScanFiltering(pushedFilters);
+    if (forcePostScanFiltering) {
+      LOG.info(
+          "Using Spark post-scan filtering for segmented zonemap query on dataset {}",
+          readOptions.getDatasetUri());
+    }
```
hamersaw: I'm not following why we need to decide which indexes are used at LanceScan construction time. Does this not get picked up by the existing ZoneMap pruning logic? We should be able to prune fragments based on ZoneMap indexes and then use scalar indexes to lookup. And there are drafted PRs to update Lance Spark partitions to fragment row ranges (rather than fragments) so we can have even finer granularity on ZoneMap pruning.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -413,4 +427,49 @@ private static Set<String> extractReferencedColumns(Filter[] filters) {
     }
     return columns;
   }
+
+  /**
+   * Segmented zonemap indexes are currently used safely for fragment pruning, but scan-time filter
```
hamersaw: Javadoc on shouldForcePostScanFiltering is truncated mid-sentence: The first line of the doc comment ends mid-thought: "Segmented zonemap indexes are currently used safely for fragment pruning, but scan-time filter" and is then followed by what looks like a separate sentence "Detects segmented zonemap indexes by ...". The original rationale (why post-scan filtering is needed at scan time) is lost, which makes the heuristic's correctness harder to audit later.

**Fix:** Complete the first sentence to state the actual reason segmented zonemaps cannot be relied on for scan-time filtering (so future readers know whether the heuristic still applies as the underlying Lance behavior evolves).

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -413,4 +427,49 @@ private static Set<String> extractReferencedColumns(Filter[] filters) {
...
+      String columnName = fieldIdToName.get(idx.fields().get(0));
+      if (columnName == null || !referencedColumns.contains(columnName)) {
+        continue;
+      }
+
+      String key = idx.name() + ":" + columnName;
+      segmentedZonemapCounts.merge(key, 1, Integer::sum);
+    }
+
+    return segmentedZonemapCounts.values().stream().anyMatch(count -> count > 1);
```
hamersaw: shouldForcePostScanFiltering() detection logic has no unit test: This method detects segmented zonemap indexes via complex logic (counting indexes by name+column, checking fragment presence). The method has multiple branches and conditions (empty filters, no columns, iteration logic, count comparison at line 473). No test exercises this detection—if the segmentation detection breaks, queries would silently use the wrong execution path.

**Fix:** Add a unit test that creates a dataset with a segmented zonemap index (multiple Index objects with same name+column), calls shouldForcePostScanFiltering() with a filter on that column, and asserts it returns true. Also test the negative case (monolithic zonemap or non-indexed column).

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -148,18 +203,70 @@ case class AddIndexExec(
...
-      fragmentIds: List[Integer]): IndexJob = {
-    // Get namespace info from catalog if available (for credential vending on workers)
-    val (nsImpl, nsProps, tableId, initialStorageOpts): (
-        Option[String],
-        Option[Map[String, String]],
-        Option[List[String]],
-        Option[Map[String, String]]) = catalog match {
+      column: String,
+      segments: Seq[Index]): Unit = {
+    val dataset = Utils.openDatasetBuilder(readOptions).build()
```
hamersaw, line 216: Is there a specific reason we are opening two datasets within this function?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -148,18 +203,70 @@ case class AddIndexExec(
...
+          column,
+          segments.toList.asJava).asScala.toList
+
+      if (existingIndices.nonEmpty) {
+        val committedUuids = committedSegments.map(_.uuid()).toSet
+        val refreshedDataset = Utils.openDatasetBuilder(readOptions).build()
+        try {
+          val removedIndices = refreshedDataset.getIndexes.asScala
+            .filter(index => index.name() == indexName && !committedUuids.contains(index.uuid()))
+            .toList
```
hamersaw: Cleanup can delete a concurrent writer's zonemap segments with the same name: `refreshedDataset.getIndexes` is filtered by `index.name() == indexName && !committedUuids.contains(index.uuid())`. If another writer concurrently committed additional segments with the same index name between our `commitExistingIndexSegments` call (line 221) and the `refreshedDataset` open (line 228), those segments will appear with UUIDs that are not in our `committedUuids` and will be removed by the `AddIndexOperation` at line 236-239, destroying the concurrent writer's work. The in-code comment at line 206-209 acknowledges a race but claims Lance OCC will surface a conflict — OCC only protects against version mismatch on commit, not against this read-modify-write removal pattern which uses `refreshedDataset.version()` as its readVersion and so will succeed without conflict.

**Fix:** Either capture the pre-commit set of index UUIDs from `existingIndices` and remove only those (`removedIndices = refreshed.indexes.filter(uuid in existingUuids)`), or push the replace semantics into a single atomic Lance operation so removal of old segments is part of the same transaction as the new commit.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -544,6 +657,115 @@ case class RangeBTreeIndexBuilder(
...
+      fragmentIds: List[Integer],
+      numSegments: Option[Int] = None): Seq[List[Integer]] = {
+    val targetTasks = numSegments match {
+      case Some(n) => math.max(1, math.min(fragmentIds.size, n))
+      case None => math.max(
+          1,
+          math.min(fragmentIds.size, addIndexExec.session.sparkContext.defaultParallelism))
+    }
+    val batchSize = math.ceil(fragmentIds.size.toDouble / targetTasks.toDouble).toInt
+    fragmentIds.grouped(batchSize).map(_.toList).toSeq
```
hamersaw: batchFragments produces fewer segments than requested num_segments: The `ceil(N/K)` batch sizing does not guarantee K batches. With `fragmentIds.grouped(batchSize)`, when N is not a clean multiple of K, you can end up with fewer than `targetTasks` groups. Examples: N=4, K=3 → batchSize=2 → 2 batches; N=8, K=5 → batchSize=2 → 4 batches; N=10, K=7 → batchSize=2 → 5 batches. This silently violates the documented `num_segments` contract (`docs/src/operations/ddl/create-index.md:42`) and the test assertion in BaseAddIndexTest.testCreateZonemapIndexWithNumSegments (line 289-292) only passes because `prepareDataset()` happens to create exactly 2 fragments; a 4-fragment table with `num_segments=3` would yield 2 segments and fail. Same risk applies to the default-parallelism path when `defaultParallelism` lies just above `fragmentCount/2`.

**Fix:** Distribute remainder evenly across the first `remainder` batches (e.g. compute `base = N / K`, `remainder = N % K`, then build K slices where the first `remainder` get `base+1` and the rest get `base`). Alternatively partition by index: `(0 until targetTasks).map(i => fragmentIds.slice(i * N / targetTasks, (i + 1) * N / targetTasks))`.

hamersaw (commented):
This looks pretty good. Thanks for the PR! A few things that we should tighten up IMO.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Good point — you're right that zonemap pruning already works via `ZonemapFragmentPruner` and scalar indexes should work independently. I've removed the entire `useScalarIndex` / `forcePostScanFiltering` mechanism in the latest push. The PR now has zero scan-side changes — only the index creation path is modified.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Removed — the entire `shouldForcePostScanFiltering` method and its Javadoc have been dropped in the latest push since we no longer need scan-side changes for segmented zonemaps.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
beinan: Removed — `shouldForcePostScanFiltering` has been dropped entirely. No scan-side changes in the latest push.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: The two opens are needed because `commitExistingIndexSegments` creates a new dataset version, so we need a fresh dataset handle to read the updated index state for the cleanup transaction. I've added a comment in the code explaining this. The first dataset is used for the segment commit, and the second reads the post-commit state to remove old segments.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Great catch. Fixed — the cleanup now captures pre-commit UUIDs (`existingUuids`) before calling `commitExistingIndexSegments`, and only removes indexes whose UUIDs were in that pre-commit set. This way concurrent writers' new segments (which have different UUIDs) are never touched.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Good catch — the `ceil(N/K)` approach indeed produces fewer batches than requested. Fixed by switching to index-based slicing: `(0 until k).map(i => fragmentIds.slice(i * n / k, (i + 1) * n / k))`. This guarantees exactly `k` batches (or fewer only when `n < k`).

beinan:
Thanks for the thorough review! All feedback has been addressed in the latest push (force-pushed as a single clean commit on latest main):

**Scan-side changes removed entirely:**
- Removed `useScalarIndex`, `forcePostScanFiltering`, and `shouldForcePostScanFiltering` — zonemap fragment pruning works via the existing `ZonemapFragmentPruner` path without needing special scan flags. No scan-side files are modified in this PR anymore.

**Index creation fixes:**
- **Race in `commitIndexSegments`**: Now captures pre-commit UUIDs and only removes indexes with those UUIDs, so concurrent writers' segments are never deleted.
- **`batchFragments` accuracy**: Switched from `ceil(N/K)` to index-based slicing (`slice(i*N/K, (i+1)*N/K)`) to guarantee the requested segment count.
- **`num_segments` validation**: Bounds check for `<= 0` and type validation, passed through constructor (no duplicate extraction).
- **Segment failure handling**: try/catch with clear error message about Lance GC cleanup.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
hamersaw: I am pretty concerned that this can leave us in a poor state if we commit the new index segments and there is a failure before removing the old ones. I think results should still be correct, but there will be quite a bit of overhead until the indexing process is reran.

It doesn't look like there is a solution for this in the lance core SDK when committing segmented indices, maybe we need to devise a solution for transactionally replacing segmented indices?

beinan: Great catch. After investigating Lance core's `commit_existing_index_segments` ([index.rs:1065-1164](https://github.com/lance-format/lance/blob/main/rust/lance/src/index.rs#L1065)), it turns out the core API **already handles atomic replacement** — it finds existing segments whose fragments overlap with incoming ones and removes them in the same `CreateIndex` transaction. The Spark-side manual cleanup (second dataset open + removal transaction) was redundant and is what introduced the race.

Fixed in the latest push by simplifying `commitIndexSegments` to just call `dataset.commitExistingIndexSegments()` and let Lance core handle the atomic add+remove. The method went from ~50 lines to ~10.

beinan:
@hamersaw All review feedback has been addressed — the key change since your last review is simplifying `commitIndexSegments` to rely on Lance core's built-in atomic replacement (single transaction for add+remove). Would you mind taking another look when you get a chance? Thanks!

hamersaw:
@LuciferYang do you mind making a pass here? Specifically, I'm interested with how this compares to your proposal (https://github.com/lance-format/lance-spark/pull/513) to support building distributed ZoneMap indexes.

LuciferYang:
> @LuciferYang do you mind making a pass here? Specifically, I'm interested with how this compares to your proposal (#513) to support building distributed ZoneMap indexes.

will give feedback later today.

LuciferYang:
@hamersaw @beinan Had a closer look. After the May 12 force-push on #516, the two PRs are adjacent rather than overlapping.

|                  | #516                                       | #513 distributed                                              | #513 consolidated (opt-in)             |
|------------------|--------------------------------------------|---------------------------------------------------------------|----------------------------------------|
| Tasks            | 1 per `num_segments` batch                 | 1 per fragment                                                | 1 per fragment (compute), driver write |
| Segments on disk | `num_segments` (default = `min(fragments, defaultParallelism)`) | = fragment count                                | 1                                      |
| Commit API       | `commitExistingIndexSegments`              | manual `AddIndexOperation` + `Transaction` + `CommitBuilder`  | same as distributed                    |
| Upstream blocker | none (project's already on lance-core `7.0.0-beta.10`) | none                                              | lance-format/lance#6779 + #6780, both unmerged |

The main difference between the two distributed paths is that `num_segments` on #516 is one knob doing two jobs: parallelism and segment count are the same lever. At `num_segments=1` you get one segment, but the work serialises onto a single executor — not the same as #513's consolidated path, which keeps compute parallel and only centralises the write. That decoupled corner only opens up once #6779/#6780 land.

For reference, sf=100 `store_sales` (234 fragments, ~288M rows) under #513: distributed = 15.0 s / 234 segments / 1.1 MB; consolidated = 28.1 s / 1 segment / 138 KB. The ~8× footprint drop is per-file header amortisation.

What I'd want to inherit from #516:
- It uses `commitExistingIndexSegments`. That's the right API; #513 reimplements the add+remove transaction by hand and should be rebased onto it.
- Dropping the scan-side `useScalarIndex` / `forcePostScanFiltering` machinery in the May 12 force-push was the right call — pruning works through `ZonemapFragmentPruner` without it.

Suggested path:
1. Land #516 as-is.
2. After it lands, I'll rebase #513 distributed onto `commitExistingIndexSegments`. The fragment-id exception wrapping in `ZonemapFragmentTask` is worth porting onto `ZonemapIndexTask` while we're there. After that, #513 only carries the consolidated path.
3. Consolidated lands as an opt-in (`spark.lance.zonemap.consolidate.enabled`) once #6779/#6780 release. Default-off — driver allocator holds every per-fragment Arrow batch until `writeZonemapIndexFromBatches` consumes them, so it can regress at very-high fragment counts.

One nit on #516, will leave inline: `targetTasks = math.min(fragmentIds.size, n)` silently clamps `num_segments=1000` to fragment count. The doc string reads like `num_segments` is a target, not an upper bound. Either log when clamping or reword the doc.

beinan:
@LuciferYang Thanks for the thorough comparison — the side-by-side table is really helpful.

You're right about `num_segments` doing double duty. The latest push addresses the nit:
- `num_segments` doc now clarifies it's an upper bound clamped to fragment count
- Added a log message when clamping occurs so it's not silent
- Switched `batchFragments` to index-based slicing to guarantee the requested segment count

Agreed on the suggested path — happy to land #516 as the distributed foundation, then have #513 rebase its distributed path onto `commitExistingIndexSegments` and carry the consolidated path as an opt-in once #6779/#6780 land. The fragment-id exception wrapping from `ZonemapFragmentTask` is a good addition to port over as well.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -85,10 +86,65 @@ case class AddIndexExec(
...
+
+    val numSegmentsOpt = args.find(_.name == "num_segments")
+    if (numSegmentsOpt.isDefined && !useLogicalSegmentCommit) {
+      throw new IllegalArgumentException(
+        "num_segments option is only supported for index types that use segmented builds (e.g., zonemap)")
+    }
+    val validatedNumSegments: Option[Int] = numSegmentsOpt.map { arg =>
+      val value =
+        try {
+          arg.value.asInstanceOf[Number].intValue()
```
LuciferYang: Scala's `null.asInstanceOf[Number]` returns `null` instead of throwing `ClassCastException`, so a `WITH (num_segments = null)` argument bypasses the friendly error path and dies with an opaque `NullPointerException` at `.intValue()`. 

On the other hand, if the parser delivers a `java.lang.Long` (e.g., for an out-of-range literal), `.intValue()` silently truncates rather than rejecting. Negative Longs below `Int.MinValue` truncate to *positive* Ints and slip past the `value <= 0` check on line 111. Validate the Long bounds before narrowing. 

It can be revised as follows:

```scala
val value = arg.value match {
  case null =>
    throw new IllegalArgumentException("num_segments must be a positive integer, got: null")
  case n: Number =>
    val asLong = n.longValue()
    if (asLong < 1L || asLong > Int.MaxValue) throw new IllegalArgumentException(
      s"num_segments must be a positive integer that fits in Int, got: $asLong")
    asLong.toInt
  case other =>
    throw new IllegalArgumentException(
      s"num_segments must be a positive integer, got: $other")
}
```

With this in place the redundant `if (value <= 0) throw …` block on line 111 can be removed.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -148,18 +204,32 @@ case class AddIndexExec(
...
-      fragmentIds: List[Integer]): IndexJob = {
-    // Get namespace info from catalog if available (for credential vending on workers)
-    val (nsImpl, nsProps, tableId, initialStorageOpts): (
-        Option[String],
-        Option[Map[String, String]],
-        Option[List[String]],
-        Option[Map[String, String]]) = catalog match {
+      column: String,
+      segments: Seq[Index]): Unit = {
+    val dataset = Utils.openDatasetBuilder(readOptions).build()
```
LuciferYang, line 216: The driver opens once on line 72 to enumerate `fragmentIds`, closes it, then `commitIndexSegments` opens a fresh one on line 214 to call `commitExistingIndexSegments`. Both go through the same `Utils.openDatasetBuilder(readOptions)`. Consider consolidating to a single driver-side open scoped to the entire zonemap branch — derive `fragmentIds` from it and pass that same handle into `commitIndexSegments`.

Caveat worth verifying first: if `commitExistingIndexSegments` requires a handle at the latest dataset version (not the version captured at line 72), reusing the older handle could fail the commit on a version mismatch. If the Lance core contract requires a fresh handle, leave `commitIndexSegments` as-is and only optimize the line 72 enumeration (e.g., enumerate via `lanceDataset` if it already exposes fragment IDs).

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -544,6 +620,125 @@ case class RangeBTreeIndexBuilder(
...
+            s"num_segments=$requested clamped to $clamped " +
+              s"(fragment count=$n)")
+        }
+        clamped
+      case None => math.max(
+          1,
+          math.min(n, addIndexExec.session.sparkContext.defaultParallelism))
+    }
+    (0 until k).map { i =>
+      fragmentIds.slice(i * n / k, (i + 1) * n / k)
```
LuciferYang: `i * n` (both `Int`) overflows once the product exceeds `Int.MaxValue ≈ 2.1×10^9`. Triggering requires a deliberately large `num_segments` *and* a fragment count where `i*n` crosses the boundary — e.g., 200k fragments with `num_segments` near `n` puts `i*n` near `4×10^10`. Not a hot-path concern, but cheap to make overflow-safe. Promote one operand to `Long`:

```scala
fragmentIds.slice((i.toLong * n / k).toInt, ((i.toLong + 1) * n / k).toInt)
```

Equivalent and overflow-safe.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -148,18 +204,32 @@ case class AddIndexExec(
       UTF8String.fromString(indexName))))
   }
 
-  private def createIndexJob(
-      dataset: Dataset,
-      lanceDataset: LanceDataset,
+  // Lance core's commitExistingIndexSegments handles atomic replacement:
```
LuciferYang, line 209: nit: Drop a one-line comment at the call site (line 132) — e.g., `// atomic add+remove via Lance core; see commitIndexSegments` — so the replacement semantics are visible without jumping definitions.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -85,10 +86,65 @@ case class AddIndexExec(
...
+        }
+      if (value <= 0) {
+        throw new IllegalArgumentException(
+          s"num_segments must be a positive integer, got: $value")
+      }
+      value
+    }
+
+    // Zonemap uses logical segment commit path
+    if (useLogicalSegmentCommit) {
```
LuciferYang, line 119: nit: Cosmetic micro-allocation, but it signals incorrectly that the UUID is needed by every branch. Move `val uuid = UUID.randomUUID()` below the `if (useLogicalSegmentCommit) { … return … }` block so it's only generated for the merge-metadata branch.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -44,9 +45,9 @@ import scala.collection.JavaConverters._
  * Physical execution of distributed CREATE INDEX (ALTER TABLE ... CREATE INDEX ...) for Lance datasets.
  *
  * <ul>
- * <li>For BTREE index, it uses a range-based approach that redistributes and sorts data across partitions, creates indexes for each range in parallel, and finally merges them into a global index structure.
- * <li>For other index types, it processes each fragment independently in parallel, merges index metadata
- * and commits an index-creation transaction.
+ * <li>For BTREE index, range mode redistributes and sorts data across partitions, creates indexes for each range in parallel, and finally merges them into a global index structure.
```
LuciferYang: The class doc lists "BTREE range → zonemap → BTREE-fragment/FTS", but the code dispatches "zonemap → BTREE (range vs fragment) → FTS". Reorder bullets to mirror the dispatch flow so the doc reads top-down.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -544,6 +620,125 @@ case class RangeBTreeIndexBuilder(
...
+        throw new RuntimeException(
+          "Zonemap segment build failed. Uncommitted segments (if any) " +
+            "will be cleaned up by Lance's garbage collection.",
+          e)
+    }
+  }
+
+  private def batchFragments(
+      fragmentIds: List[Integer],
+      numSegments: Option[Int] = None): Seq[List[Integer]] = {
```
LuciferYang: `batchFragments` is private and called from a single site that always passes the argument. Drop `= None` to avoid signaling an extension point that doesn't exist.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -544,6 +620,125 @@ case class RangeBTreeIndexBuilder(
...
+        .parallelize(tasks, tasks.size)
+        .map(t => t.execute())
+        .collect()
+        .map(decode[Index])
+        .toSeq
+    } catch {
+      case e: Exception =>
+        throw new RuntimeException(
+          "Zonemap segment build failed. Uncommitted segments (if any) " +
+            "will be cleaned up by Lance's garbage collection.",
```
LuciferYang: Will it really be cleaned up automatically?

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
```diff
@@ -198,6 +199,151 @@ public void testCreateBTreeIndexWithZoneSize() {
     Assertions.assertEquals("text_15", r.getString(1));
   }
 
+  @Test
+  public void testCreateZonemapIndex() {
```
LuciferYang, line 203: nit: no negative test cases, such as
- negative test for multi-column zonemap
- negative test for `num_segments` on `btree`/`fts`
- test for `num_segments = 0` / negative values

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
```diff
@@ -198,6 +199,151 @@ public void testCreateBTreeIndexWithZoneSize() {
...
+          fragmentCount,
+          coveredFragments,
+          "Expected committed segments to cover all fragments exactly once");
+    } finally {
+      lanceDataset.close();
+    }
+  }
+
+  @Test
+  public void testRepeatedCreateZonemapIndexReplacesExistingSegments() {
```
LuciferYang, line 309: The test runs the SQL twice and asserts segment count stays at `expectedSegmentCount`. That catches the duplication failure mode (second run *adds* instead of replacing) but not the no-op failure mode (second run silently does nothing). Capture the segment UUIDs (or `createdAt`) after the first run and assert they differ after the second. This assumes Lance's `createIndex` mints fresh UUIDs per call — if UUIDs are content-addressed or otherwise stable across rebuilds, fall back to comparing `createdAt`.

docs/src/operations/ddl/create-index.md
```diff
@@ -24,13 +24,23 @@ The following index methods are supported:
...
 
 The `CREATE INDEX` command supports options via the `WITH` clause to control index creation. These options are specific to the chosen index method.
 
+### ZoneMap Options
+
+For the `zonemap` method, the following options are supported:
+
+| Option          | Type | Description                                  |
+|-----------------|------|----------------------------------------------|
+| `rows_per_zone` | Long | The approximate number of rows per zonemap zone. |
```
LuciferYang: If both are passed through `IndexUtils.toJson` the same way, label them the same way. `zone_size` in the btree section below should match too for cross-section consistency.

LuciferYang (commented):
Does the PR description also need to be updated?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Fixed — switched to pattern match handling null, Long bounds, and non-Number types explicitly. Removed the redundant `<= 0` check.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Good point. Left as-is for now since `commitExistingIndexSegments` may require a handle at the latest version. Worth consolidating in a follow-up if we confirm the version contract.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Fixed — promoted to `i.toLong * n / k` to avoid overflow.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Added.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Done — moved `val uuid` below the zonemap early return.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Reordered to match dispatch flow: zonemap → BTREE (range vs fragment) → FTS.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Dropped the `= None` default.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
beinan: Good question — updated the message. Uncommitted segments are not visible to readers and do not affect query correctness. They are orphaned artifacts that occupy storage but have no semantic impact.

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
beinan: Added negative tests for: multi-column zonemap, `num_segments` on btree, and zero/negative `num_segments`.

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
beinan: Fixed — now captures segment UUIDs after the first run and asserts they differ after the second.

docs/src/operations/ddl/create-index.md
beinan: Fixed table alignment for consistency.

LuciferYang (approved):
The "What Changed" section still references `LanceScanBuilder.java`, `LanceScan.java`, `LanceInputPartition.java`, `LanceFragmentScanner.java`, and `LanceCountStarPartitionReader.java`, plus the bullet *"Add segmented zonemap scan support with Spark-side post-scan filtering fallback"* in Summary. None of this is in the current diff. 

others LGTM

beinan:
> The "What Changed" section still references `LanceScanBuilder.java`, `LanceScan.java`, `LanceInputPartition.java`, `LanceFragmentScanner.java`, and `LanceCountStarPartitionReader.java`, plus the bullet _"Add segmented zonemap scan support with Spark-side post-scan filtering fallback"_ in Summary. None of this is in the current diff.
> 
> others LGTM

Sorry for my delay, just updated.  can we merge this pr?  @LuciferYang @hamersaw

hamersaw (approved):
Thanks for the iterations on this!

PR #519, opened 2026-05-12
feat: use Lance MemWAL sharding for bucketed writes and SPJ

beinan:
## Summary

Use Lance MemWAL sharding specs as the connector metadata for Spark bucket and identity transforms instead of custom table-property metadata.

- Initializes MemWAL sharding through the Lance Java API for create and staged-create flows.
- Routes writes with `org.lance.memwal.ShardingEvaluator`, using Lance schema resolution for sharding fields.
- Reports Spark `KeyGroupedPartitioning` from Lance sharding metadata and fragment sharding keys for SPJ.
- Removes the legacy `lance.partition.*` table-property path and the primary-key-based sharding restriction.
- Bumps the Lance Java dependency to `7.0.0-rc.1`.


hamersaw:
## Reconsidering `lance.partition.bucket.columns` before it hardens

Introducing `lance.partition.bucket.columns` + `.num_buckets` as a sibling namespace to `lance.partition.columns` bifurcates a concept that Iceberg, Spark's V2 API, and `PartitionTransformUtil` itself all treat as unified. I'd like to suggest folding both into a single partition spec before the bucket-specific properties harden into a stable API.

### The bifurcation today

| | Identity path | Bucket path |
|---|---|---|
| Table property | `lance.partition.columns` | `lance.partition.bucket.columns` + `.num_buckets` |
| Scan-side info type | `PartitionInfo` | `BucketPartitionInfo` |
| Writer branch | `partitionColumnIndices > 0` | `bucketColumnIndices > 0` |
| `SparkWrite` distribution | `clustered(partCol)` | `clustered(bucketCol)` |

Every consumer (`LanceScan`, `LanceScanBuilder`, `LanceDataWriter`, `SparkWrite`, `PartitionTransformUtil`) carries an `if bucket … else if partition …` ladder. `SparkWrite.toBatch()` manually enforces a mutual-exclusion invariant that nothing in the type system prevents. Spark itself unifies via `Transform[]` — the bifurcation is introduced at our storage layer, not the API boundary.

### Proposed format

A single `lance.partition.spec` property, JSON-encoded list of fields:

```json
[
  {"transform": "bucket",   "col": "customer_id", "num_buckets": 16},
  {"transform": "year",     "col": "event_ts"},
  {"transform": "identity", "col": "region"}
]
```

Parsed once via Jackson into a typed sealed interface:

```java
sealed interface PartitionTransform
    permits IdentityTransform, BucketTransform, YearTransform, ... {
  String column();
  String canonicalName();                                  // Spark SPJ matching
  Object applyToRow(InternalRow row, int colIdx, DataType type);
  org.apache.spark.sql.connector.expressions.Expression toSparkExpression();
  Optional<Object> fragmentKeyFromZones(List<ZoneStats> zones);
}
```

Consumers take `List<PartitionTransform>` and dispatch through the interface.

### What we gain

1. **No more parallel branches.** `LanceDataWriter.write()` becomes one path that asks the spec for the partition tuple and rolls when it changes. `SparkWrite.requiredDistribution/Ordering` map directly over the spec. Scan-side `PartitionInfo`/`BucketPartitionInfo` collapse into one type.

2. **The existing TODO in `SparkWrite.java:152` resolves for free.** Sort key becomes `t.toSparkExpression()` per field, so the bucket path sorts by `bucket(N, col)` instead of raw column — distinct values hashing to the same bucket arrive contiguous, eliminating extra fragment rolls.

3. **New transforms become subclasses, not new namespaces.** Adding `year(ts)`, `month(ts)`, `day(ts)`, `truncate(W, col)` is one class each. Under the current design each needs a new property family, info class, writer branch, and validation rule.

4. **Per-transform fragment detection.** Strict `min == max` for identity/bucket; range-tolerant for year/day/month (a fragment spanning Mar 1–31 is fully partition-compatible for `month(ts)` even when `min != max`). Today's shared `computeFragmentPartitionValues` can't express this.

5. **`canonicalName` becomes a first-class, reviewable field.** Cross-source SPJ silently fails when canonical names diverge — `LanceBucketFunction.canonicalName()` is currently set to `iceberg.bucket(<type>)` precisely to match Iceberg's, but that's a buried string literal. As a field on each transform subclass it's explicit and a natural code-review checkpoint when adding the next variant.

6. **Better matches the interop landscape.** Identity is the universal floor — it's the path that gives Lance↔Delta and Lance↔Hudi SPJ today (neither exposes V2 bucket transforms). Bucket is effectively a Lance↔Iceberg bilateral. Date transforms (year/month/day) are the more common Iceberg interop case in analytical workloads and are the next gap to fill. A unified spec puts identity, bucket, and dates on equal footing — none privileged via the storage schema.

7. **Compatible with planned zone-granularity SPJ.** When `LanceInputPartition` moves from fragment-level to zone-row-range granularity, the only `PartitionSpec` change is `fragmentKeyFromZones(List<ZoneStats>)` → `zoneKey(ZoneStats)` — a one-line interface edit per transform. The writer's fragment-rolling becomes optional rather than load-bearing, and `requiredDistribution()` downgrades to a layout hint. The unified abstraction localizes that future change to the transform interface; the bifurcated design forces parallel updates to both code paths, with each able to regress the other.

### Migration

`LanceScanBuilder` reads both formats during a deprecation window, preferring the new property; CREATE/ALTER writes only the new one. No on-disk data moves. Drop the legacy reader after a release.

### Recommendation

Unify now, while bucket is brand new and only two table-property families exist. The cost of doing this after the third transform lands — or after users have created tables against `lance.partition.bucket.*` — is meaningfully higher than doing it before merging this PR.

beinan:
@hamersaw Great feedback — fully agreed on unifying before this hardens. I've implemented your suggestion in the latest commit (`bcb629c`):

**What changed:**
- New `PartitionTransform` class (`partition/PartitionTransform.java`) with `Identity` and `Bucket` subclasses — each transform owns its partition key computation, zonemap detection, and Spark expression generation
- Single `lance.partition.spec` JSON property replaces both `lance.partition.columns` and `lance.partition.bucket.*`:
  ```json
  [{"transform": "bucket", "col": "region", "num_buckets": 4}]
  ```
- All consumers (`LanceScan`, `LanceScanBuilder`, `SparkWrite`, `LanceBatchWrite`, `LanceDataWriter`) now take `List<PartitionTransform>` — no more parallel if/else ladders
- `PartitionTransformUtil.parseSpec()` reads both the new format and legacy properties for backward compat
- Adding future transforms (year, month, truncate) is now one class each

**What's still tracked as follow-up:**
- Sort-by-bucket-ID optimization (the TODO in `Bucket.toSortOrder()`) — currently sorts by raw column value, so fragments may split unnecessarily when different values hash to the same bucket

All 53 tests pass including the SPJ shuffle-elimination assertion.

jackye1995:
Updated this PR with commit 1188318. Spark no longer builds or passes a custom source-id-to-column map on the write path. Existing tables now pass dataset.getLanceSchema() into Lance's ShardingEvaluator so Lance derives source field IDs from the table schema. The staged-create fallback still works by embedding the source column in the temporary sharding field parameters before dataset metadata exists.\n\nThis removes SparkLanceShardingAdapter.sourceIdToColumnMap and keeps the remaining adapter code focused on Spark/catalog/read-side conversion. The change depends on lance-format/lance#6854, specifically commit e35a92a5a, which adds the schema-derived Rust/Java/Python sharding evaluator API. Once that Lance PR is merged and released, this Spark PR should bump to the published Lance artifact containing that API.

jackye1995:
Updated this PR with commit `632f466` to use the published Lance artifact `org.lance:lance-core:7.0.0-rc.1`, which includes the schema-derived `ShardingEvaluator` API from lance-format/lance#6854.

Validation:
- `./mvnw -U dependency:get -Dartifact=org.lance:lance-core:7.0.0-rc.1 -Dtransitive=false`
- `./mvnw -U test -pl lance-spark-3.5_2.12 -am -Dtest=BucketWriteTest,PartitionedWriteTest,SparkLanceShardingAdapterTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw -U compile -pl lance-spark-4.1_2.13 -am`
- `./mvnw spotless:check -pl lance-spark-3.5_2.12 -am -DskipTests`
- `./mvnw -U dependency:tree -pl lance-spark-3.5_2.12 -Dincludes=org.lance:lance-core` resolved `org.lance:lance-core:jar:7.0.0-rc.1`

jackye1995 (approved):

PR #521, opened 2026-05-12
fix: backtick-quote multipart identifiers in range-based BTree index job

summaryzb:
A path-based identifier is already  unquoted in RangeBasedBTreeIndexJob, when feeds it to SparkSession.table(String), result in exception

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/ParserUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/ParserUtils.java
@@ -34,4 +34,26 @@ public final class ParserUtils {
     }
     return text;
   }
+
+  /**
+   * Backtick-quotes an identifier for use in a Spark multipart identifier string. Idempotent: if
+   * the input is already surrounded by backticks it is returned unchanged. Throws when the input
+   * has an unbalanced leading or trailing backtick.
+   *
+   * @param identifier a bare or already-quoted identifier
+   * @return the backtick-quoted identifier
+   * @throws IllegalArgumentException if the identifier has a backtick on only one side
+   */
+  public static String quoteIdentifier(String identifier) {
+    boolean startsWithBacktick = identifier.startsWith("`");
+    boolean endsWithBacktick = identifier.endsWith("`");
+    if (identifier.length() >= 2 && startsWithBacktick && endsWithBacktick) {
+      return identifier;
+    }
+    if (startsWithBacktick || endsWithBacktick) {
+      throw new IllegalArgumentException(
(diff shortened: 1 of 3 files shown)


LuciferYang:
cc @fangbo

fangbo:
@summaryzb Thank you for fixing it. Could you please provide some test cases for `when feeds it to SparkSession.table(String), result in exception`

summaryzb:
Hi @fangbo — thanks for the review. Below is  on top of  main branch  (i.e., before this fix) so the failure mode is unambiguous, plus a sketch of where I'd add the test.

  ### Why `SparkSession.table(String)` blows up

  The outer `ALTER TABLE` parses fine — the path is already backtick-quoted by the caller:

  ```sql
  ALTER TABLE lance_default.`/abs/path/to/date_dim.lance`
      CREATE INDEX idx_d_year USING btree (d_year)
      WITH (build_mode = 'range');
  ```

  The crash only happens inside `RangeBasedBTreeIndexJob.prepare()`, where the identifier is rebuilt without quoting and fed back into `session.table(...)`:

  ```scala
  val parts = if (namespace.isEmpty) Seq(catalog.name(), ident.name())
              else catalog.name() +: namespace :+ ident.name()
  val fullTableName = parts.mkString(".")          // <-- buggy
  val df = session.table(fullTableName)
  ```

  For a path-based catalog the second part contains `/`, `.`, etc. Spark's multipart identifier parser splits at the first `.` after `lance_default` and then rejects the `/`:

  ```
  [PARSE_SYNTAX_ERROR] Syntax error at or near '/'.(line 1, pos 14)

  == SQL ==
  lance_default./abs/path/to/date_dim.lance
  --------------^^^
  ```

  The same shape blows up for namespace-backed catalogs

  ### End-to-end repro using the existing benchmark module

  ```bash
  git revert --no-edit <this-fix-commit>          # roll the fix back
  make bundle SPARK_VERSION=3.5 SCALA_VERSION=2.12

  # 1) generate data so the dataset path contains `/`
  ./benchmark/scripts/generate-data.sh 1 lance "local[*]"

  # 2) switch one SQL template to range mode
  #    benchmark/src/main/resources/index/date_dim.sql
  #    e.g.  ... USING btree (d_year) WITH (build_mode = 'range');

  # 3) build & run
  rm -rf benchmark/target
  ./benchmark/scripts/create-indexes.sh "local[*]"
  ```

  Result: all five `date_dim` range-mode statements fail with the `PARSE_SYNTAX_ERROR` above; the remaining fragment-mode statements succeed (they don't take this code path).

fangbo:
+1, Thanks for fixing it.

PR #528, opened 2026-05-13
fix: advance _row_last_updated_at_version for UPDATE COLUMNS FROM

### LINKED ISSUE #418: UPDATE COLUMNS FROM: bump _row_last_updated_at_version for matched rows
## Summary

UPDATE COLUMNS FROM commits a new dataset version but does not advance
_row_last_updated_at_version on rows whose column values were rewritten.

ALTER TABLE ... UPDATE COLUMNS ... FROM (UpdateColumnsBackfill in lance-spark)
commits through Lance CommitBuilder and Transaction as a normal Update with
UpdateMode.RewriteColumns. The table version increases, but per-row change-data
metadata _row_last_updated_at_version can stay the same as before the commit
(for example still equal to _row_created_at_version), even though data in the
updated columns changed.

## Expected behavior

From Lance row lineage and change-data feed (CDF) docs, _row_last_updated_at_version
is the dataset version at which the row was last modified. If a write creates a
new dataset version and changes visible row data for matched rows, those rows
should get _row_last_updated_at_version set to that new version.
_row_created_at_version should stay at the version where the row first appeared.

## Actual behavior

After UPDATE COLUMNS FROM, rows that had columns rewritten can still show the
same _row_last_updated_at_version as before, while the dataset version has moved
forward on commit.

## Reproduction

1. Create a Lance table with stable row IDs enabled (enable_stable_row_ids).
2. Insert several rows (e.g. id 1, 2, 3) so CDF columns exist; note dataset version V0 and
   _row_created_at_version / _row_last_updated_at_version for each row.
3. Run ALTER TABLE ... UPDATE COLUMNS ... FROM with a source view that updates only one row
   (e.g. id = 2); leave id 1 and 3 unchanged in the source.
4. Read _row_last_updated_at_version for id = 2: it may still equal the pre-update value (or
   match created_at only) even though the dataset version advanced past V0.
5. id 1 and 3 should not incorrectly bump.

## Note

This ticket is on top of the following:

  - lance-format/lance#6464 — JNI loses version metadata; incorrect stable row ID
    lookup in Operation::Update (see fix PRs such as lance-format/lance#6465).

  - lance-format/lance-spark#406 — Spark connector cannot preserve stable row IDs
    across standard SQL UPDATE (native DeltaWriter.update path on Spark 3.5+, etc.).

jerryjch:
## Lance dependency — required to compile

* Depends on lance-format/lance#6748, which adds
  `Update.Builder.updatedFragmentOffsets(...)` / `Update.updatedFragmentOffsets()` and JNI
  FromJava + IntoJava so the driver commit passes matched offsets into Rust.

Bump **`lance.version`** in `pom.xml` to a release that includes
lance#6748 before building or merging this PR; otherwise build fails on the new API calls.

## Summary

* Fixes https://github.com/lance-format/lance-spark/issues/418.
* `UpdateColumnsWriter.processFragment`: after each `fragment.updateColumns()` call, reads
  `result.getUpdatedRowOffsets()` and accumulates a `Map<Long, long[]>` of fragment id →
  matched physical row offsets.
* `TaskCommit`: carries the per-fragment offset map alongside the existing
  `updatedFragments` and `fieldsModified` fields.
* `UpdateColumnsBackfillBatchWrite.commit()`: merges offset maps from all task commits and
  passes them to `Update.builder().updatedFragmentOffsets(...)`. Lance's `build_manifest`
  then calls the partial `_row_last_updated_at_version` refresh only for the matched rows,
  leaving unmatched rows and untouched fragments unchanged.
* `BaseUpdateColumnsBackfillTest`: flips
  `testUpdateColumnsPreservesCreatedAtAndAdvancesLastUpdatedWithStableRowIds` from the
  "known gap" pin (assertEquals, no change) to the correct assertion
  (`assertTrue(after > before)`); updates Javadoc accordingly.

## Background

`UPDATE COLUMNS FROM` rewrites column data in place via Lance's `Operation::Update` with
`RewriteColumns` mode. Lance's `build_manifest` can partially refresh
`_row_last_updated_at_version` for only the matched rows — but only when the
`updated_fragment_offsets` map is non-empty on the commit. Previously
`UpdateColumnsBackfillBatchWrite` never populated this map, so the partial refresh never
activated and `_row_last_updated_at_version` stayed stale after every UPDATE COLUMNS
commit, breaking CDF consumers.

The matched row offsets are already computed inside Lance during `fragment.updateColumns()`
and surfaced via `FragmentUpdateResult.getUpdatedRowOffsets()` (lance#6650). The missing
piece was wiring those offsets from the executor task result through `TaskCommit` to the
driver commit, and then setting them on the `Update` operation — which this PR does.

## Test plan

* `BaseUpdateColumnsBackfillTest#testUpdateColumnsPreservesCreatedAtAndAdvancesLastUpdatedWithStableRowIds`
  — creates a stable-row-id table, runs UPDATE COLUMNS over all rows, and asserts
  `_row_last_updated_at_version` strictly increases for each row while
  `_row_created_at_version` is unchanged.


jerryjch:
cc @hamersaw

jerryjch:
Hi, @hamersaw All the dependency PR's from lance have been merged. Can you review this PR again?

hamersaw (approved):
Looks great, thanks for your patience on this! Will merge once CI passes.

PR #532, opened 2026-05-14
fix: use describeIndices to list indices to improve performance

fangbo:
Currently `Dataset.getIndexes` will return all indexed fragments for each index. If the fragment number is huge, this method becomes slow and consume too much memory.

`Dataset.describeIndices` can improve performance and avoid too much memory consumption.


fangbo:
@LuciferYang This PR is ready. Could you please look at it ? Thank you.

LuciferYang (approved):
+1, LGTM

PR #533, opened 2026-05-14
fix: preserve Arrow field metadata for nested List/FixedSizeList/Map children

### LINKED ISSUE #515: fromArrowField drops Arrow field metadata for List/FixedSizeList elements and Map key/value types
## Summary

`LanceArrowUtils.fromArrowField` returns a bare `DataType`, so any Arrow field metadata attached to nested elements is lost when the parent is a `List`, `FixedSizeList`, or `Map`. Spark's `ArrayType` and `MapType` have no slot for per-element metadata either — `(elementType, containsNull)` and `(keyType, valueType, valueContainsNull)` — so the gap is structural, not just a missing branch.

## Impact

Round-trip fidelity is broken for these types when they appear as List/FixedSizeList element or Map key/value (top-level and `Struct`-nested cases work for `Date(MILLISECOND)`):

- `Date(MILLISECOND)` — `array<date(ms)>` round-trips as `array<date(day)>`.
- `LargeUtf8` — loses the `ARROW_LARGE_VAR_CHAR_KEY` marker.
- `FixedSizeBinary` / `FixedSizeList` — loses size metadata.
- `Float16` — loses the `ARROW_FLOAT16_KEY` marker.

This is the follow-up tracked by the TODO comments in #464 (now removed).

## Suggested fix

Adopt a namespacing convention on the parent `StructField` metadata — e.g. `_lance.element.date_millisecond = true` for `array<date(ms)>`, with paths like `_lance.value.element.date_millisecond` for nested cases such as `map<string, array<date(ms)>>`.

- `fromArrowSchema` walks the Arrow `Field` tree per top-level field and emits namespaced keys for any inner `Date(MILLISECOND)`, `LargeUtf8`, `FixedSizeBinary`, `FixedSizeList`, `Float16` under a List/FixedSizeList/Map.
- `toArrowField` reads those namespaced keys in the `ArrayType` and `MapType` branches and applies them to the element/key/value `FieldType`.

While there, also extend the `Struct` branch in `fromArrowField` (currently only handles `Date(MILLISECOND)`) to cover the other metadata-carrying types for symmetry.

## Repro hint

A test analogous to the one added in #464 for top-level `Date(MILLISECOND)` columns, but wrapped in `array<date>` or `map<string, date>`, should fail to preserve the millisecond unit through a Spark roundtrip.

LuciferYang:
## Summary

Fixes #515. `LanceArrowUtils.fromArrowField` returned a bare `DataType`, so Spark's `ArrayType` and `MapType` had nowhere to carry the special-type markers (`Date(MILLISECOND)`, `LargeUtf8`, `Float16`, `FixedSizeList` size) attached to nested element / key / value fields. On writeback those types silently degraded:

- `array<date(ms)>` → `array<date(day)>`
- `array<largeutf8>` → `array<utf8>`
- `map<largeutf8, date(ms)>` → `map<utf8, date(day)>`
- `array<fsl<float16, N>>` → `array<list<float32>>`

The Struct branch in `fromArrowField` previously only carried over `Date(MILLISECOND)` — `LargeUtf8`, `Float16`, and `FixedSizeList` markers were lost on struct children too.

## Approach

The bug is structural: Spark's `ArrayType(elementType, containsNull)` and `MapType(keyType, valueType, valueContainsNull)` have no per-element metadata slot, so child Arrow-field metadata cannot ride on the Spark `DataType` itself.

Per the suggestion in the issue, this PR adopts a namespacing convention: when walking into a `List` / `FixedSizeList` / `Map`, the child's Spark `Metadata` is serialized as JSON under three namespaced keys on the parent's metadata:

| Key | Carries |
|---|---|
| `_lance.element` | Child metadata for List / FixedSizeList |
| `_lance.key` | Map key metadata |
| `_lance.value` | Map value metadata |

The same scheme nests recursively, so `map<*, array<date(ms)>>` and `array<array<date(ms)>>` both roundtrip cleanly.

Implementation:

- New private `buildFieldMetadata(field)` helper computes the full Spark metadata for any Arrow Field. Both `fromArrowSchema` and the Struct branch in `fromArrowField` now delegate to it, so struct children pick up the same set of markers as top-level fields.
- `toArrowField` parses the embedded `_lance.*` JSON, threads it through the recursive call for the element / key / value type, and strips it from the Arrow user-metadata map so it doesn't leak downstream.
- Float16 vectors are unchanged on the read side (the marker lives on the parent FixedSizeList, not the element), but the new logic ensures the marker survives when an FSL is nested inside another List or Map.

## Tests

`LanceArrowUtilsSuite` gains 9 new tests covering each affected shape:

- `Date(MILLISECOND)` inside `Array`, inside `FixedSizeList`, as `Map` value, nested `Array<Array<...>>`, inside Struct.
- `LargeUtf8` inside `Array`, inside `FixedSizeList`, as `Map` key, inside Struct.
- `FixedSizeList<Float16>` nested inside an `Array` (read-side always, writeback gated on `Float16Utils.isFloat2VectorAvailable` so Arrow 15 builds skip it).

The pre-existing `BaseDateMilliRoundtripTest.testDateMilliInArrayDegradesToDay` was pinning the buggy behavior — renamed to `testDateMilliInArrayPreservedOnRoundtrip` and flipped to assert the correct outcome.

## Test plan

- [x] `mvn test -pl lance-spark-base_2.12` (LanceArrowUtilsSuite: 24/24 pass)
- [x] `mvn test -pl lance-spark-3.5_2.12` (688 tests, 0 failures, 46 skipped — same as main)
- [x] `mvn test -pl lance-spark-4.0_2.13` (704 tests, 0 failures, 38 skipped — exercises Arrow 18 + Float2Vector paths)
- [x] `mvn checkstyle:check spotless:check -pl lance-spark-base_2.12`

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
@@ -47,6 +47,19 @@ object LanceArrowUtils {
   val ARROW_LARGE_VAR_CHAR_KEY = LargeVarCharUtils.ARROW_LARGE_VAR_CHAR_KEY
   val ARROW_DATE_MILLISECOND_KEY = DateMilliUtils.ARROW_DATE_MILLISECOND_KEY
 
+  // Namespaced keys used to embed child Spark Metadata on a parent StructField when the
+  // child sits inside an ArrayType/MapType — Spark has no per-element metadata slot of its
+  // own. The value is the child Metadata serialized as JSON (the same shape produced by
+  // `Metadata#json`), so deeply nested cases work transparently.
+  val LANCE_ELEMENT_METADATA_KEY = "_lance.element"
+  val LANCE_MAP_KEY_METADATA_KEY = "_lance.key"
+  val LANCE_MAP_VALUE_METADATA_KEY = "_lance.value"
+
+  private val LANCE_INTERNAL_METADATA_KEYS = Set(
+    LANCE_ELEMENT_METADATA_KEY,
+    LANCE_MAP_KEY_METADATA_KEY,
+    LANCE_MAP_VALUE_METADATA_KEY)
+
   def fromArrowField(field: Field): DataType = {
     field.getType match {
       // Handle unsigned integers by mapping to larger signed types
@@ -69,7 +82,7 @@ object LanceArrowUtils {
         val elementType = fromArrowField(elementField)
         val containsNull = elementField.isNullable
         ArrayType(elementType, containsNull)
-      case struct: ArrowType.Struct =>
+      case _: ArrowType.Struct =>
         // Always recurse through LanceArrowUtils for struct children so special cases
         // like Date(MILLISECOND), FixedSizeBinary, etc. are applied in nested schemas too.
         val blobField = isBlobField(field)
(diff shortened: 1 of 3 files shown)


LuciferYang:
cc @fangbo

fangbo:
+1,  Great Optimization !

LuciferYang:
Thank you @fangbo

PR #546, opened 2026-05-20
feat: round-trip UDT columns automatically via __udt field metadata

LuciferYang:
## Summary

#471 made VectorUDT (and any UDT) writable into Lance. On read the columns came back as a plain struct, so callers had to reconstruct the UDT themselves. This PR closes the round-trip.

The write side stamps the UDT's class FQN onto the Arrow field's user metadata under the `__udt` key. The read side consults the marker and rewraps the field's dataType back to the UDT. No configuration, no per-call opt-in — symmetric with the write path.

## How it works

All in `LanceArrowUtils.scala`:

- `toArrowField` UDT branch: unwrap to `sqlType` for storage. When `sqlType` is a `StructType` (the shape used by every MLlib UDT), stamp `__udt = udt.getClass.getName` on the FieldType metadata; other shapes write through without the marker.
- `fromArrowField`: if a field's metadata carries `__udt`, load the named class, verify `udt.sqlType` matches the persisted shape, return the UDT.
- `deduplicateFieldNames` preserves `UserDefinedType` instances instead of recursively unwrapping. Without this the UDT branch wouldn't fire at every nesting depth (inside a StructType, as an ArrayType element, etc.).
- `buildFieldMetadata` strips `__udt` from the read-back Spark `StructField.metadata` so the marker doesn't surface as noise once the dataType is already the UDT.
- `loadUdt` uses `Class.forName(fqn, initialize=false, …)` so a crafted manifest can't trigger an arbitrary class's static initializer before we've verified it's actually a `UserDefinedType`.

## Tests

`BaseSparkDataTypeRoundtripTest`:

- `testVectorUDTRoundtrip` — VectorUDT (Dense + Sparse). Asserts the read-back schema carries VectorUDT and that `__udt` is stripped from user-visible metadata.
- `testMatrixUDTRoundtrip` — MatrixUDT (different sqlType, different user class) for generality.
- `testNestedUDTInStructRoundtrip` — UDT inside a StructType.
- `testUDTAsArrayElementRoundtrip` — UDT as an ArrayType element.
- `testVectorUDTReWriteRoundtrip` — read-write-read, idempotency check. Guards against subtle metadata duplication or loss.

## Test plan

- [x] CI passes across Spark 3.4 / 3.5 / 4.0 / 4.1 × Scala 2.12 / 2.13
- [x] VectorUDT round-trips with no configuration
- [x] MatrixUDT round-trips
- [x] Nested-UDT (struct, array) round-trips
- [x] Read-write-read idempotency holds

### DIFF
--- a/docs/src/operations/dql/select.md
+++ b/docs/src/operations/dql/select.md
@@ -244,6 +244,10 @@ These options control how data is read from Lance datasets. They can be set usin
 | `pushDownFilters`     | Boolean | `true`  | Whether to push down filter predicates to the Lance reader for optimized scanning.                                |
 | `topN_push_down`      | Boolean | `true`  | Whether to push down TopN (ORDER BY ... LIMIT) operations to Lance for optimized sorting.                         |
 
+## User-Defined Types
+
+Spark UDT columns (e.g. MLlib's `VectorUDT`) round-trip transparently. On write the column is unwrapped to the UDT's `sqlType` (a struct) for storage, and the UDT's fully qualified class name is stamped onto the Arrow field's user metadata under the `__udt` key. On read, that marker is consulted and the schema is rewrapped to the original UDT automatically. No configuration needed; the user-facing class (e.g. `Vector` for `VectorUDT`) is materialized in rows as usual.
+
 ### Example: Reading with Options
 
 === "SQL"
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
@@ -13,7 +13,10 @@
  */
 package org.lance.spark;
 
+import org.apache.spark.ml.linalg.DenseMatrix;
 import org.apache.spark.ml.linalg.DenseVector;
+import org.apache.spark.ml.linalg.Matrix;
+import org.apache.spark.ml.linalg.MatrixUDT;
 import org.apache.spark.ml.linalg.SparseVector;
 import org.apache.spark.ml.linalg.Vector;
 import org.apache.spark.ml.linalg.VectorUDT;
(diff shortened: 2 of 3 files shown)


docs/src/operations/dql/select.md
```diff
@@ -244,6 +244,44 @@ These options control how data is read from Lance datasets. They can be set usin
 | `pushDownFilters`     | Boolean | `true`  | Whether to push down filter predicates to the Lance reader for optimized scanning.                                |
 | `topN_push_down`      | Boolean | `true`  | Whether to push down TopN (ORDER BY ... LIMIT) operations to Lance for optimized sorting.                         |
 
+## Restoring User-Defined Types on Read
+
+Spark writes UDT columns (e.g. MLlib's `VectorUDT`) to Lance as the UDT's underlying `sqlType` (a struct). Arrow has no UDT concept, so on read the column comes back as that struct rather than the original UDT. To get the UDT type back, set a SparkConf entry per column:
+
+```
+spark.lance.udt.restore.<column> = <udt-fqn>
+```
```
hamersaw: This feel very wrong. Spark config options are global so this would apply to all table reads in the session with a column named `<column>`. Should we instead be attaching some kind of metadata to the schema? This would have to occur at table create or with a manifest update, but feels a lot cleaner IMO. Thoughts?

LuciferYang: You were right. Switched to per-field `__udt` metadata on the Arrow field — write stamps the UDT class FQN, read rewraps. No `SparkConf`, no opt-in.

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
```diff
@@ -501,47 +505,124 @@ public void testVectorUDTNullRowRoundtrip() {
...
+            .add("id", DataTypes.IntegerType, false)
+            .add("vecs", DataTypes.createArrayType(vectorUDT, true), true);
+
+    Vector v1 = new DenseVector(new double[] {1.0, 2.0});
+    Vector v2 = new DenseVector(new double[] {3.0, 4.0});
+    List<Row> data = Arrays.asList(RowFactory.create(0, Arrays.asList(v1, v2)));
+
+    Dataset<Row> result = writeAndRead(schema, data, "udt_array_element");
+
+    DataTypes.createArrayType(vectorUDT, true);
```
hamersaw: I don't think this is necessary anymore?

hamersaw (commented):
The last few nits. Right now the `__udt` arrow metadata is maintained on every data type rather than just those can can use it. Should we strip it from non struct types?

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
LuciferYang: Removed in 585ba0d. Dead from an earlier version of the test.

LuciferYang:
Scoped to `udt.sqlType: StructType` in 585ba0d. That covers VectorUDT and MatrixUDT (every MLlib UDT uses struct sqlType). For non-struct shapes the data still persists via the underlying type; only the UDT wrapper degrades on read-back.

hamersaw (approved):
Looks great! Thanks.

LuciferYang:
Thank you @hamersaw

Issue #539, opened 2026-05-18
Blob v2 support: findings and plan


geruh:
Ref to original issue #505, and to add a bit more information. 

## The problem

Setting `file_format_version = 2.2` with `<col>.lance.encoding = blob` blows up because `SchemaConverter.addBlobMetadata()` unconditionally stamps the legacy `lance-encoding:blob` field metadata. Lance-core rejects that for format >= 2.2 and wants the v2 extension type instead.

### Read: expose the native descriptor struct

Blob v2 columns get exposed as a 5-field descriptor struct. Querying individual fields is cheap and never fetches blob bytes:

```SQL
DESCRIBE t;
-- content: struct<kind:short, position:long, size:long, blob_id:long, blob_uri:string>

SELECT content.size, content.kind, content.blob_id, content.blob_uri FROM t;
-- 4, 0, 1, "s3://..."
-- kind=0 (Inline), size=4 bytes, blob_id assigned by lance-core
```

- Detection via column metadata: lance-encoding:blob-v2 = true (connector key) or ARROW:extension:name = lance.blob.v2 (native Lance writes). No table-property fallback.
- Filter pushdown suppressed for tables with blob v2 columns (Spark evaluates WHERE post-scan); zonemap pruning still works.
- LanceArrowColumnVector gets a dual accessor so getChild() walks struct children for SELECT content.size.
- v1 blobs unchanged. BinaryType with __blob_pos/__blob_size virtual columns.

### Write: BINARY in, struct out

Users write BINARY to align with V1. The connector produces the logical blob struct that lance-core's BlobPreprocessor expects.

```SQL
CREATE TABLE t (id INT, content BINARY) USING lance
TBLPROPERTIES (
  'content.lance.encoding' = 'blob',
  'file_format_version' = '2.2'
);

INSERT INTO t VALUES (1, X'DREW');
```

Spark schema stays `BinaryType`. Struct conversion happens only at the Arrow level. Write schema (`BinaryType`) and read schema (descriptor struct) are cleanly separated; Spark's analyzer sees the correct type at each stage.

Alternatively, users can insert with the struct giving either uri or binary. But, this example follows the native V1 path.

### Version gating

| file_format_version | Encoding |
|---------------------|----------|
| unset / `2.0` / `2.1` / `stable` / `legacy` | v1 (`lance-encoding:blob = true`) — existing behavior |
| `2.2` / `2.3` / `next` | v2 (`ARROW:extension:name = lance.blob.v2`) |

[...]

Issue #505, opened 2026-05-05
Support blob v2 extension type for file format >= 2.2


puchengy:
### Problem

  When creating a table with `'file_format_version' = '2.2'` and `'<column>.lance.encoding' = 'blob'`, the write fails with:

  Error: Invalid user input: Legacy blob columns (field metadata key "lance-encoding:blob") are not supported
  for file version >= 2.2. Use the blob v2 extension type (ARROW:extension:name = "lance.blob.v2") and the
  new blob APIs (e.g. lance::blob::blob_field / lance::blob::BlobArrayBuilder).

  ### Root Cause

  `SchemaConverter.addBlobMetadata()` unconditionally sets the legacy `lance-encoding:blob` field metadata. For file format >= 2.2, lance-core requires the blob v2 Arrow extension type
  (`ARROW:extension:name = "lance.blob.v2"`) with a `Struct<data: LargeBinary?, uri: Utf8?>` schema instead.

  ### Proposed Changes

  1. When `file_format_version >= 2.2` and blob encoding is requested, emit the blob v2 extension type metadata and struct schema instead of legacy metadata.
  2. Update `BlobUtils` to add blob v2 constants and detection.
  3. Update `LanceArrowUtils.scala` to handle the blob v2 struct type on the read path.
  4. Keep legacy behavior for `file_format_version < 2.2`.

  ### Workaround

  Use `'file_format_version' = '2.1'` with legacy blob encoding.

puchengy:
fyi @jackye1995 @hamersaw

PR #548, opened 2026-05-21
feat: blob v2 descriptor read support

geruh:
related to #539 and #505 

Adds read support for blob v2 columns. Instead of hiding blob metadata behind virtual columns, we surface the raw descriptor struct directly to Spark like what the original issue was stating. 

```
struct<kind: short, position: long, size: long, blob_id: long, blob_uri: string>
```

Querying blob metadata is just a column projection now, no byte fetch:

```sql
SELECT id, payload.size, payload.kind FROM lance.ns.tbl;
```

A column is blob v2 when any of these hold:
- Arrow extension name `lance.blob.v2` is set in lance 
- metadata key `lance-encoding:blob-v2 = true`

Schema rewrite lives in `BlobUtils.applyBlobV2DescriptorSchema(...)`, called from `LanceDataset.schema()`, `LanceDataSource.inferSchema()`, and the `LanceScanBuilder` constructor.

**Filter pushdown**

Suppressed for any table with a v2 column. The previous per-predicate gate had to be reverted because `calc_eager_projection` panics on filters that never reference blob columns. Zonemap fragment pruning still runs. Need more investigation here 

**Testing**

- Added some tests for BlobUtils 
- Added some tests for LanceArrowColumnVector

No end-to-end pylance interop tests here. The connector can't write v2 blobs yet, so producing test data requires pylance as an external dep. E2E coverage lands naturally once the write path exists, extending `BaseBlobCreateTableTest`'s pattern. 

**Local verification**

Wrote a v2 dataset with pylance hitting all four `BlobKind` tiers (Inline, Packed, Dedicated, External) plus a null row, then read it back through Spark on this branch:

<details>
<summary>write_blob_v2.py</summary>

```python
import lance, pyarrow as pa

with open("/tmp/blob_v2_external.bin", "wb") as f:
    f.write(b"external blob payload " * 64)

values = [
    b"hi",                                    # 2 B → Inline (kind=0, ≤ 64KB)
    b"x" * (200 * 1024),                      # 200KB → Packed (kind=1, > 64KB, ≤ 4MB)
    b"y" * (5 * 1024 * 1024),                 # 5 MB → Dedicated (kind=2, > 4MB)
    "file:///tmp/blob_v2_external.bin",        # str → External (kind=3)
    None,                                      # null
]

table = pa.table({
    "id": pa.array([0, 1, 2, 3, 4], type=pa.int32()),
    "label": pa.array(["inline", "packed", "dedicated", "external", "null"]),
    "payload": lance.blob_array(values),
})

lance.write_dataset(table, "/tmp/blob_v2_kinds.lance",
    data_storage_version="2.2",
    allow_external_blob_outside_bases=True)
```
</details>

<details>
<summary>Spark output</summary>

```
> printSchema
root
 |-- id: integer (nullable = true)
 |-- label: string (nullable = true)
 |-- payload: struct (nullable = true)
 | |-- kind: short (nullable = true)
 | |-- position: long (nullable = true)
 | |-- size: long (nullable = true)
 | |-- blob_id: long (nullable = true)
 | |-- blob_uri: string (nullable = true)
 
 
> SELECT *
+---+---------+----------------------------------------------+
|id |label    |payload                                       |
+---+---------+----------------------------------------------+
|0  |inline   |{0, 0, 2, 0, }                               |
|1  |packed   |{1, 0, 204800, 1, }                          |
|2  |dedicated|{2, 0, 5242880, 2, }                         |
|3  |external |{3, 0, 0, 0, file:///tmp/blob_v2_external.bin}|
|4  |null     |{0, 0, 0, 0, }                               |
+---+---------+----------------------------------------------+

> projected fields
+---+---------+----+-------+-------+--------+--------------------------------+
|id |label    |kind|size   |blob_id|position|blob_uri                        |
+---+---------+----+-------+-------+--------+--------------------------------+
|0  |inline   |0   |2      |0      |0       |                                |
|1  |packed   |1   |204800 |1      |0       |                                |
|2  |dedicated|2   |5242880|2      |0       |                                |
|3  |external |3   |0      |0      |0       |file:///tmp/blob_v2_external.bin|
|4  |null     |0   |0      |0      |0       |                                |
+---+---------+----+-------+-------+--------+--------------------------------+

filter id >= 2
+---+---------+----+
|id |label    |kind|
+---+---------+----+
|2  |dedicated|2   |
|3  |external |3   |
|4  |null     |0   |
+---+---------+----+
```

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobUtils.java
@@ -13,12 +13,35 @@
  */
 package org.lance.spark.utils;
 
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.spark.sql.types.DataTypes;
+import org.apache.spark.sql.types.Metadata;
 import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+
+import java.util.List;
+import java.util.Map;
 
 public class BlobUtils {
 
   public static final String LANCE_ENCODING_BLOB_KEY = "lance-encoding:blob";
   public static final String LANCE_ENCODING_BLOB_VALUE = "true";
+
+  public static final String ARROW_EXTENSION_NAME_KEY = "ARROW:extension:name";
+  public static final String ARROW_EXTENSION_BLOB_V2 = "lance.blob.v2";
+
+  /**
+   * Spark struct type for a Lance blob v2 descriptor: {@code kind, position, size, blob_id,
+   * blob_uri}.
+   */
+  public static final StructType BLOB_DESCRIPTOR_STRUCT =
+      new StructType()
+          .add("kind", DataTypes.ShortType)
+          .add("position", DataTypes.LongType)
+          .add("size", DataTypes.LongType)
+          .add("blob_id", DataTypes.LongType)
+          .add("blob_uri", DataTypes.StringType);
 
   /**
    * Check if a Spark field is a blob field based on its metadata.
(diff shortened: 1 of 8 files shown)


hamersaw:
Great PR! I think this opens a few questions. In https://github.com/lance-format/lance-spark/pull/355 we merged a similar concept for v1 blobs but take a quite different approach, maybe it's worth stating goals here to understand advantages / disadvantages.

That work chose to encode metadata into a binary format and then transparently retrieve when writing which gives us:
(1) native binary types in the schema: Able to call `getBinary` internally to grab the binary data.
(2) write side coalesced reads: If the `BlobReference` type is found we can batch together `take` operations when writing so that we don't incur 1 IOP per row.

Alternatively, this approach returns a struct with metadata, which makes it super easy to parse information about the blob but IMO may make actually retrieving the data a little less ergonomic and more expensive. I'm interested on your thoughts here?

SDK-wise I think it's a tradeoff of the root uses, with either approach to retreive both metadata and actual binary data we would probably need to add Spark fucntoins. For example:
```
# encoding as binary data
SELECT x, lance_blob_size(x) FROM y

# encoded as struct
SELECT lance_blob(x), x.size FROM y
```

But I think the performance of our use-case should be the top concern here.

geruh:
Thanks for the review @hamersaw!  Yeah, I agree that this is kind of a different goal from 355. For this PR, I really just wanted to scope v2 reads to the descriptor model to follow what people might be used to when interacting with LanceDB or Lance raw APIs. And that is defaulting to a descriptor read mode. 
 
In this mode, materialization is cheap because it's just the descriptor column value, that's intentionally different from the v1 binary blob reference path, which optimizes copy through Spark and right-side preservation by keeping `getBinary` semantics and batching blob fetches.
 
For fetching the actual bytes, we can add an explicit follow-up read mode similar to the blob handling modes defined on the LanceDB side(i.e. `lazy`, `materialize`). That follow-up would have that same performance story we have in V1 with the additional gains that stem from V2 on the format side.

Ultimately, my intent here is to keep things simple with descriptor first, follow up with read modes.

hamersaw:
For posterity - we addressed the conversation ^^^ offline. The consensus is that materializing the blob metadata as a `Struct` field in the Spark schema is a reasonable approach as long as it's well documented.

The only other comment I have is that the PR description suggests that filter pushdowns:
```
Suppressed for any table with a v2 column. The previous per-predicate gate had to be reverted because calc_eager_projection panics on filters that never reference blob columns. Zonemap fragment pruning still runs. Need more investigation here
```

I'm not seeing this logic anywhere in the code? It looks like the `hasBlobV2Fields` function was originally meant for this, but it doesn't seem to be used anywhere. I think it's OK to allow filter pushdowns as long as they're not on the blobv2 column specifically right?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobUtils.java
```diff
@@ -13,13 +13,36 @@
...
+  /**
+   * Spark struct type for a Lance blob v2 descriptor: {@code kind, position, size, blob_id,
+   * blob_uri}.
+   */
+  public static final StructType BLOB_DESCRIPTOR_STRUCT =
+      new StructType()
+          .add("kind", DataTypes.ShortType)
+          .add("position", DataTypes.LongType)
+          .add("size", DataTypes.LongType)
+          .add("blob_id", DataTypes.LongType)
```
hamersaw, line 43: Does it make sense to store position / size or would it be better to store the _rowaddr? When we read this the former means we need to open an objectstore reader and request the bytes, if we use `_rowaddr` then we can use the lance-native tools to call like `take_blobs` I think which handles coalesced reads / etc.

geruh: Sorry for the late response here Dan. Got sidetracked with some other work. 

Row addresses should already be exposed on the same row along side the descriptor as a metadata column.

https://github.com/lance-format/lance-spark/blob/20345deed0ee43b545647c9c2b9c88fd3adb4dbf/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java#L99-L104

I think adding it to the struct would be redundant, and kind of blur the the line between row identity, and storage layout.

PR #550, opened 2026-05-21
test: pin DELETE on a VectorUDT column against position-delta path

LuciferYang:
## Summary

Final item in the UDT-write series (after #471, #546, #549). Audit of write paths showed all lanes — standard `LanceBatchWrite`, position-delta `SparkPositionDeltaWrite` (DELETE / UPDATE / MERGE INTO), and `AbstractBackfillWriter` — funnel through the same `LanceArrowWriter` + `LanceArrowUtils` plumbing that #471 and #549 already fix. No new code fix needed; this PR adds one smoke test against the catalog + position-delta DELETE path to lock that in.

The docstring on `testDeleteOnVectorUDTColumn` documents what's intentionally NOT covered (PR #471's UDT case is reached only from `.save(path)`, and DELETE doesn't call `writer.write()` so #549's relaxation isn't exercised either).

## Test plan

- [ ] CI passes (lint, unit tests, integration tests across all Spark/Scala versions)
- [ ] \`testDeleteOnVectorUDTColumn\` passes
- [ ] No regression in existing \`MergeIntoTest\` (4 tests including the new one)

### DIFF
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseMergeIntoTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseMergeIntoTest.java
@@ -13,9 +13,15 @@
  */
 package org.lance.spark.update;
 
+import org.apache.spark.ml.linalg.DenseVector;
+import org.apache.spark.ml.linalg.Vector;
+import org.apache.spark.ml.linalg.VectorUDT;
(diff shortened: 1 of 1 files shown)


yanghua (approved):
LGTM

LuciferYang:
Thank you @yanghua ~

PR #551, opened 2026-05-21
fix: wrong LANCE_SPARK_VERSION and package name

yanghua:
Currently, the Dockerfile has two issues:

* The prompt for extracting `LANCE_SPARK_VERSION` is not right;
* The package name is also wrong;

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -42,7 +42,7 @@ WORKDIR ${SPARK_HOME}
 ENV SPARK_VERSION=${SPARK_DOWNLOAD_VERSION}
 ENV SPARK_MAJOR_VERSION=${SPARK_MAJOR_VERSION}
 ENV SCALA_VERSION=${SCALA_VERSION}
-ENV LANCE_SPARK_VERSION=0.5.0
+ENV LANCE_SPARK_VERSION=0.5.0-beta.1
 ENV LANCE_NS_VERSION=0.7.5
 
 # Download spark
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -52,11 +52,11 @@ RUN mkdir -p ${SPARK_HOME} \
  && rm -rf spark.tgz
 
 # Download lance-spark bundle JAR from Maven
-RUN curl -fL https://repo1.maven.org/maven2/com/lancedb/lance-spark-bundle-3.5_2.12/${LANCE_SPARK_VERSION}/lance-spark-bundle-3.5_2.12-${LANCE_SPARK_VERSION}.jar \
+RUN curl -fL https://repo1.maven.org/maven2/org/lance/lance-spark-bundle-3.5_2.12/${LANCE_SPARK_VERSION}/lance-spark-bundle-3.5_2.12-${LANCE_SPARK_VERSION}.jar \
  -o /opt/spark/jars/lance-spark-bundle_2.12-${LANCE_SPARK_VERSION}.jar
 
 # Download additional Lance namespace implementation jars from Maven
-RUN curl -fL https://repo1.maven.org/maven2/com/lancedb/lance-namespace-glue/${LANCE_NS_VERSION}/lance-namespace-glue-${LANCE_NS_VERSION}.jar \
+RUN curl -fL https://repo1.maven.org/maven2/org/lance/lance-namespace-glue/${LANCE_NS_VERSION}/lance-namespace-glue-${LANCE_NS_VERSION}.jar \
```
geruh: This curl returns a 404 for me.

```
curl: (56) The requested URL returned error: 404
```

Listing in`org/lance/lance-namespace-glue/` only shows `0.1.0/`, `0.1.3/`, `0.2.0/`, `0.3.0/`. There is no `0.7.5`. and you can see that with:

```
curl -s https://repo1.maven.org/maven2/org/lance/lance-namespace-glue/maven-metadata.xml
```
I think this is because, it was initially wrong, and `LANCE_NS_VERSION` is sourced from the wrong `pom.xml` property, and  `lance-namespace-glue` is in the impl family.

```xml
<lance-namespace.version>0.7.5</lance-namespace.version>          <!-- core / -apache-client / -async-client -->
<lance-namespace-impl.version>0.3.0</lance-namespace-impl.version> <!-- impl artifacts: -glue, -hive3, -iceberg, etc. -->
```
So we probably need to update this to use something like: `LANCE_NAMESPACE_IMPL_VERSION=0.3.0`

The codex file will also need to reflect this

yanghua:
@geruh Thanks, you are right. have addressed your suggestion.

hamersaw (approved):

PR #556, opened 2026-05-25
feat: expose scan stats to spark sql metrics

fangbo:


fangbo:
@yanghua Could you please take a look at it ?

yanghua:
Got it! Will review it later.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
```diff
@@ -54,6 +55,9 @@ public boolean next() throws IOException {
         return true;
       }
     }
+    if (fragmentReader != null) {
+      metricsTracker.addScanStats(fragmentReader.getScanStats());
```
yanghua: In some cases, this logic may cause duplicated accumulation. Because the `next()` method may be re-called in some scenarios. So, it would be better to accumulate it when `fragmentReader#close()`, or do something like this:

```
private boolean currentReaderStatsAdded = false;

while (...) {
    if (fragmentReader != null) {
        if (!currentReaderStatsAdded) {
            metricsTracker.addScanStats(fragmentReader.getScanStats());
            currentReaderStatsAdded = true;
        }
        fragmentReader.close();
    }
    fragmentReader = LanceFragmentColumnarBatchScanner.create(...);
    currentReaderStatsAdded = false;
    ...
}
if (fragmentReader != null && !currentReaderStatsAdded) {
    metricsTracker.addScanStats(fragmentReader.getScanStats());
    currentReaderStatsAdded = true;
}
return false;
```

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
yanghua, line 95: Maybe we also need to accumulate the metrics here to avoid the missing? We can combine with the `currentReaderStatsAdded` (in the front comment) flag to avoid duplicate accumulation.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/metric/BaseLanceReadMetricsIntegrationTest.java
```diff
@@ -174,6 +174,25 @@ void testReadMetricsValues() throws Exception {
     assertTrue(
         metrics.getOrDefault(LanceCustomMetrics.BATCH_LOAD_TIME_NS, 0L) > 0,
         "batchLoadTimeNs should be > 0");
+
+    // Test new metrics
+    assertTrue(
+        metrics.getOrDefault(LanceCustomMetrics.NUM_IOPS, 0L) >= 0, "numIops should be >= 0");
```
yanghua: May these assertions not work? Because the default value is `0`, right?

lance-spark-base_2.12/src/test/java/org/lance/spark/read/metric/BaseLanceCustomMetricsTest.java
```diff
@@ -31,7 +35,7 @@ public abstract class BaseLanceCustomMetricsTest {
   @Test
   void testAllMetricsReturnsSixMetrics() {
```
yanghua: `Six` is not correct now, right?

docs/src/performance.md
```diff
@@ -174,6 +174,12 @@ Naming conventions:
 | `numFragmentsScanned` | counter | Lance fragments actually opened by this task. Compare against the table fragment count to verify pruning is working. |
 | `numBatchesLoaded` | counter | Arrow batches returned from the JNI scanner. |
 | `numRowsScanned` | counter | Rows read from storage before filter evaluation. Pair with Spark's built-in `numOutputRows` to compute filter selectivity (`numOutputRows / numRowsScanned`). |
+| `numIops` | counter | Number of I/O operations performed during scanning. |
```
yanghua: Number of I/O operations performed by the scanner.

docs/src/performance.md
```diff
@@ -174,6 +174,12 @@ Naming conventions:
...
+| `numRequests` | counter | Number of requests made to the storage layer. |
+| `numBytesRead` | counter | Total bytes read from storage. |
```
yanghua: Total bytes read from storage by the scanner.

yanghua (commented):
Left some comments.

fangbo:
@yanghua I have fixed the CR comments. Could you please look at it again? Thanks a lot.

yanghua (approved):
LGTM. @fangbo thanks

PR #558, opened 2026-05-25
feat: support deferred index creation with WITH (train=false)

ivscheianu:
## Problem

`ALTER TABLE ... CREATE INDEX` always builds the index eagerly over all existing data before returning. On large datasets this is a long-running, blocking operation. There was no way to register an index definition without immediately training it, making it impossible to separate the "declare the index" step from the "populate it" step.

## Solution

Add a `train` option to the `WITH` clause:

```sql
ALTER TABLE my_table CREATE INDEX idx_text_fts USING fts (text)
WITH (train=false)
```

When `train=false`, no Spark tasks are launched and no data is processed. An empty index is committed directly on the driver with an empty fragment bitmap — all existing rows appear as unindexed. A subsequent `OPTIMIZE` call covers them incrementally at a time of the caller's choosing.

The default (`train=true`) is unchanged, so all existing behaviour is preserved.

As a side fix, Spark-level execution options (`train`, `build_mode`, `rows_per_range`) are now filtered before being forwarded to the Lance index backend as index parameters.


hamersaw (approved):
Thanks for the PR! I added support for ZoneMap indexes, it's probably not useful the way BTree + FTS is but at least we're complete. Once CI passes, lets merge!

ivscheianu:
> Thanks for the PR! I added support for ZoneMap indexes, it's probably not useful the way BTree + FTS is but at least we're complete. Once CI passes, lets merge!

Thanks for the review and the additional support, @hamersaw!

PR #559, opened 2026-05-26
fix: widen narrow numeric literals in zonemap pruner

### LINKED ISSUE #557: exception throws when zonemap pruner on int column
# Description

## Table Schema
```json
{
  "fields": [
    {
      "{ name": "seq",
      "id": 0,
      "parent_id": -1,
      "logical_type": {
        "type": "int32"
      },
      "metadata": {},
      "encoding": "Plain",
      "nullable": true,
      "children": [],
      "dictionary": null,
      "unenforced_primary_key_position": null
    },
    {
      "{ name": "content",
      "id": 1,
      "parent_id": -1,
      "logical_type": {
        "type": "string"
      },
      "metadata": {},
      "encoding": "VarBinary",
      "nullable": false,
      "children": [],
      "dictionary": null,
      "unenforced_primary_key_position": null
    }
  ],
  "metadata": {}
}
```

A zonemap index is created on the column `seq`.

## Exception

The sql:

```sql
select * from zonemap_test where seq=1;
```

throws an exception:

```java
java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Integer (java.lang.Long and java.lang.Integer are in module java.base of loader 'bootstrap')
        at java.base/java.lang.Integer.compareTo(Integer.java:71)
        at org.lance.spark.read.ZonemapFragmentPruner.zoneMatchesComparison(ZonemapFragmentPruner.java:209)
        at org.lance.spark.read.ZonemapFragmentPruner.analyzeComparison(ZonemapFragmentPruner.java:184)
        at org.lance.spark.read.ZonemapFragmentPruner.analyzeFilter(ZonemapFragmentPruner.java:115)
        at org.lance.spark.read.ZonemapFragmentPruner.pruneFragments(ZonemapFragmentPruner.java:88)
        at org.lance.spark.read.LanceScanBuilder.build(LanceScanBuilder.java:186)
```

LuciferYang:
Closes #557.

`ZonemapFragmentPruner` throws `ClassCastException: Long cannot be cast to Integer` whenever a `WHERE int32_col = N` query hits a column with a zonemap index. The two sides of `target.compareTo(min)` use different Java boxes:

- Lance JNI (`scalar_value_to_java` in `lance-jni/utils.rs`) materializes `ZoneStats.min/max` with every integer width as `Long` and every float width as `Double`.
- Spark V2 `Literal.value()` keeps the Catalyst type — `Integer` for `IntegerType`, `Short` for `ShortType`, `Byte` for `ByteType`, `Float` for `FloatType`, `Integer` epoch days for `DateType`.

`Integer.compareTo(Object)` then rejects the `Long`. Same mismatch affects `tinyint`, `smallint`, `float`, and `date` predicates, plus the IN-list path. The existing `catch (ClassCastException)` in `zoneMatchesComparison` does fire on the reported workload but means we silently drop pruning instead of getting it.

## Fix

Widen the narrow boxes inside `normalizeLiteral` to match the JNI wire types: `Byte`/`Short`/`Integer` → `Long`, `Float` → `Double`. Both conversions are lossless and order-preserving; `analyzeIn` already routes each list element through the same normalizer so it gets the fix for free.

## Tests

Seven new cases in `ZonemapFragmentPrunerTest` — one per affected literal type (Integer / Short / Byte / Float / Date) plus IN-list coverage (homogeneous and mixed-width). Verified by reverting `normalizeLiteral` locally and watching all seven fail through the conservative-catch fallback.

## Out of scope (surfaced during review)

Pre-existing JNI gaps, not regressions from this PR — noting for follow-up:

- `ScalarValue::Decimal*` and other unhandled types box to `null` for both `min` and `max`. The pruner then treats those zones as all-null and excludes them — risk of silent under-pruning on decimal predicates.
- `Date64` and the four `Timestamp*` variants are all flattened to `Long` without a unit tag, so columns written with non-`Date32` / non-microsecond units cannot be safely pruned against Spark's days/micros literals.

## Test plan

- [x] \`make lint\` clean
- [x] \`./mvnw test -pl lance-spark-base_2.12 -Dtest=ZonemapFragmentPrunerTest\` (37/37 passes locally)
- [x] CI green across all Spark/Scala modules

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
@@ -335,13 +335,22 @@ public final class ZonemapFragmentPruner {
   }
 
   /**
-   * V2 {@link Literal} exposes values in Spark's internal representation ({@code UTF8String} for
-   * strings). Zone stats from lance-core store String values — normalize here so {@code compareTo}
-   * against min/max works.
+   * V2 {@link Literal} exposes values in Spark's internal representation, while Lance's JNI
+   * materializes {@code ZoneStats.min/max} with every integer width boxed as {@code Long} and every
+   * floating-point width as {@code Double}. Widen narrow Java boxed primitives (Byte / Short /
+   * Integer / Float) to match — otherwise an Integer literal against a Long zone bound would throw
+   * {@code ClassCastException} from {@code Comparable.compareTo}. Also normalizes {@code
+   * UTF8String} → {@code String} for the same reason.
    */
   private static Object normalizeLiteral(Object value) {
     if (value instanceof UTF8String) {
       return value.toString();
+    }
+    if (value instanceof Byte || value instanceof Short || value instanceof Integer) {
+      return ((Number) value).longValue();
+    }
+    if (value instanceof Float) {
+      return ((Float) value).doubleValue();
     }
     return value;
   }
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
@@ -20,6 +20,7 @@ import org.apache.spark.sql.connector.expressions.FieldReference;
 import org.apache.spark.sql.connector.expressions.filter.Predicate;
 import org.junit.jupiter.api.Test;
 
+import java.sql.Date;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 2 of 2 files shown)


LuciferYang:
cc @fangbo FYI

fangbo:
+1, Thanks for fixing it !

LuciferYang:
Thank you @fangbo

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
```diff
@@ -335,14 +335,23 @@ private static String columnName(NamedReference ref) {
...
+   * Integer / Float) to match — otherwise an Integer literal against a Long zone bound would throw
+   * {@code ClassCastException} from {@code Comparable.compareTo}. Also normalizes {@code
+   * UTF8String} → {@code String} for the same reason.
    */
   private static Object normalizeLiteral(Object value) {
     if (value instanceof UTF8String) {
       return value.toString();
     }
+    if (value instanceof Byte || value instanceof Short || value instanceof Integer) {
+      return ((Number) value).longValue();
```
summaryzb, line 350: Maybe this can be resolved by https://github.com/lance-format/lance/pull/6906/changes#diff-375a608c3405295ba4ec1a28f04c8d1ca8721342d0609f955a385f88c8e47d09R132

PR #567, opened 2026-05-31
perf(read): consolidate driver-side Dataset opens and pin version for…

yanghua:
… snapshot isolation

## Summary

Consolidate the two driver-side `Dataset.open()` calls during scan planning into one, and pin the resolved table version onto the read options shipped to executors. This reduces manifest IO and closes a cross-task snapshot-isolation gap on large tables.

## Motivation

For each Spark scan, `LanceScanBuilder.build()` opens a dataset to read manifest summary, schema, sharding spec, and zonemap stats. Then `LanceScan.planInputPartitions()` opens the dataset *again* via `LanceSplit.planScan(readOptions)` to enumerate fragments and per-fragment row counts.

Two issues:

1. **Performance** — Driver pays the manifest IO / `Dataset.open()` cost twice per query. On very large tables this measurably increases planning latency.
2. **Snapshot isolation bug** — When the user does not specify a version, both opens resolve "latest" independently. If a concurrent writer commits a new version between the two opens (or between the driver-side open and an executor-side open), tasks can observe a newer version than the one used for fragment pruning / statistics. The query then sees an inconsistent view.

### `LanceSplit`
- New overload `planScan(Dataset)` that accepts an already-opened dataset and does not close it.
- Existing `planScan(LanceSparkReadOptions)` becomes a thin wrapper, kept for tests and external callers.

### `LanceScanBuilder.build()`
- Calls `LanceSplit.planScan(dataset)` against the same handle used for manifest / zonemap loading, before `closeLazyDataset()`.
- Calls `readOptions.withVersion(resolvedVersion)` to produce a pinned, immutable copy of the read options.
- Passes the pre-computed splits, per-fragment row counts, and pinned options to `LanceScan`.

### `LanceScan`
- Constructor accepts `precomputedSplits` and `precomputedFragmentRowCounts`.
- `planInputPartitions()` no longer opens the dataset; it consumes the pre-computed result and skips the redundant `withVersion` wrap (the options are already pinned upstream).
- The per-fragment row-count map is marked `transient` so it does not bloat plan-tree serialization or affect `BatchScanExec` / `ReusedExchange` comparisons.

### Tests
- `LanceScanTest` updated for the new constructor parameters.
- `LanceSplitTest` unchanged — both overloads remain covered.

## Why this is safe

- `LanceSparkReadOptions.withVersion(...)` returns a new instance via the existing builder; the user-supplied options object is never mutated.
- When the user explicitly pinned a version upstream, `dataset.getVersion().getId()` returns that same version, so the pin is a no-op.
- All existing pruning paths (`pruneByRowAddrFilters`, `pruneByZonemapStats`, `pruneByLimit`) continue to operate on the same `List<LanceSplit>` shape; only the source of the list changed.

## Performance impact

- Driver `Dataset.open()` calls per scan: **2 → 1**.
- Manifest reads per scan: **2 → 1**.
- No change to executor-side IO.

## Correctness impact

- All tasks of a single query are now guaranteed to read the same dataset version, even under concurrent writes.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
@@ -107,7 +107,7 @@ public class LanceSparkReadOptions implements Serializable {
   private final String datasetName;
   private final boolean pushDownFilters;
   private final Integer blockSize;
-  private final Integer version;
+  private final Long version;
   private final Integer indexCacheSize;
   private final Integer metadataCacheSize;
   private final int batchSize;
@@ -238,7 +238,7 @@ public class LanceSparkReadOptions implements Serializable {
     return blockSize;
   }
 
-  public Integer getVersion() {
+  public Long getVersion() {
     return version;
   }
 
@@ -312,7 +312,7 @@ public class LanceSparkReadOptions implements Serializable {
    * @param newVersion the version to use
    * @return a new LanceSparkReadOptions with the specified version
    */
-  public LanceSparkReadOptions withVersion(int newVersion) {
+  public LanceSparkReadOptions withVersion(long newVersion) {
     return builder()
         .datasetUri(this.datasetUri)
         .pushDownFilters(this.pushDownFilters)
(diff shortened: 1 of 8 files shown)


fangbo:
+1, Good performance improvement !

PR #571, opened 2026-06-02
docs: upgrade lance spark bundle version

eddyxu:
It seems that `claude code` picks the version from the docs, which is still `com.lancedb:lance-bundle-spark_..:0.0.7`. 

Update the version for future vibe coding examples

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -33,14 +33,14 @@ Lance provides SQL extensions that add additional functionality beyond standard 
 === "Spark Shell"
     ```shell
     spark-shell \
-      --packages org.lance:lance-spark-bundle-3.5_2.12:0.0.7 \
+      --packages org.lance:lance-spark-bundle-3.5_2.12:0.4.0 \
       --conf spark.sql.extensions=org.lance.spark.extensions.LanceSparkSessionExtensions
     ```
 
 === "Spark Submit"
     ```shell
     spark-submit \
-      --packages org.lance:lance-spark-bundle-3.5_2.12:0.0.7 \
+      --packages org.lance:lance-spark-bundle-3.5_2.12:0.4.0 \
       --conf spark.sql.extensions=org.lance.spark.extensions.LanceSparkSessionExtensions \
       your-application.jar
     ```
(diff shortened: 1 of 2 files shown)


docs/src/config.md
```diff
@@ -291,7 +291,7 @@ Using the Glue namespace requires additional dependencies beyond the main Lance
 Example with Spark Shell:
 ```shell
 spark-shell \
-  --packages org.lance:lance-spark-bundle-3.5_2.12:0.0.7,org.lance:lance-namespace-glue:0.0.7,software.amazon.awssdk:bundle:2.20.0 \
+  --packages org.lance:lance-spark-bundle-3.5_2.12:0.4.0,org.lance:lance-namespace-glue:0.7.7,software.amazon.awssdk:bundle:2.20.0 \
```
geruh: The lance ns glue jar doesn't exist, and I think should break? The lance ns core is 0.7.7, but glues impl artifact is latest 0.3.0.

```
❯ curl -s https://repo1.maven.org/maven2/org/lance/lance-namespace-glue/maven-metadata.xml
<?xml version="1.0" encoding="UTF-8"?>
<metadata>
  <groupId>org.lance</groupId>
  <artifactId>lance-namespace-glue</artifactId>
  <versioning>
    <latest>0.3.0</latest>
    <release>0.3.0</release>
    <versions>
      <version>0.1.0</version>
      <version>0.1.3</version>
      <version>0.2.0</version>
      <version>0.3.0</version>
    </versions>
    <lastUpdated>20260428025922</lastUpdated>
  </versioning>
</metadata>
```

I think it should be  `--packages org.lance:lance-spark-bundle-3.5_2.12:0.4.0,org.lance:lance-namespace-glue:0.3.0,software.amazon.awssdk:bundle:2.20.0 \` matching the hive3 example

eddyxu: kk, thanks

Should i bump awssdk version ?

geruh: I don't think it's required, I think we could do that as a follow up since i see a gap between our integration test version and the release dep version. 

Our integration tests are ran against 2.20.43, and the glue jar is compiled with 2.32.2

wdyt?

geruh (approved):

PR #573, opened 2026-06-02
perf(read): read compact Arrow decimals directly

LuciferYang:
## Summary

Decimal reads currently fall through Spark's Arrow object path, which materializes a `BigDecimal` before building the Catalyst `Decimal`. That is unnecessary for compact Spark decimals: when `precision <= 18`, the unscaled value is guaranteed to fit in a `long`.

This adds a `DecimalVector` accessor in `LanceArrowColumnVector` that keeps Spark's existing wide-decimal behavior, but reads compact decimals directly from the Arrow data buffer.

## How it works

Arrow decimal128 stores each value as a 16-byte two's-complement unscaled integer. For `precision <= 18`, the high eight bytes are only sign extension, so the low eight bytes contain the complete Spark compact-decimal value.

The new accessor:

- routes `DecimalVector` through Lance's own accessor instead of the generic Spark `ArrowColumnVector` wrapper;
- returns null before touching the value buffer;
- reads the low 64 bits directly for compact decimals and constructs `Decimal` with `Decimal.createUnsafe`;
- keeps the existing `Decimal.apply(vector.getObject(...))` path for `precision > 18`;
- routes `hasNull`, `numNulls`, `isNullAt`, and `close` through the same decimal accessor, so an owned decimal vector is released on close exactly like every sibling accessor.

This follows Spark's compact-decimal representation used in `WritableColumnVector` / `UnsafeRow`, while preserving Spark `ArrowColumnVector`'s object-path semantics for wider decimals.

## Tests

Added direct `LanceArrowColumnVectorTest` coverage for:

- `precision == 18` fast path — positive values, negative values, nulls, and null counts;
- `precision == 1` boundary, positive and negative;
- `precision > 18` fallback through the Arrow object path;
- owning-close lifecycle — `closeVectorOnClose=true`, then assert the allocator is fully drained after `close()` (guards against the decimal vector leaking).

## Test plan

- [x] CI passes across supported Spark / Scala modules
- [x] Unit coverage added for compact, boundary, wide, and owning-close paths
- [x] `make lint` clean
- [x] Decimal suite compiles and passes against Spark 4.1

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/LanceArrowColumnVector.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/LanceArrowColumnVector.java
@@ -15,7 +15,9 @@ package org.lance.spark.vectorized;
 
 import org.lance.spark.utils.BlobUtils;
 
+import org.apache.arrow.memory.ArrowBuf;
 import org.apache.arrow.vector.DateMilliVector;
+import org.apache.arrow.vector.DecimalVector;
 import org.apache.arrow.vector.FixedSizeBinaryVector;
 import org.apache.arrow.vector.LargeVarCharVector;
 import org.apache.arrow.vector.TimeMicroVector;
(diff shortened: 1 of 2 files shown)


LuciferYang:
cc @hamersaw @fangbo

fangbo:
+1, Great optimization

hamersaw (approved):
Looks great! Just waiting for CI to pass and we'll merge.

LuciferYang:
Thanks @hamersaw @fangbo

PR #574, opened 2026-06-02
feat: add SQL vector_search table-valued function

### LINKED ISSUE #65: vector and full text search SQL extension

summaryzb:
## Summary

This PR adds a Spark SQL `vector_search` table-valued function for querying Lance vector columns through the Lance Spark SQL extension. The implementation resolves catalog tables or direct Lance paths, executes Lance nearest-neighbor scans per fragment and applies a Spark-side global TopK by Lance's internal `_distance` value while keeping `_distance` out of user-visible results.

## Problem
Resolve https://github.com/lance-format/lance-spark/issues/65
Lance Spark already supports nearest-neighbor reads through read options, but SQL users did not have a first-class query form for vector search. Users needed a SQL extension entry point that works with both Lance catalog tables and direct dataset paths, validates query arguments consistently, preserves fragment-parallel scans, and returns globally ordered TopK results without exposing connector-internal distance metadata.

## Approach

The PR registers Lance table-valued functions through `LanceSparkSessionExtensions` and adds a resolution rule that rewrites unresolved Lance TVF logical nodes once their arguments are resolved. `vector_search(table, column, query_vector, limit)` validates foldable literal arguments, resolves the target as either a catalog `LanceDataset` or path-based Lance dataset, builds a Lance nearest-neighbor `Query`, and passes it through `LanceSparkReadOptions.CONFIG_NEAREST` into a `DataSourceV2Relation`.

To support final ranking, the relation schema is widened with a reserved internal `_distance` column. Fragment scanners request `_distance` from Lance only for nearest-neighbor scans, while normal column projection continues to filter internal columns. The logical plan then sorts globally by `_distance`, applies local and global limits, and projects `_distance` away so the public result schema matches the source table. This explicitly adds the cross-fragment TopK merge that the earlier DataFrame nearest API did not perform. TopN pushdown is disabled when a nearest-neighbor query is active so Spark keeps control of the final global ordering step.

## Notice
PathOnly table pattern changes from "file://" to "file:/"

## Test Coverage

- Catalog-table `vector_search` returns the expected global TopK result ordering.
- `vector_search` keeps multiple fragment scan partitions and leaves final TopK ordering to Spark.
- Internal `_distance` is not exposed in `SELECT *` results.
- Direct-path `vector_search` works against a Lance dataset URI and applies the requested global limit.
- Invalid non-positive limits are rejected with an `IllegalArgumentException`.


lance-spark-base_2.12/src/main/java/org/lance/spark/LanceConstant.java
```diff
@@ -17,6 +17,7 @@ public class LanceConstant {
   public static final String FRAGMENT_ID = "_fragid";
   public static final String ROW_ID = "_rowid";
   public static final String ROW_ADDRESS = "_rowaddr";
+  public static final String VECTOR_DISTANCE = "_distance";
```
LuciferYang, line 22: There are duplicate constant definitions in `LanceFragmentScanner.java:40` and `LanceTableValuedFunctions.scala:36`.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+
+  private def resolveCatalogTable(spark: SparkSession, tableName: String): ResolvedLanceTable = {
+    val catalogManager = spark.sessionState.catalogManager
+    val parts = spark.sessionState.sqlParser.parseMultipartIdentifier(tableName)
+    val currentCatalog = catalogManager.currentCatalog.asInstanceOf[TableCatalog]
+
+    val (catalog, ident) = parts match {
+      case Seq(table) =>
+        (currentCatalog, Identifier.of(catalogManager.currentNamespace, table))
+      case Seq(namespace, table) =>
```
LuciferYang: I think we should replace the hand-rolled matcher with Spark's resolution
  (`LookupCatalog.CatalogAndIdentifier` against `spark.sessionState.catalogManager`), which already
  encodes the current-catalog/current-namespace rules consistently with the rest of Spark.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+      limited.output.filterNot(attr => resolver(attr.name, VECTOR_DISTANCE)),
+      limited)
+  }
+
+  private def createNearestQuery(column: String, queryVector: Array[Float], limit: Int): Query = {
+    val builder = new Query.Builder()
+    builder.setColumn(column)
+    builder.setKey(queryVector)
+    builder.setK(limit)
+    builder.setUseIndex(true)
```
LuciferYang: forcing `useIndex=true` means that on an *indexed* table users always get approximate ANN results and have no way to request exact search; and because `distanceType` is left unset, Lance applies its data-type default — L2 for float vectors (per `Query.Builder.setDistanceType`'s contract) — so vectors meant to be compared with cosine/dot distance cannot be queried with the matching metric. Both are silently-wrong-results risks, not merely missing knobs.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+    val resolvedTable =
+      if (looksLikePath(tableName)) {
+        resolvePathTable(tableName)
+      } else {
+        resolveCatalogTable(spark, tableName)
+      }
+
+    val resolver = spark.sessionState.conf.resolver
+    val sourceSchema = resolvedTable.table.schema()
+    val vectorField = sourceSchema.fields.find(field => resolver(field.name, requestedColumn))
```
LuciferYang: The resolver checks the column *exists* but never that it is a vector column before passing `vectorField.name` into the nearest query (`:95`). Pointing `vector_search` at a scalar column (e.g. an `INT`) builds an invalid kNN query that fails deep inside Lance at scan time on the executors, with a low-level message. 

Note that a plain `dataType is ArrayType` check is **insufficient** — ordinary variable-length array columns also map to `ArrayType`; a Lance vector column is specifically an `ArrayType(FloatType)` carrying fixed-size-list metadata. The repo already ships the right helpers: validate with `VectorUtils.isVectorField(vectorField)` during resolution (and optionally cross-check the query vector's length against `VectorUtils.getVectorDimension(vectorField)`), failing fast with a clear analysis-time error

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseSparkConnectorSqlVectorSearchTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+    assertEquals(1, result.collectAsList().size());
+    assertFalse(
+        java.util.Arrays.asList(result.schema().fieldNames()).contains("_distance"),
+        "vector_search should not expose _distance for path-based reads");
+  }
+
+  @Test
+  public void testVectorSearchRejectsInvalidLimit() {
+    String tableName = "invalid_limit_" + System.currentTimeMillis();
```
LuciferYang: A large amount of validation logic ships with only one negative test. Add cases for: 
1. reserved-`_distance` collision
2. unknown column
3. 2-part `catalog.table` resolution
4. non-array vector column 
5. non-integral limit literal
6. empty/null-element query vector, 
7. non-Lance table rejection, 
8. non-foldable argument. 

Prefer asserting on the exception **message substring** rather than the exact `IllegalArgumentException` type, so the tests stay valid if a future Spark version surfaces analyzer-time failures through a different exception class.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+  }
+
+  private def extractLimit(expr: Expression): Int = {
+    val value = evalFoldable(expr, "limit")
+    val limit = value match {
+      case n: java.lang.Byte => n.longValue()
+      case n: java.lang.Short => n.longValue()
+      case n: java.lang.Integer => n.longValue()
+      case n: java.lang.Long => n.longValue()
+      case n: java.lang.Number => n.longValue()
```
LuciferYang: A `FLOAT`/`DOUBLE` limit literal such as `10.5f` boxes to `java.lang.Float`/`Double`(both `java.lang.Number`), so it matches this catch-all and `longValue()` floors `10.5 → 10`, despite every nearby error message insisting "limit must be a positive integer."

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+      table.getFileFormatVersion)
+  }
+
+  private case class ResolvedLanceTable(
+      table: LanceDataset,
+      catalog: Option[TableCatalog],
+      identifier: Option[Identifier])
+
+  private def resolvePathTable(path: String): ResolvedLanceTable = {
+    val readOptions = LanceSparkReadOptions.from(path)
```
LuciferYang: Resolution opens the path dataset just to read the schema and closes it; each fragment scanner re-opens it at scan time, and catalog `loadTable` likewise hits storage in the analyzer. Acceptable, but it is real I/O inside a resolution rule (which can re-run on re-analysis, e.g. cached plans).

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,323 @@
...
+        case _ => None
+      }
+    } catch {
+      case NonFatal(_) => None
+    }
+  }
+
+  private def looksLikePath(value: String): Boolean = {
+    value.startsWith("/") ||
+    value.contains("/") ||
```
LuciferYang: A backtick-quoted catalog identifier containing `/` would be misclassified as a path; conversely a relative dataset name without `/` (e.g. `mydata.lance`) is treated as a catalog  identifier.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
summaryzb: follow suggestion

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
summaryzb: Pass nearest json string instead of the `Query Builder` element param

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
summaryzb: Remove this, limit can only be int, since it will be passed to the field `k` of `Query` in lance java side

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,210 @@
...
+    val tableName = extractString(args(0), "table")
+    val nearestJson = extractString(args(1), "nearest")
+    val nearestQuery = extractNearestQuery(nearestJson)
+    val limit = extractLimit(nearestQuery)
+
+    val resolvedTable = resolveCatalogTable(spark, tableName)
+
+    val resolver = spark.sessionState.conf.resolver
+    val sourceSchema = resolvedTable.table.schema()
+    if (sourceSchema.fields.exists(field => resolver(field.name, VECTOR_DISTANCE))) {
```
LuciferYang: resolveVectorSearch only ever reads `k` from the parsed query - it never checks that `column` is
present, exists in the table, or is actually a vector column. So vector_search('t', '{"column":"id",
"key":[...],"k":10}') against a scalar `id` (or a typo'd column name) passes analysis and only blows
up deep in Lance on the executors with a low-level message.

The JSON makes it sharper: QueryUtils sets FAIL_ON_UNKNOWN_PROPERTIES=false, so a misspelled key is
silently dropped. `"distance_type"` instead of `"distanceType"` just falls back to the default metric
and you get quietly wrong results, no error.

Could we validate `column` (and `key`) at resolution time? VectorUtils.isVectorField is already in
the repo, and getVectorDimension would let us sanity-check key.length too. Failing fast here with a
clear message beats an executor stack trace. Worth documenting the exact accepted JSON keys in
select.md as well, since unknown ones vanish silently. (Longer term a typed
vector_search(table, column, query_vector, k [, options]) signature would make the common case
SQL-native, but the validation is the important part.)

docs/src/operations/dql/select.md
```diff
@@ -53,6 +53,49 @@ Query data from Lance tables using SQL or DataFrames.
...
+
+```sql
+SELECT id, title
+FROM vector_search(
+  'lance.default.documents',
+  '{"column":"embedding","key":[0.12,0.34,0.56],"k":10,"distanceType":"Cosine","useIndex":true}'
+);
+```
+
+Direct path example:
```
LuciferYang: Heads up that the direct-path form in the docs doesn't actually work. resolveCatalogTable feeds the
table arg straight into parseMultipartIdentifier, and a bare path like '/data/lance/documents.lance'
isn't a valid multipart identifier, so it throws before we resolve anything. The only form that
parses is the catalog-qualified, backtick-quoted one the test uses (`lance_default`.`file:/...`).

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,210 @@
...
+import org.lance.ipc.Query
+import org.lance.spark.{LanceDataset, LanceSparkReadOptions}
+import org.lance.spark.utils.QueryUtils
+
+import scala.collection.JavaConverters._
+
+object LanceTableValuedFunctions {
+
+  val VECTOR_SEARCH = "vector_search"
+  private val VECTOR_DISTANCE = "_distance"
```
LuciferYang: `_distance` is hardcoded again here as `private val VECTOR_DISTANCE = "_distance"`,
separate from LanceConstant.VECTOR_DISTANCE.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseSparkConnectorReadWithVectorSearchTest.java
```diff
@@ -93,4 +88,207 @@ public void validateData() {
...
+    Dataset<Row> result = vectorSearch("i", nearestQuery(1));
+
+    assertEquals(1, result.collectAsList().size());
+    assertFalse(
+        Arrays.asList(result.schema().fieldNames()).contains("_distance"),
+        "vector_search should not expose _distance for path-based reads");
+  }
+
+  @Test
+  public void testVectorSearchDefaultsToExactL2Query() throws Exception {
```
LuciferYang: Happy-path coverage looks good; the error/edge cases are thin. Could we add negatives for: a column
that doesn't exist or is a scalar, a source table that already has a `_distance` column, and a
non-Lance or missing table? Also worth asserting the message rather than just that something throws -
testVectorSearchRejectsInvalidLimit currently asserts nothing about the message, unlike
testVectorSearchRejectsInvalidNearestJson.

One naming nit: testVectorSearchDefaultsToExactL2Query doesn't actually test a default -
nearestQueryBuilder sets useIndex(false) and distanceType(L2) explicitly. The real Query.Builder
default is useIndex=true (ANN) with no distance type, so the omitted-field behavior isn't covered.
Either rename it or add a test that genuinely omits those fields.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceTableValuedFunctions.scala
```diff
@@ -0,0 +1,210 @@
...
+  private def extractLimit(query: Query): Int = {
+    val limit = query.getK
+    if (limit <= 0) {
+      throw new IllegalArgumentException(
+        s"$VECTOR_SEARCH nearest query k must be a positive integer, but got $limit.")
+    }
+    limit
+  }
+
+  private def withDistanceColumn(table: LanceDataset): LanceDataset = {
```
LuciferYang, line 161: withDistanceColumn reaches for the 7-arg LanceDataset constructor, which quietly drops
tableProperties/stagedCommit/shardingSpec (they become empty/null). Harmless today on the read path -the scan re-derives the sharding spec and never forwards tableProperties - but it's a fragile way to
do "copy with a new schema". A LanceDataset.withSchema(...) helper that preserves the other fields
would be safer for the next person who touches this.

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -127,6 +127,7 @@ private static boolean isPathBasedIdentifier(Identifier ident) {
           || firstPart.startsWith("az://")
           || firstPart.startsWith("abfss://")
           || firstPart.startsWith("file://")
+          || firstPart.startsWith("file:/")
```
LuciferYang, line 129: anything starting with `file://` already starts with `file:/`. Could drop them, or collapse to the single `file:/` check.

Also worth calling out in the PR description that this isPathBasedIdentifier change affects all
catalog operations, not just vector search: any identifier beginning `file:/` is now treated as a
path everywhere.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseSparkConnectorReadWithVectorSearchTest.java
```diff
@@ -93,4 +88,207 @@ public void validateData() {
...
+            .append("', '")
+            .append(nearestJson)
+            .append("'");
+    sql.append(")");
+    return spark.sql(sql.toString());
+  }
+
+  private static Query scanNearestQuery(Dataset<Row> result) throws Exception {
+    LanceScan scan = lanceScan(result);
+    Field readOptionsField = LanceScan.class.getDeclaredField("readOptions");
```
LuciferYang: The distance-type tests pull readOptions out of LanceScan via reflection
(getDeclaredField("readOptions")), which breaks silently if the field is ever renamed. A small
package-private accessor on LanceScan (or asserting through the public plan/results) would be more
durable.

docs/src/config.md
```diff
@@ -499,4 +500,4 @@ Lance Spark maintains index and metadata caches to minimize redundant I/O. Cache
 | `LANCE_INDEX_CACHE_SIZE`   | 6GB     | Index cache size in bytes.       |
 | `LANCE_METADATA_CACHE_SIZE`| 1GB     | Metadata cache size in bytes.    |
 
-For details on how caching works and tuning recommendations, see [Performance Tuning - Caching](performance.md#caching).
\ No newline at end of file
+For details on how caching works and tuning recommendations, see [Performance Tuning - Caching](performance.md#caching).
```
LuciferYang: please revert this change

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BaseSparkConnectorReadWithVectorSearchTest.java
summaryzb: add negative test

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
summaryzb: Changes the PR description

LuciferYang:
cc @hamersaw
also cc @Xuanwo

summaryzb:
Closed, duplicate with https://github.com/lance-format/lance-spark/pull/582 , @LuciferYang  thanks for review

PR #576, opened 2026-06-02
feat: support ddl for branch operation

fangbo:
## Summary

This PR adds DDL support for branch operations in Lance Spark.

Previously, branch-related operations did not support DDL workflows properly. With this change, users can perform DDL on branches more consistently, improving branch management usability and aligning branch behavior with expected table operation semantics.

## Changes

- Add DDL support for branch operations
- Enable branch-related workflows to work with DDL statements
- Improve consistency of branch management behavior in Lance Spark

## Motivation

Branch operations are an important part of versioned data workflows. Supporting DDL on branches makes these workflows more complete and practical, especially for users who need to manage schema or table-related changes in isolated branches before merging.

### DIFF
--- a/docs/src/operations/ddl/create-branch.md
+++ b/docs/src/operations/ddl/create-branch.md
@@ -1,0 +1,101 @@
+# CREATE BRANCH
+
+Create a named branch for a Lance table.
+
+!!! warning "Spark Extension Required"
+    This feature requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+## Overview
+
+The `CREATE BRANCH` command creates a new branch that points to a specific table state. A branch can be created from:
+
+- the latest version of `main`
+- a specific version on `main`
+- the head of another branch
+- a specific version on another branch
+- a tag
+
+Creating a branch records a new reference in table metadata. It does not duplicate the table's data files.
+
+## Syntax
+
+=== "SQL"
+
+    ```sql
+    ALTER TABLE <table> CREATE BRANCH [IF NOT EXISTS] <branch_name>;
+
+    ALTER TABLE <table> CREATE BRANCH [IF NOT EXISTS] <branch_name>
+    AS OF VERSION <version>;
+
+    ALTER TABLE <table> CREATE BRANCH [IF NOT EXISTS] <branch_name>
+    AS OF BRANCH <source_branch>;
+
+    ALTER TABLE <table> CREATE BRANCH [IF NOT EXISTS] <branch_name>
+    AS OF BRANCH <source_branch> VERSION <version>;
+
+    ALTER TABLE <table> CREATE BRANCH [IF NOT EXISTS] <branch_name>
+    AS OF TAG <tag_name>;
+    ```
+
+If the `AS OF` clause is omitted, the new branch is created from the latest version of `main`.
+
+## Examples
+
+### Create a branch from the latest `main`
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE BRANCH feature_x;
+    ```
+
+### Create a branch from a specific `main` version
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE BRANCH IF NOT EXISTS snapshot_v5
+    AS OF VERSION 5;
+    ```
+
+### Create a branch from another branch head
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE BRANCH experiment_b
+    AS OF BRANCH experiment_a;
+    ```
+
+### Create a branch from a specific version on another branch
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE BRANCH experiment_b_v3
+    AS OF BRANCH experiment_a VERSION 3;
+    ```
+
+### Create a branch from a tag
+
+=== "SQL"
+    ```sql
+    ALTER TABLE lance.db.users CREATE BRANCH release_fix
+    AS OF TAG release_candidate;
+    ```
+
+## Output
+
+The `CREATE BRANCH` command returns:
+
+| Column | Type   | Description                |
+|--------|--------|----------------------------|
+| `name` | String | The name of the new branch |
+
+## Notes and Limitations
+
+- `CREATE BRANCH` is implemented as a Spark SQL extension command.
+- The referenced table must be a Lance table.
+- Creating a branch from a non-existent branch, tag, or version returns an error.
+- The branch name is returned even when the command result contains only a single row.
+
+## See Also
+
+- [SHOW BRANCHES](./show-branches.md)
+- [DROP BRANCH](./drop-branch.md)
--- a/docs/src/operations/ddl/show-branches.md
+++ b/docs/src/operations/ddl/show-branches.md
@@ -1,0 +1,62 @@
+# SHOW BRANCHES
+
+List branches defined on a Lance table.
+
+!!! warning "Spark Extension Required"
+    This feature requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+## Overview
+
+The `SHOW BRANCHES` command returns one row for each branch on a Lance table. It is useful for inspecting branch lineage, verifying branch creation, and understanding which version each branch points to.
+
+## Syntax
+
+=== "SQL"
+    ```sql
+    SHOW BRANCHES FROM <table>;
+    SHOW BRANCHES IN <table>;
+    SHOW BRANCH FROM <table>;
+    SHOW BRANCH IN <table>;
+    ```
+
+Both plural and singular forms are accepted. Both `FROM` and `IN` are accepted.
+
+## Examples
+
+### List all branches on a table
+
+=== "SQL"
+    ```sql
+    SHOW BRANCHES FROM lance.db.users;
+    ```
+
+### Use the singular alias
+
+=== "SQL"
+    ```sql
+    SHOW BRANCH IN lance.db.users;
+    ```
+
+## Output
+
+The `SHOW BRANCHES` command returns the following columns:
+
+| Column | Type | Description                                                                      |
+|--------|------|----------------------------------------------------------------------------------|
+| `name` | String | Branch name                                                                      |
+| `parent_branch` | String | Source branch name if the branch was created from another branch; otherwise NULL |
+| `parent_version` | Long | Version used as the branch starting point                                        |
+| `created_at` | Long | Branch creation timestamp as returned by Lance metadata                          |
+| `manifest_size` | Integer | Manifest size recorded for the branch                                            |
+
+## Notes
+
+- `SHOW BRANCHES` is implemented as a Spark SQL extension command.
+- The target table must be a Lance table.
+- The result includes one row per branch currently registered in the table metadata.
+- Use [CREATE BRANCH](./create-branch.md) to add a branch and [DROP BRANCH](./drop-branch.md) to remove one.
+
+## See Also
+
+- [CREATE BRANCH](./create-branch.md)
+- [DROP BRANCH](./drop-branch.md)
--- a/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -149,7 +149,7 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
     }
   }
 
-  def _parseVersion(ctx: ParserRuleContext, value: String): Long = {
+  private def _parseVersion(ctx: ParserRuleContext, value: String): Long = {
     try {
       java.lang.Long.valueOf(value)
     } catch {
--- a/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -15,8 +15,8 @@ package org.apache.spark.sql.catalyst.parser.extensions
 
 import org.antlr.v4.runtime.ParserRuleContext
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedIdentifier, UnresolvedRelation}
-import org.apache.spark.sql.catalyst.parser.ParserInterface
-import org.apache.spark.sql.catalyst.plans.logical.{AddColumnsBackfill, AddIndex, LanceDropIndex, LanceNamedArgument, LogicalPlan, Optimize, SetUnenforcedPrimaryKey, ShowIndexes, UpdateColumnsBackfill, Vacuum}
+import org.apache.spark.sql.catalyst.parser.{ParseException, ParserInterface}
+import org.apache.spark.sql.catalyst.plans.logical.{AddColumnsBackfill, AddIndex, LanceCreateBranch, LanceDropBranch, LanceDropIndex, LanceNamedArgument, LanceShowBranches, LogicalPlan, Optimize, SetUnenforcedPrimaryKey, ShowIndexes, UpdateColumnsBackfill, Vacuum}
 import org.lance.spark.utils.{FieldPathUtils, ParserUtils}
 
 import scala.collection.JavaConverters._
--- a/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
+++ b/lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
@@ -13,6 +13,7 @@
  */
 package org.apache.spark.sql.catalyst.parser.extensions
 
+import org.antlr.v4.runtime.ParserRuleContext
 import org.apache.spark.sql.catalyst.analysis.{UnresolvedIdentifier, UnresolvedRelation}
 import org.apache.spark.sql.catalyst.parser.ParserInterface
 import org.apache.spark.sql.catalyst.plans.logical.{AddColumnsBackfill, AddIndex, LanceDropIndex, LanceNamedArgument, LogicalPlan, Optimize, SetUnenforcedPrimaryKey, ShowIndexes, UpdateColumnsBackfill, Vacuum}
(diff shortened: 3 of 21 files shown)


LuciferYang:
will review this one later

fangbo:
> will review this one later

Thanks very much.

I have an idea about grammar definitions that I'd like to discuss. There is another grammar definition like:

```sql
-- create branch from main's specific version
ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF MAIN <main_version>

-- create branch from branch's specific version
ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF BRANCH <source_branch_name>  <branch_version>

-- create branch from tag
ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF TAG <source_tag_name> 
```

What do you think about these two choices ?

cc @hamersaw

lance-spark-base_2.12/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensions.g4
```diff
@@ -66,8 +74,15 @@ number
     | MINUS? DOUBLE_LITERAL             #doubleLiteral
     ;
 
+versionNumber
+    : BIGINT_LITERAL
+    ;
+
 ADD: 'ADD';
 ALTER: 'ALTER';
+AS: 'AS';
```
LuciferYang: Maybe  we should introduce a `nonReserved` rule folded into `identifier` (the pattern Spark's own grammar uses) so these words stay usable as identifiers where they aren't structurally a keyword. This is a legacy issue and can be refined in a standalone PR.

LuciferYang:
> > will review this one later
> 
> Thanks very much.
> 
> I have an idea about grammar definitions that I'd like to discuss. There is another grammar definition like:
> 
> ```sql
> -- create branch from main's specific version
> ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF MAIN <main_version>
> 
> -- create branch from branch's specific version
> ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF BRANCH <source_branch_name>  <branch_version>
> 
> -- create branch from tag
> ALTER TABLE <table_name> CREATE BRANCH <branch_name> VERSION AS OF TAG <source_tag_name> 
> ```
> 
> What do you think about these two choices ?
> 
> cc @hamersaw

Thanks @fangbo, the keyword form is a lot better than `REF/MAIN/5` — the `/` separators aren't really SQL, and this grammar becomes a public contract once it ships, so worth settling now.

Since we already follow Iceberg's extension model closely, the cleanest option is probably to match its branch DDL where we can:

```sql
ALTER TABLE t CREATE BRANCH [IF NOT EXISTS] <name> [AS OF VERSION <id>] [RETAIN <n> DAYS]
ALTER TABLE t DROP BRANCH [IF EXISTS] <name>
```

The catch is that Iceberg has one global snapshot space, so `AS OF VERSION <id>` is enough. Our versions are per-ref, so we still have to say which ref the version belongs to. So something like:

```sql
ALTER TABLE t CREATE BRANCH [IF NOT EXISTS] <name>
   [ AS OF VERSION <v>                 -- main @ v
   | AS OF BRANCH <src> [VERSION <v>]
   | AS OF TAG <tag> ]
```

That keeps Iceberg's `AS OF VERSION <v>` verbatim for the common case and only adds `AS OF BRANCH` / `AS OF TAG` where our model actually needs it. Two small things: `AS OF` reads better first (matches Iceberg's `CREATE BRANCH … AS OF VERSION`), and `AS OF TAG <tag>` is cleaner than `VERSION AS OF TAG` since a tag isn't a version.

Couple of cheap wins worth folding in now, both already in Iceberg: `IF NOT EXISTS` / `CREATE OR REPLACE`, and — more important — declaring the new keywords as `nonReserved` so `branch`/`tag`/`version`/`as`/`of` don't become reserved inside every extension statement (today a column named `version` would fail to parse).

Iceberg also skips `SHOW BRANCHES` and just exposes a `<table>.refs` metadata table — not a blocker, just an option if we'd rather keep it queryable.

lance-spark-base_2.12/src/test/java/org/lance/spark/branch/BaseBranchDDLTest.java
```diff
@@ -0,0 +1,303 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+/** Base tests for BRANCH DDL commands. */
+public abstract class BaseBranchDDLTest {
```
LuciferYang, line 38: I think we should add the negative-path integration tests for the error cases the docs guarantee, a backtick-quoted branch-name test (locks in `cleanIdentifier`), and a few `LanceSqlExtensionsAstBuilderTest` cases for the new visit methods.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/CreateBranchExec.scala
```diff
@@ -0,0 +1,53 @@
...
+
+  override def output: Seq[Attribute] = CreateBranchOutputType.SCHEMA
+
+  override protected def run(): Seq[InternalRow] = {
+    val lanceDataset = catalog.loadTable(ident) match {
+      case d: LanceDataset => d
+      case _ => throw new UnsupportedOperationException("CreateBranch only supports LanceDataset")
+    }
+
+    val dataset = Utils.openDatasetBuilder(lanceDataset.readOptions()).build()
```
LuciferYang: The branch execs look like they'll miss credential-vended catalogs. `CreateBranchExec` (and `DropBranchExec` / `ShowBranchesExec`) open with a bare `Utils.openDatasetBuilder(readOptions).build()`, whereas `OptimizeExec` / `AddIndexExec` thread the catalog's vended options into the open:

```scala
// OptimizeExec
val initialStorageOpts = catalog match {
  case ns: BaseLanceNamespaceSparkCatalog =>
    Option(lanceDataset.getInitialStorageOptions).map(_.asScala.toMap)
  case _ => None
}
val dataset = Utils.openDatasetBuilder(readOptions)
  .initialStorageOptions(initialStorageOpts.map(_.asJava).orNull)
  .build()
```

So on a credential-vending namespace catalog over S3/GCS, the open (and the branch write) won't have the vended creds. It passes today because the dir catalog's options are non-empty and the Glue/S3 docker test is skipped, so CI never hits this path.

Related sharp edge: `CreateBranchExec` calls the 3-arg `createBranch(name, ref, dataset.getLatestStorageOptions)`. `getLatestStorageOptions` is the manifest map (no creds), and the 3-arg overload throws on a null/empty map (`checkArgument(opts != null && !opts.isEmpty(), ...)`). The 2-arg `createBranch(name, ref)` has no such precondition.

So I'd thread `initialStorageOptions` into the open like above, then use the 2-arg `createBranch(name, ref)` — the write picks up creds from the open.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/CreateBranch.scala
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.spark.sql.catalyst.plans.logical
+
+import org.apache.spark.sql.catalyst.expressions.{Attribute, AttributeReference}
+import org.apache.spark.sql.types.{DataTypes, StructField, StructType}
+import org.lance.Ref
+
+case class CreateBranch(
```
LuciferYang: Using `LanceCreateBranch` could proactively prevent potential naming conflicts.

lance-spark-4.0_2.13/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
```diff
@@ -105,6 +105,50 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
...
 
+  override def visitCreateBranchRefMain(ctx: LanceSqlExtensionsParser.CreateBranchRefMainContext)
+      : CreateBranch = {
+    val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
+    val branchName = cleanIdentifier(ctx.branchName.getText)
+    if (ctx.refMainVersion == null) CreateBranch(table, branchName, org.lance.Ref.ofMain())
+    else CreateBranch(
+      table,
+      branchName,
+      org.lance.Ref.ofMain(java.lang.Long.valueOf(ctx.refMainVersion.getText)))
```
LuciferYang: There is a risk of out-of-bounds overflow. It is recommended to catch `NumberFormatException` and wrap it into a `ParseException`.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/CreateBranchExec.scala
```diff
@@ -0,0 +1,53 @@
...
+    try {
+      val branchDs = dataset.createBranch(
+        branchName,
+        ref,
+        dataset.getLatestStorageOptions)
+      branchDs.close()
```
LuciferYang: `branchDs.close()` sits in the `try` body, not protected by its own `finally`.

docs/src/operations/ddl/show-branches.md
```diff
@@ -0,0 +1,62 @@
...
+    ```
+
+## Output
+
+The `SHOW BRANCHES` command returns the following columns:
+
+| Column | Type | Description |
+|--------|------|-------------|
+| `name` | String | Branch name |
+| `parent_branch` | String | Source branch name if the branch was created from another branch; otherwise empty |
```
LuciferYang: otherwise empty -> otherwise `NULL` ?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/ShowBranches.scala
```diff
@@ -0,0 +1,44 @@
...
+  }
+}
+
+object ShowBranchesOutputType {
+  val SCHEMA = StructType(
+    Array(
+      StructField("name", DataTypes.StringType, nullable = false),
+      StructField("parent_branch", DataTypes.StringType, nullable = true),
+      StructField("parent_version", DataTypes.LongType, nullable = false),
+      StructField("create_at", DataTypes.LongType, nullable = false),
```
LuciferYang: created_at

lance-spark-base_2.12/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensions.g4
```diff
@@ -133,6 +153,4 @@ fragment DIGIT
 
 fragment LETTER
     : [A-Z]
-    ;
-
-
+    ;
```
LuciferYang: should restore the trailing newline

fangbo:
@LuciferYang Greatly thanks for your detailed suggestion. I have made some modification according to your advice.

hamersaw:
Thanks for the PR! This is great. Overall looks great, my only comment is the main reason I never wrapped up the initial proposal for this - basically I didn't have time to drive consensus. There was quite a bit of conversation on https://github.com/lance-format/lance-spark/pull/198 about how we wanted to handle branches / tags / versions in this connector, the two competing approaches:

1. New keywords: Use new `AS OF BRANCH` / `AS OF TAG`, etc. For example:
```
SELECT * FROM foo AS OF BRANCH 'b0';
ALTER TABLE foo CREATE BRANCH 'b0' AS OF VERSION 12345;
```

2. Bake the references directly into the table URI
```
SELECT * FROM foo/__ref/b0;
ALTER TABLE foo/__ref/12345 CREATE BRANCH 'b0';
```

The former is what is proposed in this PR, but IIR we had loose consensus that the latter was a little cleaner?! Also the latter means we can have a single style that is used across connectors (Spark / Presto / Trino / DuckDB) with the base parsing code in lance core repo.

fangbo:
@hamersaw Thanks for your feedback. 

Branch/Tag operations involve DDL and DML.

## DDL

This PR focuses on branch DDL. Branch/Tag is part of the Dataset, so from a Branch/Tag management perspective, it's more intuitive for the Dataset to be the target of ALTER operations. Therefore, I lean more towards:

```
ALTER TABLE foo CREATE BRANCH 'b0' AS OF VERSION 12345;
```

## DML

As for DML, I believe one consensus reached in the previous [PR#198](https://github.com/lance-format/lance-spark/pull/198) was to treat Branch/Tag as ordinary tables. For example::

```
SELECT * FROM <table_name>__branch/<branch_name>
UPDATE <table_name>__branch/<branch_name> SET ...
DELETE <table_name>__branch/<branch_name> WHERE
```

or

```
SELECT * FROM <table_name>__branch__<branch_name>
UPDATE <table_name>__branch__<branch_name> SET ...
DELETE <table_name>__branch__<branch_name> WHERE
```

>  By the way, I lean more towards `<table_name>__branch__<branch_name>`, because it avoids the extra '/' symbol.

fangbo:
cc @majin1102

fangbo:
@LuciferYang @hamersaw Do you have some more suggestion about this PR ?

lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
```diff
@@ -105,6 +106,70 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
...
+        ifNotExists)
+    }
+  }
+
+  def _parseVersion(ctx: ParserRuleContext, value: String): Long = {
+    try {
+      java.lang.Long.valueOf(value)
+    } catch {
+      case _: NumberFormatException =>
+        throw new ParseException("Can't parse value:" + value + " to version", ctx)
```
LuciferYang: On **every supported Spark version**, the `ParseException(String, ParserRuleContext)` constructor's first parameter is an **error class**, not a message — it delegates to `SparkThrowableHelper.getMessage(errorClass, Map.empty)`.

lance-spark-4.1_2.13/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
```diff
@@ -105,6 +105,57 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
...
+      : LanceCreateBranch = {
+    val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
+    val branchName = cleanIdentifier(ctx.branchName.getText)
+    val ifNotExists = ctx.EXISTS() != null
+    if (ctx.refMainVersion == null)
+      LanceCreateBranch(table, branchName, org.lance.Ref.ofMain(), ifNotExists)
+    else LanceCreateBranch(
+      table,
+      branchName,
+      org.lance.Ref.ofMain(java.lang.Long.valueOf(ctx.refMainVersion.getText)),
```
LuciferYang: use `org.lance.Ref.ofMain(_parseVersion(ctx, ctx.refMainVersion.getText))`?

lance-spark-4.1_2.13/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
```diff
@@ -105,6 +105,57 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
...
+    val table = UnresolvedIdentifier(visitMultipartIdentifier(ctx.multipartIdentifier()))
+    val branchName = cleanIdentifier(ctx.branchName.getText)
+    val refBranchName = cleanIdentifier(ctx.refBranchName.getText)
+    val ifNotExists = ctx.EXISTS() != null
+    if (ctx.refBranchVersion == null)
+      LanceCreateBranch(table, branchName, org.lance.Ref.ofBranch(refBranchName), ifNotExists)
+    else LanceCreateBranch(
+      table,
+      branchName,
+      org.lance.Ref.ofBranch(refBranchName, java.lang.Long.valueOf(ctx.refBranchVersion.getText)),
```
LuciferYang: ditto

lance-spark-base_2.12/src/main/antlr4/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensions.g4
```diff
@@ -66,25 +74,39 @@ number
...
 CREATE: 'CREATE';
 DROP: 'DROP';
+EXISTS: 'EXISTS';
 FROM: 'FROM';
+IF: 'IF';
 IN: 'IN';
 INDEX: 'INDEX';
 INDEXES: 'INDEXES';
 KEY: 'KEY';
+MAIN: 'MAIN';
```
LuciferYang: `MAIN: 'MAIN';` is declared but no parser rule references it.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceCreateBranchExec.scala
```diff
@@ -0,0 +1,64 @@
...
+ * limitations under the License.
+ */
+package org.apache.spark.sql.execution.datasources.v2
+
+import org.apache.spark.sql.catalyst.InternalRow
+import org.apache.spark.sql.catalyst.expressions.{Attribute, GenericInternalRow}
+import org.apache.spark.sql.catalyst.plans.logical.LanceCreateBranchOutputType
+import org.apache.spark.sql.connector.catalog.{Identifier, TableCatalog}
+import org.apache.spark.unsafe.types.UTF8String
+import org.lance.spark.{BaseLanceNamespaceSparkCatalog, LanceDataset}
```
LuciferYang: BaseLanceNamespaceSparkCatalog seems unused import

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceCreateBranchExec.scala
```diff
@@ -0,0 +1,64 @@
...
+    val dataset = Utils.openDatasetBuilder(lanceDataset.readOptions())
+      .initialStorageOptions(lanceDataset.getInitialStorageOptions)
+      .build()
+
+    try {
+      val alreadyExists = dataset.branches().list().asScala.exists(_.getName == branchName)
+      if (!ifNotExists || !alreadyExists) {
+        var branchDs: org.lance.Dataset = null
+        try {
+          branchDs = dataset.createBranch(branchName, ref, dataset.getInitialStorageOptions)
```
LuciferYang:
```scala
val opts = dataset.getInitialStorageOptions
branchDs =
  if (opts == null || opts.isEmpty) dataset.createBranch(branchName, ref)
  else dataset.createBranch(branchName, ref, opts)
```

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceCreateBranchExec.scala
```diff
@@ -0,0 +1,64 @@
...
+      case d: LanceDataset => d
+      case _ => throw new UnsupportedOperationException("CreateBranch only supports LanceDataset")
+    }
+
+    val dataset = Utils.openDatasetBuilder(lanceDataset.readOptions())
+      .initialStorageOptions(lanceDataset.getInitialStorageOptions)
+      .build()
+
+    try {
+      val alreadyExists = dataset.branches().list().asScala.exists(_.getName == branchName)
```
LuciferYang, line 46: Both `LanceCreateBranchExec` and `LanceDropBranchExec` do `branches().list()` then create/delete. A concurrent `CREATE BRANCH` between the two calls still surfaces the native "already exists" error despite `IF NOT EXISTS`. For DDL this is acceptable, but catching the native already-exists/not-found error instead would be atomic and saves one branch-listing call per command (object-store IO against the table root — a `read_dir` over the refs path). Also: **both** execs compute the existence check eagerly (`LanceCreateBranchExec.scala:46`, `LanceDropBranchExec.scala:45`), so the `list()` call is issued even when the `IF [NOT] EXISTS` flag is absent and its result goes unused.

lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
```diff
@@ -105,6 +106,70 @@ class LanceSqlExtensionsAstBuilder(delegate: ParserInterface)
...
+    } else {
+      LanceCreateBranch(
+        table,
+        branchName,
+        org.lance.Ref.ofBranch(refBranchName, _parseVersion(ctx, ctx.refBranchVersion.getText)),
+        ifNotExists)
+    }
+  }
+
+  def _parseVersion(ctx: ParserRuleContext, value: String): Long = {
```
LuciferYang: Public `def _parseVersion` with a leading underscore is unidiomatic Scala

docs/src/operations/ddl/create-branch.md
```diff
@@ -0,0 +1,101 @@
...
+
+| Column | Type   | Description                |
+|--------|--------|----------------------------|
+| `name` | String | The name of the new branch |
+
+## Notes and Limitations
+
+- `CREATE BRANCH` is implemented as a Spark SQL extension command.
+- The referenced table must be a Lance table.
+- Creating a branch from a non-existent branch, tag, or version returns an error.
```
LuciferYang, line 95: Lacks corresponding test cases

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceCreateBranchExec.scala
fangbo: Exception throws when creating a existed branch:

```
java.lang.RuntimeException: Encountered internal error. Please file a bug report at https://github.com/lance-format/lance/issues. Clone operation should not enter build_manifest., /Users/runner/work/lance/lance/rust/lance/src/dataset/transaction.rs:1866:28

	at org.lance.Dataset.nativeCreateBranch(Native Method)
	at org.lance.Dataset.innerCreateBranch(Dataset.java:1761)
	at org.lance.Dataset.createBranch(Dataset.java:1753)
``` 

This means that existence can not be checked from the error. I think it is reasonable to list branches to check the branch exist or not.

docs/src/operations/ddl/create-branch.md
fangbo: BaseBranchDDLTest's three corresponding test cases are added.

- testCreateBranchFailsWhenSourceBranchDoesNotExist
- testCreateBranchFailsWhenSourceTagDoesNotExist
- testCreateBranchFailsWhenSourceVersionDoesNotExist

fangbo:
@LuciferYang Thanks for your more detailed reviews. I have made some modification. Could you please look at it ?

lance-spark-3.5_2.12/src/main/scala/org/apache/spark/sql/catalyst/parser/extensions/LanceSqlExtensionsAstBuilder.scala
fangbo: This method is changed to private

LuciferYang (approved):
LGTM

PR #314, opened 2026-03-19
feat: support external existing table by using location

summaryzb:
This PR introduces support for creating external Lance tables by specifying a `LOCATION` property during table creation. This feature decouples the creation logic for managed tables and external tables. 

Key changes include:
- **External Table Creation:** If a `LOCATION` is provided in the table properties, the system will now  register the table metadata directly in the namespace via `RegisterTableRequest`, rather than creating a new dataset from scratch.
- **Managed Table Fallback:** If no `LOCATION` is provided, the system falls back to the existing behavior of creating a managed table (`createManagedTable`).
- **Schema:** The table schema is provided by the actual meta data in location, other than the definition of create table statement

PR #579, opened 2026-06-03
feat: add LOCATION support to CREATE TABLE for custom paths and existing datasets

### LINKED ISSUE #158: The location of a spark table does not take effect when creating a spark table
When I try to create a Lance table via the following code

```
spark = SparkSession.builder \
    .appName("gcs_fielset_test") \
    .config("spark.sql.catalog.lance", "org.lance.spark.LanceNamespaceSparkCatalog") \
    .config("spark.sql.catalog.lance.impl", "dir") \
    .config("spark.executor.memory", "3g") \
    .config("spark.sql.catalog.lance.parent", "lance_catalog") \
    .config("spark.sql.defaultCatalog", "lance") \
    .getOrCreate()  

results = spark.sql("""
create table schema.t23(id int, score float) USING lance LOCATION '/tmp/schema/t23.lance/' TBLPROPERTIES  ('format' = 'lance')
""")
results.show()
```

The location `/tmp/schema/t23.lance/` can't take effect. 

Image

ykozxy:
## Summary
Support `CREATE TABLE ... LOCATION 'path'` for namespace-based catalogs (e.g., REST catalogs backed by Gravitino or other Lance namespace servers).
| `LOCATION` provided? | Data exists? | Behavior |
|---|---|---|
| No | N/A | Managed table at server-assigned path (unchanged) |
| Yes | No | Create new empty dataset at the custom path |
| Yes | Yes | Register existing dataset (schema read from disk) |

Closes #158

## Motivation
Users need to create tables at arbitrary storage paths through a centralized catalog — for example, registering existing Lance datasets or placing new tables in specific storage accounts. Without this fix, LOCATION is silently ignored and all tables go to the server-assigned default path. This follows standard Spark/Iceberg semantics where LOCATION controls the physical storage location of a table's data.

## Changes
All changes are in `BaseLanceNamespaceSparkCatalog.java`:
- **`createTable()`**: Extract LOCATION from properties before `copyUserTableProperties()` strips it. If data exists at the path, register via `RegisterTableRequest`. Otherwise, pass location to `DeclareTableRequest.setLocation()` and create a new empty dataset there. Includes cleanup (deregister) on failure to avoid dangling catalog entries.
- **`stageCreate()`**: Pass LOCATION to `DeclareTableRequest` when present.
- **`stageCreateOrReplace()`**: Same, in the `!exists` branch.
- **`lanceDatasetExists()`**: New helper — probes whether a Lance dataset exists at a given path.
- **`registerExistingTable()`**: New helper — registers existing data via `RegisterTableRequest` with cleanup on failure.

## Design Decisions
- **LOCATION is extracted before `copyUserTableProperties()`** strips it — this preserves the existing behavior where LOCATION is not persisted as a table property, while still using it to control storage placement.
- **Existing data detection**: When LOCATION points to an existing Lance dataset, we register it (similar to #314) rather than failing. This enables the common use case of cataloging pre-existing datasets.
- **Cleanup on failure**: Both the declare-new and register-existing paths clean up via `deregisterTable` if downstream operations fail, preventing dangling catalog entries.
- **Table properties are not persisted for registered external tables** — the dataset already exists with its own configuration.

## Related
- #158 — LOCATION silently ignored (this PR closes it)
- #314 — External table registration via LOCATION (handles existing data only; this PR handles both new + existing)
- apache/gravitino#9520 — Gravitino server-side support for custom locations (already resolved)

## Test plan
- [x] 362 existing unit tests pass (0 failures)
- [x] Lint + format pass (`make lint`, `make format`)
- [x] Integration test: CREATE TABLE without LOCATION — managed table works as before
- [x] Integration test: CREATE TABLE with LOCATION on empty path — new dataset created at custom path
- [x] Integration test: CREATE TABLE with LOCATION on existing data — registers and reads data
- [x] Integration test: Normal Spark operations unaffected after LOCATION usage
- All integration tests run against a live Gravitino Lance REST server

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -606,14 +606,8 @@ public abstract class BaseLanceNamespaceSparkCatalog
       Map<String, String> tableProperties = copyUserTableProperties(properties);
 
       if (lanceDatasetExists(userLocation)) {
-        // Register existing dataset
         return registerExistingTable(
-            userLocation,
-            tableIdList,
-            processedSchema,
-            tableProperties,
-            fileFormatVersion,
-            shardingSpec);
+            userLocation, tableIdList, processedSchema, tableProperties, null, shardingSpec);
       }
 
       // Create new table at custom location
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -32,6 +32,7 @@ import org.lance.namespace.model.DropNamespaceRequest;
 import org.lance.namespace.model.DropTableRequest;
 import org.lance.namespace.model.ListTablesRequest;
 import org.lance.namespace.model.ListTablesResponse;
+import org.lance.namespace.model.RegisterTableRequest;
 import org.lance.namespace.model.RenameTableRequest;
 import org.lance.operation.UpdateConfig;
 import org.lance.operation.UpdateMap;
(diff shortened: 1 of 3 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -714,6 +805,75 @@ private Table createTableAtPath(
...
+  /** Probe whether a Lance dataset already exists at the given location. */
+  private boolean lanceDatasetExists(String location) {
+    try {
+      LanceSparkReadOptions probeOptions =
+          createReadOptions(
+              location, catalogConfig, Optional.empty(), Optional.empty(), Optional.empty(), name);
+      try (Dataset ds = Utils.openDatasetBuilder(probeOptions).build()) {
+        return true;
+      }
+    } catch (Exception e) {
```
geruh: Can we narrow this exception catch? We could get possibly get false positives with auth or IO. 

I did some investigation and it looks like the JNI maps exceptions like `DatasetNotFound`, `DatasetAlreadyExists` to one `IllegalArgumentException` Looks like a good reason to surface error mapping.

https://github.com/lance-format/lance/blob/main/java/lance-jni/src/error.rs#L185

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -893,12 +1053,17 @@ public StagedTable stageCreate(
           "Namespace not configured. Use 'impl' config for namespace-based access.");
     }
 
+    String userLocation = properties.get(CREATE_TABLE_PROPERTY_LOCATION);
```
geruh, line 1064: For `CTAS & writeTo()`, I’d keep staged create as declare and write only, and keep register-existing on `createTable`. I think this is what we are doing here

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -714,6 +805,75 @@ private Table createTableAtPath(
...
+          schema,
+          initialStorageOptions,
+          namespaceImpl,
+          namespaceProperties,
+          managedVersioning,
+          fileFormatVersion,
+          tableProperties,
+          shardingSpec);
+    } catch (Exception e) {
+      DeregisterTableRequest deregisterRequest = new DeregisterTableRequest();
```
geruh: nit: this logic appears to show up twice we can refactor into a shared private rollback method like:

```java
private void deregisterQuietly(List<String> tableIdList) {
  try {
    DeregisterTableRequest request = new DeregisterTableRequest();
    tableIdList.forEach(request::addIdItem);
    namespace.deregisterTable(request);
  } catch (Exception cleanupEx) {
    logger.warn("Failed to deregister table during rollback", cleanupEx);
  }
}
```

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
ykozxy: Good point! I narrowed to `IllegalArgumentException` in the catch clause.

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
ykozxy: Good suggestion. I extracted these to a shared helper

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
ykozxy: Right, register existing path is only in `createTable`, and `CTAS & writeTo()` always pass LOCATION and write new data.

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -594,6 +595,90 @@ public Table createTable(
...
     }
 
+    // Extract user-specified LOCATION before properties are filtered
+    String userLocation = properties.get(CREATE_TABLE_PROPERTY_LOCATION);
+
+    if (userLocation != null && !userLocation.trim().isEmpty()) {
+      Identifier actualIdent = transformIdentifierForApi(ident);
+      List<String> tableIdList = buildTableId(actualIdent);
+      StructType processedSchema = SchemaConverter.processSchemaWithProperties(schema, properties);
+      String fileFormatVersion = catalogConfig.getFileFormatVersion(properties);
```
summaryzb: `fileFormatVersion` of existing table is already set inside, not determined by the properties

ykozxy: Good point. Changed to pass `null` for `fileFormatVersion` when registering existing table

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -594,14 +595,90 @@ public Table createTable(
           "Namespace not configured. Use 'impl' config for namespace-based access.");
     }
 
+    // Extract user-specified LOCATION before properties are filtered
+    String userLocation = properties.get(CREATE_TABLE_PROPERTY_LOCATION);
```
geruh, line 599: non blocking: would be nice to move this to a helper since create is so many loc.

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseTestLocationSupport.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * <p>Custom LOCATION for new tables requires a REST namespace backend (e.g., Gravitino) that
+ * supports DeclareTableRequest.setLocation(). The DirectoryNamespace backend enforces server-
+ * assigned paths and rejects custom locations. Full LOCATION tests run as integration tests against
+ * a live REST namespace server.
+ */
+public abstract class BaseTestLocationSupport extends BaseTestSparkDirectoryNamespace {
+
+  @Test
+  public void testCreateTableWithoutLocationUnchanged() {
```
geruh, line 42: since the scope here is also adding external/register table support it would be nice to add a test here

geruh (approved):
LGTM, have some minor comments could be done in a follow up

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
ykozxy: Done! Moved the registration to a helper function

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseTestLocationSupport.java
ykozxy: Added more tests to cover table creation cases

ykozxy:
> LGTM, have some minor comments could be done in a follow up

Thanks for the review! Addressed both followups in the latest commit

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
summaryzb: Pass the existing `fileFormatVersion` maybe better, i'm not sure that passing null for fileFormatVersion when registering existing table is equivalent

summaryzb (approved):

geruh:
Thanks for the PR @ykozxy, and thanks for the review @summaryzb!

PR #587, opened 2026-06-06
feat: add namespace insert write path

jackye1995:
Adds an opt-in namespace insert append path for namespace-backed Lance tables.

Expected user experience:
- Existing INSERT INTO and append writes keep using the default Spark writer unless users set `use_namespace_insert=true`.
- With namespace insert enabled, users still submit a normal DataFrame append; Spark runs executor-side writer tasks that send Arrow batches through the Lance namespace insert API.
- `namespace_insert_parallelism` lets users request the number of writer tasks; sharded tables use the sharding distribution, and unsharded tables repartition by the first output column.
- Directory and REST namespaces use the same API path, so REST namespaces can handle ingestion without changing Spark user code.
- Insert requests commit as they run, so users should use the default writer when they need Spark driver-side atomic commit behavior.

Includes local/rest-dir Docker CI coverage and documentation updates.

### DIFF
--- a/.github/workflows/spark-namespace-insert.yml
+++ b/.github/workflows/spark-namespace-insert.yml
@@ -1,0 +1,178 @@
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Spark Namespace Insert Docker
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - ready_for_review
+      - reopened
+    paths:
+      - ".github/workflows/spark-namespace-insert.yml"
+      - "Makefile"
+      - "docker/**"
+      - "integration-tests/**"
+      - "lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java"
+      - "lance-spark-base_2.12/src/main/java/org/lance/spark/write/**"
+      - "lance-spark-base_2.12/src/test/java/org/lance/spark/LanceSparkWriteOptionsTest.java"
+      - "pom.xml"
+      - "*/pom.xml"
+  workflow_dispatch:
+    inputs:
+      spark-version:
+        description: "Spark version to test"
+        required: true
+        default: "3.5"
+      scala-version:
+        description: "Scala version to test"
+        required: true
+        default: "2.13"
+      backends:
+        description: "Comma-separated test backends: local or local,rest-dir"
+        required: true
+        default: "local,rest-dir"
+      rest-uri:
+        description: "Optional REST namespace URI. If omitted, tests start a local REST directory namespace."
+        required: false
+        default: ""
+      rest-database:
+        description: "Optional database header value for an external REST namespace"
+        required: false
+        default: ""
+      docker-run-args:
+        description: "Extra docker run args for docker-test"
+        required: false
+        default: ""
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+env:
+  SPARK_VERSION: ${{ github.event.inputs['spark-version'] || '3.5' }}
+  SCALA_VERSION: ${{ github.event.inputs['scala-version'] || '2.13' }}
+  NAMESPACE_INSERT_TEST_BACKENDS: ${{ github.event_name == 'workflow_dispatch' && github.event.inputs.backends || 'local,rest-dir' }}
+  NAMESPACE_INSERT_PYTEST_CMD: >-
+    pytest /home/lance/tests/test_lance_spark.py::TestDMLNamespaceInsert
+    -v --timeout=180
+
+jobs:
+  namespace-insert-docker-test:
+    name: Namespace Insert Docker Test
+    runs-on: ubuntu-24.04
+    timeout-minutes: 90
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+      - name: Set up Java
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+          cache: "maven"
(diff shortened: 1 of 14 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/write/NamespaceInsertBatchWrite.java
```diff
@@ -0,0 +1,481 @@
...
+      if (rowCount == 0) {
+        return;
+      }
+
+      arrowWriter.finish();
+      root.setRowCount(rowCount);
+      byte[] requestData = serializeCurrentBatch();
+      InsertIntoTableRequest request = new InsertIntoTableRequest().id(tableId).mode("append");
+      try {
+        namespace.insertIntoTable(request, requestData);
```
LuciferYang, line 299: This path commits visible data directly from executor tasks via `namespace.insertIntoTable(...)`. If a task inserts one or more batches and Spark later retries or speculatively reruns the same partition because a later batch, executor loss, or close fails, the replacement attempt can insert the same rows again and the Spark job can still succeed with duplicates. Please do not describe the current path as safe or atomic. Either add first-class namespace idempotency with a stable cross-attempt key such as `write_id + partition_id + batch_ordinal`, gated by namespace capability/version discovery, or move visible writes behind a staged driver commit. Disabling retries/speculation only reduces exposure; it is not a correctness fix.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/NamespaceInsertBatchWrite.java
```diff
@@ -0,0 +1,481 @@
...
+      } catch (RuntimeException e) {
+        throw new IOException("Failed to insert rows through Lance namespace", e);
+      }
+
+      rowsInserted += rowCount;
+      insertRequests++;
+      hasRowsInCurrentBatch = false;
+      allocateBatch();
```
LuciferYang, line 307: After a successful insert, `flush()` allocates the next Arrow batch, and the final `commit()` flush uses the same path. If the last insert has already succeeded but a later `allocateBatch()` or `close()` cleanup fails, Spark treats the task as failed and may retry data that was already inserted. Please split normal flush from final flush so the last successful insert does not allocate another batch. Once any insert has happened, cleanup of the sharding evaluator, Arrow batch, blob resolver, and namespace should be best-effort: log failures but do not throw them back to Spark. Close failures before any insert can still fail the task.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/NamespaceInsertBatchWrite.java
```diff
@@ -0,0 +1,481 @@
...
+      if (batchAllocator != null) {
+        batchAllocator.close();
+        batchAllocator = null;
+      }
+    }
+
+    @Override
+    public WriterCommitMessage commit() throws IOException {
+      if (shardingKeyEvaluator != null) {
+        shardingKeyEvaluator.flush(this::writePartitionedRow);
```
LuciferYang, line 349: This new namespace insert writer calls `shardingKeyEvaluator.flush(this::writePartitionedRow)` in `commit()`. Inside `ShardingBatchKeyEvaluator.flush(...)`, it calls the row consumer in the loop, and that consumer may trigger namespace inserts. The no-extra-batch validation and the reset/allocation in `finally` happen after the consumer has run. If some rows have already been inserted and a later validation or reset fails, Spark can retry and duplicate those rows. Please finish evaluator validation and key extraction before invoking the consumer, including the no-extra-batch check. After the consumer may have forced namespace inserts, only cleanup/close failures should be best-effort.

integration-tests/test_lance_spark.py
```diff
@@ -1989,6 +2004,91 @@ def test_insert_append_data(self, spark):
         assert count == 4
 
 
+@pytest.mark.rest_dir_compatible
+class TestDMLNamespaceInsert:
```
LuciferYang, line 2008: `TestDMLNamespaceInsert` currently covers basic append and fixed-size-list vector writes, but not the blob-reference resolver path that namespace insert depends on. Please add an integration test that reads from a source path producing blob references, then writes to a namespace-insert target with `use_namespace_insert=true` through a shuffled or joined DataFrameWriterV2 append, and verifies byte equality in the target table. The test also needs to prove that the namespace insert writer path was selected; byte equality alone can pass through the default writer. If the test is outside `TestDMLNamespaceInsert`, update `NAMESPACE_INSERT_PYTEST_CMD` and mark it `@pytest.mark.rest_dir_compatible`.

.github/workflows/spark-namespace-insert.yml
```diff
@@ -0,0 +1,178 @@
...
+name: Spark Namespace Insert Docker
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - ready_for_review
+      - reopened
+    paths:
```
LuciferYang, line 22: This workflow is intended to cover namespace insert, but the current `paths` mostly cover write/options/integration files. Namespace insert also depends on the Arrow IPC writer, blob-reference read/write path, source-context optimizer rule, session-extension injection, sharding utilities, namespace catalog construction, and runtime namespace wiring. Prefer source-area filters instead of enumerating individual dependencies, for example `lance-spark-base_2.12/src/main/**`, `lance-spark-base_2.12/src/test/**`, `lance-spark-*/src/main/**`, and `lance-spark-*/src/test/**`, so key path changes still run the local/rest-dir namespace insert tests.

docs/src/config.md
```diff
@@ -71,6 +71,23 @@ and namespace-specific options:
...
+
+| Option                         | Type    | Default | Description                                                                                              |
+|--------------------------------|---------|---------|----------------------------------------------------------------------------------------------------------|
+| `use_namespace_insert`         | Boolean | `false` | Use the Lance Namespace insert API for eligible append writes to namespace-backed tables.                 |
+| `namespace_insert_parallelism` | Integer | `0`     | Number of writer tasks to request from Spark for namespace insert writes. `0` preserves Spark's plan. For sharded tables Spark uses the sharding distribution; for unsharded tables Spark repartitions by the first output column. |
+| `batch_size`                   | Integer | `8192`  | Maximum rows per Arrow batch/request before flushing.                                                    |
+| `max_batch_bytes`              | Long    | `268435456` | Maximum approximate bytes per Arrow batch/request before flushing.                                  |
+
+Namespace insert writes are intended for append ingestion through a namespace implementation,
```
LuciferYang, line 86: The docs currently say that rows may be visible after failures, but they do not say that namespace insert is currently an at-least-once path, nor that a Spark job can succeed with duplicate rows after task retry or speculation. Please update both user-facing docs to state that replay de-duplication can only be claimed when the connector verifies a first-class namespace idempotency capability with defined replay semantics. Otherwise users should treat this as at-least-once ingestion. Users who need Spark driver-side atomic commit semantics should use the default writer.

PR #589, opened 2026-06-08
fix: preserve write options through defaults and overwrite

LuciferYang:
## Summary

Two write-option paths were lossy:

- `SparkWriteBuilder.truncate()` rebuilt `LanceSparkWriteOptions` by hand, so newer fields such as `max_batch_bytes`, `blob_pack_file_size_threshold`, and `version` were dropped on overwrite. The overwrite path now uses `writeOptions.toBuilder().writeMode(OVERWRITE)`, which carries every field.
- `LanceSparkWriteOptions.Builder.withCatalogDefaults()` merged catalog storage options but did not promote catalog-level write defaults into the typed fields used by the writer. It now re-parses the merged map through `fromOptions` (mirroring `LanceSparkReadOptions.Builder.withCatalogDefaults`), so typed fields and the storage-options map stay consistent.

Precedence after the re-parse (same semantics as the read path):

- Options supplied via `fromOptions` / per-write `.option(...)` override catalog defaults.
- Typed Builder setters called before `withCatalogDefaults` are re-derived from the merged map, so a catalog default wins for keys it defines; setters for keys absent from the catalog survive.

## Tests

- `LanceSparkWriteOptionsTest`: catalog-default promotion into typed fields, fromOptions-over-catalog precedence (typed fields and storage options asserted consistent), catalog-over-prior-setter semantics, setter survival for catalog-absent keys, no mutation of caller option maps, and a Java-serialization round-trip of the rebuilt overwrite options (including the transient `namespace` invariant via a non-serializable stub).
- `SparkWriteTest#testTruncatePreservesWriteOptionsAndOverwritesMode`: every write-option field (including `namespace`, `tableId`, `version`) survives `truncate()`.
- `./mvnw -pl <module> -Dtest=LanceSparkWriteOptionsTest,SparkWriteTest test` for `lance-spark-3.4_2.12`, `lance-spark-3.5_2.13`, `lance-spark-4.0_2.13`, `lance-spark-4.1_2.13`
- `./mvnw -pl lance-spark-3.5_2.13 test` (818 tests, 0 failures)
- `make lint` and `git diff --check`

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
@@ -28,7 +28,10 @@ import java.util.Objects;
 /**
  * Write-specific options for Lance Spark connector.
  *
- * <p>These options override catalog-level settings for write operations.
+ * <p>Options supplied via {@link Builder#fromOptions(Map)} override catalog-level settings for
+ * write operations. Typed Builder setters applied before {@link
+ * Builder#withCatalogDefaults(LanceSparkCatalogConfig)} are superseded by catalog values for the
+ * same key (see that method's javadoc).
  *
  * <p>Usage:
  *
@@ -516,14 +519,17 @@ public class LanceSparkWriteOptions implements Serializable {
     }
 
     /**
-     * Merges catalog config options as defaults (write options override).
+     * Merges catalog config options as defaults; options supplied via {@link #fromOptions(Map)}
+     * override catalog values.
      *
      * <p>Re-parses the merged map through {@link #fromOptions(Map)} so that catalog-level write
-     * settings are promoted into their typed Builder fields and stay consistent with the storage
-     * options map. Mirrors {@code LanceSparkReadOptions.Builder#withCatalogDefaults}. Typed values
+     * settings (e.g. {@code spark.sql.catalog.<name>.<key>}) are promoted into their typed Builder
+     * fields and take effect on write paths that do not supply per-write {@code .option(...)}
+     * overrides. Mirrors {@code LanceSparkReadOptions.Builder#withCatalogDefaults}. Typed values
      * previously assigned via Builder setters are re-derived from the merged map, so a catalog
-     * default wins over an earlier setter call for the same key; options supplied via {@link
-     * #fromOptions(Map)} always stay ahead of catalog defaults.
+     * default wins over an earlier setter call for the same key. Call this after {@link
+     * #fromOptions(Map)} / {@link #storageOptions(Map)}: a later {@code fromOptions} call replaces
+     * the storage options map and discards the merged catalog defaults.
      *
      * @param catalogConfig the catalog config
      * @return this builder
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
@@ -525,8 +525,7 @@ public class LanceSparkWriteOptions implements Serializable {
       // Merge storage options: catalog options are defaults, current options override
       Map<String, String> merged = new HashMap<>(catalogConfig.getStorageOptions());
       merged.putAll(this.storageOptions);
-      this.storageOptions = merged;
-      return this;
+      return fromOptions(merged);
     }
 
     public LanceSparkWriteOptions build() {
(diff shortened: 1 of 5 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
```diff
@@ -522,10 +558,12 @@ public Builder fromOptions(Map<String, String> options) {
      * @return this builder
      */
     public Builder withCatalogDefaults(LanceSparkCatalogConfig catalogConfig) {
```
geruh, line 537: we can follow the read behavior here without the state unless im missing somehting. 


```java
public Builder withCatalogDefaults(LanceSparkCatalogConfig catalogConfig) {
  Map<String, String> merged = new HashMap<>(catalogConfig.getStorageOptions());
  merged.putAll(this.storageOptions);
  return fromOptions(merged);
}
```

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
```diff
@@ -535,5 +573,102 @@ public LanceSparkWriteOptions build() {
       }
       return new LanceSparkWriteOptions(this);
     }
+
+    private static class WriteOptionSnapshot {
```
geruh: Do we need this? This makes it such that every future write option now has to touch five different places. The field, the setter, the from branches, to builder and the snapshot. Seems like a lot for one option and it might be easy to miss this. If we move to a reparser strategy we could get around this

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
```diff
@@ -535,5 +573,102 @@ public LanceSparkWriteOptions build() {
...
+        this.useLargeVarTypesSet = builder.useLargeVarTypesSet;
+        this.maxBatchBytesSet = builder.maxBatchBytesSet;
+        this.blobPackFileSizeThresholdSet = builder.blobPackFileSizeThresholdSet;
+      }
+
+      private static WriteOptionSnapshot capture(Builder builder) {
+        return new WriteOptionSnapshot(builder);
+      }
+
+      private void restoreExplicitOptions(Builder builder) {
```
geruh: I have a question. So when a field was set from the builder method and not from options, this restores the typed value. But storage options still carry the catalog string for that key so if we have:

```java
.batchSize(1024).withCatalogDefaults(cfg)   // <-- cfg has batch_size=512
```

ends with `getBatchSize() == 1024` but `getStorageOptions().get("batch_size") == "512"`.

Idk if we'd actually hit this but looks like there are two sources of truths here. If we handle reparsing like mentioned above we could get around it maybe.

geruh (commented):
Thanks for this @LuciferYang, took a quick pass over this, and left some comments.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
LuciferYang: Done in 4785018 — `withCatalogDefaults` now matches the read path exactly: merge catalog options under current options, then `fromOptions(merged)`. No snapshot state.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
LuciferYang: Removed the `*Set` flags and `WriteOptionSnapshot` entirely in 4785018 and switched to the re-parse strategy. A new write option is back to the same touch points as before this PR (field, setter, `fromOptions` branch — plus `toBuilder`, which pre-dates this PR and is used by `withVersion`).

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkWriteOptions.java
LuciferYang: With the re-parse in 4785018 this divergence can no longer happen: typed fields are re-derived from the merged map, so `.batchSize(1024).withCatalogDefaults(cfg)` with `cfg` carrying `batch_size=512` now yields `getBatchSize() == 512` and `getStorageOptions().get("batch_size") == "512"` — one source of truth, same semantics as the read path (builder setters before `withCatalogDefaults` are re-derived from the merged map; options supplied via `fromOptions` always stay ahead of catalog defaults). Added `testCatalogDefaultsKeepTypedFieldsAndStorageOptionsConsistent` covering exactly this case.

geruh (approved):
LGTM, thanks @LuciferYang!

LuciferYang:
Thank you @geruh

PR #600, opened 2026-06-09
ci: improve Docker test-image build

geruh:
The Glue/S3 job failed before running any test when Docker Hub returned a 504 resolving the frontend image. This PR removes some of the logic causing the failures:

1.  The header `# syntax=docker/dockerfile:1` across a few files that was forcing the buildkit to fetch frontend images from docker on every build. 
2. `docker-build-test` ran build without cache and the we are busting a cache in the image so it shouldnt be needed. Especailly with the copy operation since it's invalidating on operation.

### Testing

Run the tests here, and we have only removals

### DIFF
--- a/docker/Dockerfile.test-base
+++ b/docker/Dockerfile.test-base
@@ -1,5 +1,3 @@
-# syntax=docker/dockerfile:1
-
 FROM ubuntu:24.04
 
 RUN apt-get update && \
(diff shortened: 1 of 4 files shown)


docker/Dockerfile.test-base
geruh, line 27: will need to fix in a follow up :)

PR #616, opened 2026-06-11
fix: pass namespace storage options to driver commits

### LINKED ISSUE #615: Lance Spark writes fail with REST namespace returned storage options
## Problem

When using Lance Spark with a REST namespace implementation, writes can still fail if storage options are not configured on the Spark client and are instead returned by the REST namespace from `describeTable`.

Example Spark catalog config:

```text
spark.sql.catalog.lance=org.lance.spark.LanceNamespaceSparkCatalog
spark.sql.catalog.lance.impl=rest
spark.sql.catalog.lance.uri=http://127.0.0.1:9101/lance
spark.sql.catalog.lance.parent=lance_catalog
```

The parent catalog/table metadata on the REST namespace side contains storage options, for example:

```text
lance.storage.endpoint=http://127.0.0.1:9000
lance.storage.allow_http=true
lance.storage.access_key_id=...
lance.storage.secret_access_key=...
lance.storage.region=us-east-1
```

A simple write can still fail:

```sql
CREATE NAMESPACE lance.demo;
CREATE TABLE lance.demo.orders (id INT, name STRING) USING lance;
INSERT INTO lance.demo.orders VALUES (1, 'alice'), (2, 'bob');
```

This means the REST namespace response contains the information needed to access object storage, but Lance Spark does not successfully use those returned storage options in the write path.

## Expected Behavior

If the Spark client does not configure `spark.sql.catalog.lance.storage.*`, Lance Spark should use the `storage_options` returned by the REST namespace `describeTable` response when opening the dataset.

This should work for both read and write paths.

## Actual Behavior

The write path still fails unless the Spark client explicitly configures storage options locally.

## Why This Matters

One key benefit of using a REST namespace is centralizing table metadata and storage configuration on the server side. Spark clients should not need to duplicate object storage credentials/options locally if the namespace can return them through `describeTable`.

FANNG1:

Fixes #615.

Driver-side commit paths now pass merged storage options to Lance commit builders.

When a REST namespace returns storage options from `describeTable`, executor-side fragment writes can use them, but several driver-side commit paths still used only client-configured storage options. This caused writes to fail when Spark clients relied on namespace-provided storage options instead of configuring `spark.sql.catalog.<catalog>.storage.*` locally.

This PR applies the merged storage options to:

- append/overwrite commits
- add-columns backfill commits
- update-columns backfill commits
- row-level position delta commits for Spark 3.4 and 3.5

## Are these changes tested?

- `./mvnw -pl lance-spark-bundle-3.5_2.12 -am package -DskipTests`
- `./mvnw -pl lance-spark-bundle-3.4_2.12 -am package -DskipTests`
- Manual E2E with a Gravitino Lance REST namespace returning storage options, MinIO storage, and no Spark client-side `spark.sql.catalog.lance.storage.*` config:
  - `CREATE TABLE`
  - `INSERT`
  - `SELECT COUNT(*)`
  - `ALTER TABLE ... ADD COLUMNS ... FROM`
  - `ALTER TABLE ... UPDATE COLUMNS ... FROM`
  - row-level `UPDATE ... SET`


hamersaw (approved):
Thanks for the fix! I just updated to use the existing `LanceRuntime.mergeStorageOptions` helper function instead of defining a new one.

FANNG1:
> Thanks for the fix! I just updated to use the existing `LanceRuntime.mergeStorageOptions` helper function instead of defining a new one.

Thanks for the review and help!

PR #617, opened 2026-06-11
fix: close dataset when add-index job fails

LuciferYang:
`AddIndexExec.run()`'s FTS/INVERTED path opens a driver-side Lance `Dataset` (a native handle), then runs `FragmentBasedIndexJob` outside the `try/finally` that closes it. A failure during the build or the commit leaks the handle on every failed `CREATE INDEX`, and since the handle is native it stays leaked until the driver exits. Moving the job inside the `try` lets the `finally` close it on all paths.

The BTREE and ZONEMAP paths commit via segments and don't hold a driver dataset across the job, so they're unaffected — #612's segment-commit rework already removed the same pattern from the BTREE path.

### TESTING

New test `testCreateIndexFailureLeavesTableUsable`: an FTS index on a non-text column fails inside the distributed build, after the driver dataset is opened. A `SparkException` in the cause chain pins the failure to the build job, so the test turns red if the failure ever moves before the dataset open. It then checks the manifest holds no index, a full scan still returns every row, and a later valid `CREATE INDEX` succeeds.

`AddIndexTest` passes on Spark 3.5/Scala 2.12 (26 tests) and Spark 4.0/Scala 2.13 (53 tests); spotless and checkstyle clean.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -183,9 +183,10 @@ case class AddIndexExec(
     val dataset = Utils.openDatasetBuilder(readOptions).build()
 
     // The finally closes the dataset on every exit path, including a failure of the
-    // index-build job or the commit below. Partial per-fragment metadata that workers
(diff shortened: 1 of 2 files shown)


hamersaw (approved):
Thanks for the fix!

LuciferYang:
Thank you @hamersaw

PR #630, opened 2026-06-16
feat(cbo): add ANALYZE TABLE persistent column statistics

### LINKED ISSUE #629: feat(cbo): ANALYZE TABLE persistent column statistics (reuse Spark's stats infra)
## Summary

Add `ANALYZE TABLE … COMPUTE STATISTICS [FOR ALL COLUMNS | FOR COLUMNS …]` support for
Lance tables, persisting per-column statistics into the Lance manifest so Spark's
cost-based optimizer (CBO) can consume them on subsequent scans — O(1) per column, with no
per-scan re-aggregation.

The proposed approach **reuses Spark's own statistics infrastructure** rather than
hand-rolling aggregation or a custom serialization format, and adds **no new SQL grammar**
(Spark already parses `ANALYZE`; an injected resolution rule rewrites it for Lance tables).
It complements the connector's existing zonemap-derived runtime statistics (e.g. Dynamic
File Pruning) and is independent of them — no zonemap or btree index is required for the
persisted-stats path.

A prototype implementing this exists on a branch; a PR will follow. This issue is intended to
capture the design and invite feedback before that PR lands — in particular on the two
deliberate tradeoffs called out under **Risks & known limitations** (coupling to Spark
internals, and storing stats in manifest `TBLPROPERTIES`).

## Motivation

Spark's CBO needs column-level statistics (min, max, null count, distinct count, avg/max
length) to produce good join orders and selectivity estimates. Without persisted stats every
scan must either re-aggregate or fall back to coarse heuristics. `ANALYZE TABLE` is the
standard Spark entry point for computing and storing these — but Spark rejects it for V2
tables, and Lance has no place to put the results. This change closes both gaps.

## Approach

**1. Reuse Spark's stats engine instead of hand-rolling aggregation or a codec.**
`ANALYZE` delegates to Spark's own `CommandUtils.computeColumnStats` — one aggregation job
producing min, max, nullCount, HLL-approximate distinctCount, avgLen, maxLen — and persists
each column serialized with Spark's `CatalogColumnStat.toMap`. The read path reads the
payload back via `CatalogColumnStat.fromMap` / `toPlanStat` and feeds Spark's CBO directly.
Using Spark's serialization (rather than a bespoke codec) keeps the connector's stored form
in lockstep with Spark's own semantics — at the cost of coupling to a Spark-internal format
(see **Risks**). Distinct count is always HLL-approximate (there is no EXACT mode), matching
native `ANALYZE`.

**2. No new grammar — an injected resolution rule.**
Spark parses `ANALYZE` natively. A `LanceAnalyzeTableResolution` analyzer rule rewrites the
native `AnalyzeColumn` / `AnalyzeTable` logical plans into `LanceAnalyzeTable` when the
target resolves to a Lance dataset, pre-empting the V2-table rejection Spark would otherwise
throw at planning time (`DataSourceV2Strategy` / `NOT_SUPPORTED_COMMAND_FOR_V2_TABLE`).
Non-Lance `ANALYZE` is intentionally left untouched for Spark to handle. The connector's
custom grammar is retained for the 8 other Lance commands that have no native equivalent;
only the `ANALYZE` rule is moved to native interception.

`NOSCAN`, and a partition spec on the bare `COMPUTE STATISTICS` form, are not supported:
they fall through to Spark's native V2 rejection rather than being silently accepted. A
partition spec combined with `FOR COLUMNS` / `FOR ALL COLUMNS` is dropped by Spark's parser
with a warning (matching native `ANALYZE` column-stats semantics, where column stats are not
partition-scoped) and the full table is analyzed. A bare `COMPUTE STATISTICS` with no `FOR`
clause is intercepted and treated as `FOR ALL COLUMNS` — a deliberate divergence from native
Spark, where the bare form computes only table-level row/size stats and runs no per-column
aggregation.

## Wire format (`lance.stats.version=1`)

Stored in Lance manifest `TBLPROPERTIES`. Envelope plus one block per analyzed column:

```
lance.stats.{version, complete, computedAtVersion, computedAt, numRows, schemaHash}
lance.stats.column.<name>.{version, distinctCount, min, max, nullCount, avgLen, maxLen}
```

Per-column values use Spark's own `CatalogColumnStat` map form. Spark can also emit a
`histogram` entry; the writer **strips it before persisting** and the reader ignores it, so
no `histogram` key is ever written to disk. A reader that encounters an unrecognized
`lance.stats.version` ignores the **entire** payload and falls back to live aggregation. A
per-column `version` newer than the reader's Spark `CatalogColumnStat` version drops **only
that column** (it is omitted from the reported stats) while the remaining columns are still
served. Either way the format is forward-compatible by degradation, not by hard error.

## Correctness guarantees

- **Single-write commit.** The writer assembles one `TableChange[]` ending in
  `lance.stats.complete=true` and submits it via a single `catalog.alterTable` →
  one `Dataset.updateConfig`, so no half-written stats become visible (the read path refuses
  to consume stats unless `complete="true"`). Note this is a read-merge-write of manifest
  properties, not a connector-level compare-and-swap; isolation from other concurrent
  property writers relies on Lance's own commit-conflict handling, not on this connector.
- **Schema-drift guard.** A SHA-256 fingerprint of `(name, dataType, nullable)` triples is
  recorded under `lance.stats.schemaHash`; the read path recomputes it and falls back to
  live aggregation on mismatch or absence (fail-safe, not silent pass-through).
- **Invalidation on data change.** Stats are tagged with the manifest version they describe
  (`computedAtVersion`). Any subsequent commit (append/overwrite/delete/optimize) bumps the
  manifest version, so the read path's exact `computedAtVersion == currentVersion` check no
  longer matches and it falls back to live aggregation — unless `allow.stale` is set.
- **Concurrency (read-side staleness guard).** `ANALYZE` captures the manifest version
  *before* the aggregation read and tags stats with `readVersion + 1`. This guards the
  *reader*: a writer interleaving between capture and `alterTable` makes the tag not match
  the version the stats actually landed at, and the read path rejects them
  ("safer-stale-than-wrong"). It does not add isolation between two concurrent
  `ANALYZE`/writer commits
  (those are subject to the read-merge-write above); the guard's purpose is solely to keep a
  *reader* from consuming stats that don't describe the live data.

## Configuration

| SparkConf | Per-scan option | Default | Purpose |
|---|---|---|---|
| `spark.lance.cbo.column.stats.enabled` | `cbo.column.stats.enabled` | `true` | Master kill-switch for the persisted-stats fast path on the READ side (ANALYZE still writes). The SparkConf key overrides the per-scan option. |
| `spark.lance.cbo.column.stats.allow.stale` | (none) | `false` | Accept stats whose `computedAtVersion` differs from the current manifest version (trades correctness for hit rate). |

## Skipped / rejected types

Complex containers (Struct/Array/Map/UDT/Null), interval types (Calendar / AnsiInterval),
and TimestampNTZ are unsupported. TimestampNTZ is excluded specifically because a filtered
min/max comparison on it throws inside Spark's CBO `FilterEstimation` (`EstimationUtils`
has no `toDouble` case for `TimestampNTZType` → `MatchError`).

- Under `FOR ALL COLUMNS`, unsupported columns are **skipped** (logged at INFO).
- Under `FOR COLUMNS …`, naming an unsupported column **fails fast** with a clear message.

Column names containing `.` cannot be targeted by `FOR COLUMNS` (they would produce
ambiguous keys under the `lance.stats.column.<name>.<suffix>` scheme); `FOR ALL COLUMNS`
skips them. This is a known limitation — see below.

## Risks & known limitations

- **Coupling to Spark internals.** `CommandUtils.computeColumnStats` is reached by reflective
  name/arity lookup (its first parameter type differs between Spark 3.x and 4.x), and the
  on-disk per-column payload is exactly `CatalogColumnStat`'s internal map encoding. Both are
  non-public Spark APIs. The reflective lookup matches by method name and arity, so an arity
  change produces a clear diagnostic, but a same-arity parameter-type or return-type drift
  surfaces as a raw reflective error rather than a named one. Either way, a Spark upgrade that
  changes these will require a corresponding connector change, and the `lance.stats.version`
  envelope does not by itself protect against Spark silently reformatting the per-column map.
- **Manifest size on wide tables.** Stats live in manifest `TBLPROPERTIES`, which is read on
  every scan. For tables with many columns this grows the manifest with each `ANALYZE`. A
  sidecar-file storage medium is a possible alternative and is open for discussion.
- **Dotted column names.** `ANALYZE … FOR COLUMNS` cannot target a top-level column whose
  name contains `.`; the workaround is to omit it (or rename). `FOR ALL COLUMNS` skips such
  columns rather than failing. This is not merely the connector's flat
  `lance.stats.column.<name>.<suffix>` key scheme: Spark's V2 CBO matches column stats to plan
  attributes via `attr.name().equals(ref.describe())` (`DataSourceV2Relation.transformV2Stats`),
  and `ref.describe()` backtick-quotes a dotted name (`` `a.b` `` ≠ `a.b`), so the optimizer
  cannot consume such stats for *any* V2 source — connector-side key encoding alone would not
  make them usable.

## Key classes

- `read/LanceStatsKeys.java` — wire-format spec, key constants, SparkConf constants, schema hash.
- `v2/LanceColumnStatCodec.scala` — read-side bridge: `CatalogColumnStat.fromMap`/`toPlanStat` → V2 `ColumnStatistics`; histogram strip, value-length cap, version gate, fail-safe decode.
- `v2/LanceNativeColumnStats.scala` — reflective bridge to `CommandUtils.computeColumnStats` (3.x/4.x `SparkSession` split).
- `v2/LanceAnalyzeTableResolution.scala` — analyzer rule rewriting native `AnalyzeColumn`/`AnalyzeTable` → `LanceAnalyzeTable` for Lance tables.
- `v2/LanceAnalyzeTableExec.scala` — ANALYZE executor (writer, single-commit ordering, prefix-based orphan-key GC, histogram strip).
- `read/LanceScanBuilder.java` — `loadPersistedColumnStats` fast path.

## Observability

- `LanceScanBuilder` logs at INFO when the fast path engages.
- Schema-hash mismatch logs at INFO; schema-hash absence and stale-version fallback log at DEBUG.
- The `ANALYZE` writer logs analyzed column count + post-write manifest version, and a
  per-orphan-key removal count when schema columns were dropped between runs.

## Test plan

- Unit: `LanceStatsKeysTest`, `LanceAnalyzeTableSchemaHashTest`, `LoadPersistedColumnStatsTest`,
  `LanceSparkReadOptionsSerializationTest`.
- End-to-end: `BaseAnalyzeTableTest` (run per Spark version via `AnalyzeTableTest`) —
  `FOR ALL COLUMNS` / `FOR COLUMNS`, decimal / empty / all-null / re-analyze / partial,
  stats-reach-scan, TimestampNTZ-skip, case-insensitivity, NOSCAN-falls-through-to-Spark,
  and kill-switch paths.
- Coverage: the Spark 3.5 module suite is green locally; Spark 4.0/4.1 reuse the 3.5 test
  sources via `build-helper` and run them in CI; 3.4 compiles. The full four-version +
  integration matrix runs via CI when the PR opens.

Feedback welcome on the design — especially the two tradeoffs in **Risks & known
limitations** — before the PR is opened.

LuciferYang:
## Summary

Adds `ANALYZE TABLE … COMPUTE STATISTICS [FOR ALL COLUMNS | FOR COLUMNS …]` for Lance tables, persisting per-column statistics into the Lance manifest so Spark's CBO can consume them on subsequent scans — O(1) per column, with no per-scan re-aggregation.

The implementation **reuses Spark's own statistics infrastructure** rather than hand-rolling aggregation or a serialization format, and adds **no new SQL grammar** (Spark already parses `ANALYZE`; an injected resolution rule rewrites it for Lance tables).

Full design, wire format, correctness guarantees, configuration, and known limitations are in **#629** — this PR implements that design.

Closes #629
Fixes part of the CBO roadmap (persistent column statistics).

## What's in here

- **Reuse Spark's stats engine.** `ANALYZE` delegates to `CommandUtils.computeColumnStats` (one aggregation job: min / max / nullCount / HLL-approx distinctCount / avgLen / maxLen) and persists each column with Spark's `CatalogColumnStat.toMap`. The read path decodes via `CatalogColumnStat.fromMap` / `toPlanStat` and feeds the CBO directly.
- **No new grammar.** `LanceAnalyzeTableResolution` rewrites Spark's native `AnalyzeColumn` / `AnalyzeTable` → `LanceAnalyzeTable` when the target is a Lance dataset, pre-empting the V2-table rejection thrown in `DataSourceV2Strategy`. Non-Lance `ANALYZE` is untouched.
- **Wire format** (`lance.stats.version=1`): an envelope plus `lance.stats.column.<name>.<suffix>` per-column keys in Spark's `CatalogColumnStat` map form, in manifest `TBLPROPERTIES`.
- **Correctness guarantees:** single-commit write ending in `complete=true`; SHA-256 schema-drift guard; `computedAtVersion` exact-equality invalidation; a `readVersion+1` read-side staleness guard. Fail-safe everywhere (any decode problem → live aggregation).
- **Config:** `spark.lance.cbo.column.stats.enabled` (master kill-switch, default `true`) and `spark.lance.cbo.column.stats.allow.stale` (default `false`).
- **Skipped/rejected types:** complex containers, interval types, and TimestampNTZ are skipped under `FOR ALL COLUMNS` / rejected under `FOR COLUMNS`.

## Key classes

- `read/LanceStatsKeys.java` — wire-format spec, key constants, schema hash.
- `v2/LanceNativeColumnStats.scala` — reflective bridge to `CommandUtils.computeColumnStats` (3.x/4.x split).
- `v2/LanceColumnStatCodec.scala` — read-side `CatalogColumnStat` → V2 `ColumnStatistics` bridge (fail-safe).
- `v2/LanceAnalyzeTableResolution.scala` — analyzer rule rewriting native ANALYZE for Lance tables.
- `v2/LanceAnalyzeTableExec.scala` — ANALYZE executor (writer, single-commit ordering, orphan-key GC).
- `read/LanceScanBuilder.java` — `loadPersistedColumnStats` fast path.

## Test plan

- **Unit:** `LanceStatsKeysTest`, `LanceAnalyzeTableSchemaHashTest`, `LoadPersistedColumnStatsTest`, `LanceSparkReadOptionsSerializationTest`.
- **End-to-end:** `BaseAnalyzeTableTest` (run per Spark version via `AnalyzeTableTest`) — `FOR ALL COLUMNS` / `FOR COLUMNS`, decimal / empty / all-null / re-analyze / partial, stats-reach-scan, TimestampNTZ-skip, case-insensitivity, NOSCAN-falls-through-to-Spark, and kill-switch paths.
- Verified locally: the Spark **3.5** module suite is green (base + 3.5, incl. the new `AnalyzeTableTest`). Spark **4.0 / 4.1** reuse the 3.5 test sources via `build-helper`; **3.4** compiles. The full four-version + integration matrix runs in CI.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatsKeys.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatsKeys.java
@@ -1,0 +1,237 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.read;
+
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
+
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+
+/**
+ * Constants and helpers that define the on-disk wire format of {@code ANALYZE TABLE} persisted
+ * column statistics in TBLPROPERTIES. Centralizing these in one base-package utility keeps the
+ * ANALYZE writer (in {@code org.apache.spark.sql.execution.datasources.v2}) and the scan-time
+ * reader (in {@code org.lance.spark.read}) on the same source of truth without forcing the read
+ * path to depend on a class inside Spark's internal package namespace.
+ *
+ * <h2>Wire format (lance.stats.version = 1)</h2>
+ *
+ * <ul>
+ *   <li>{@code lance.stats.version} = "1" — bumping this invalidates older readers.
+ *   <li>{@code lance.stats.computedAtVersion} = decimal Lance manifest version stats reflect.
+ *   <li>{@code lance.stats.computedAt} = ISO-8601 instant.
+ *   <li>{@code lance.stats.numRows} = decimal row count at ANALYZE time.
+ *   <li>{@code lance.stats.schemaHash} = SHA-256 hex of the (name, dataType, nullable) tuple
+ *       stream.
+ *   <li>{@code lance.stats.column.<name>.{version,min,max,nullCount,distinctCount,avgLen,maxLen}} —
+ *       per-column stats. The suffix set and string encoding are Spark's own {@code
+ *       CatalogColumnStat} map form (written by {@code CatalogColumnStat.toMap}, read by {@code
+ *       CatalogColumnStat.fromMap}); {@code <name>} is a top-level column name and the format does
+ *       not currently support nested-leaf paths (see {@link #validateColumnName} below). The
+ *       per-column {@code .version} is Spark's CatalogColumnStat serialization version, distinct
+ *       from the table-level {@link #VERSION}. {@code distinctCount} is HLL-approximate (Spark's
+ *       native ANALYZE behavior). No {@code histogram} entry is ever written: the ANALYZE writer
+ *       forces {@code spark.sql.statistics.histogram.enabled} off for the aggregation, so {@code
+ *       toMap} never produces one (the read path discards histograms anyway — the V2 {@code
+ *       ColumnStatistics} histogram type differs from catalyst's).
+ * </ul>
+ *
+ * <p>The format is a stability contract: existing tags must keep their semantics across connector
+ * versions. New stats tags may be added at the same {@code lance.stats.version}; format-breaking
+ * changes (existing tag's encoding altered, key prefix renamed) require bumping {@code
+ * lance.stats.version} so older readers report no column stats rather than silently misinterpret
+ * the payload.
+ *
+ * <h2>Design choices and known gaps vs. upstream conventions</h2>
+ *
+ * <ul>
+ *   <li><b>Storage medium:</b> stats live in TBLPROPERTIES on the Lance manifest, not in a Puffin
+ *       sidecar (Iceberg's choice) or a separate stats table (Delta's choice). The Lance manifest
+ *       is the single source of truth; sidecar files would require a second read path and add a
+ *       Puffin dependency. The trade-off is that very large per-column stats (histograms, sketches)
+ *       are awkward — but at v1 we only persist scalar bounds and counts, which fit.
+ *   <li><b>Per-column only, not table-level:</b> Spark's native {@code ANALYZE TABLE … COMPUTE
+ *       STATISTICS} updates both row-count and on-disk-size. This implementation persists row count
+ *       under {@code lance.stats.numRows} for human / tooling inspection only — the read path
+ *       always takes BOTH {@code numRows} and {@code sizeInBytes} from {@link
+ *       org.lance.ManifestSummary} (live), never from the persisted payload, so {@code
+ *       lance.stats.numRows} never feeds the table-size or row-count estimate, regardless of the
+ *       FOR clause.
+ *   <li><b>{@code avgLen} / {@code maxLen} persisted:</b> Spark's CBO uses these for join size
+ *       estimation on string/binary columns. They come for free from {@code
+ *       CommandUtils.computeColumnStats} and are surfaced to the CBO via the read path's {@code
+ *       ColumnStatistics}.
+ *   <li><b>Output row deviates from native ANALYZE TABLE:</b> Spark's {@code AnalyzeTableCommand}
+ *       returns zero rows; this command returns one row of {@code (columns_analyzed, rows_scanned,
+ *       manifest_version)} for tooling and operator diagnostics.
+ * </ul>
+ */
+public final class LanceStatsKeys {
+
+  /** Top-level table-property keys. */
+  public static final String VERSION = "lance.stats.version";
+
+  public static final String COMPUTED_AT_VERSION = "lance.stats.computedAtVersion";
+  public static final String COMPUTED_AT = "lance.stats.computedAt";
+  public static final String NUM_ROWS = "lance.stats.numRows";
+  public static final String SCHEMA_HASH = "lance.stats.schemaHash";
+
+  /**
+   * Per-column key prefix; a full key is {@code COLUMN_PREFIX + name + <suffix>}, where {@code
+   * <suffix>} (e.g. {@code .min}, {@code .distinctCount}) is whatever Spark's own {@code
+   * CatalogColumnStat.toMap} emits — the writer serializes each stat with {@code toMap} and the
+   * reader deserializes with {@code CatalogColumnStat.fromMap}, so the per-column payload IS
+   * Spark's native string form, prefixed with {@link #COLUMN_PREFIX}. Lance code never enumerates
+   * the suffixes (the format is delegated to Spark); only tests assert the concrete key shape, via
+   * {@code LanceStatsTestKeys}.
+   */
+  public static final String COLUMN_PREFIX = "lance.stats.column.";
+
+  /** Format version this connector knows how to read and write. */
+  public static final String SUPPORTED_VERSION = "1";
+
+  // SparkConf keys below are part of the operator-facing configuration contract — once a
+  // release ships, renaming requires a deprecation cycle. All share the prefix
+  // `spark.lance.cbo.column.stats.` for namespace introspection.
+
+  /**
+   * Master kill-switch for the column-stats fast path. Overrides per-scan {@code
+   * lance.cbo.column.stats.enabled}. Default: {@code true}.
+   */
+  public static final String SPARK_CONF_CBO_COLUMN_STATS_ENABLED =
+      "spark.lance.cbo.column.stats.enabled";
+
+  /**
+   * If {@code true}, accept persisted stats whose {@code computedAtVersion} differs from the
+   * current Lance manifest version. Default: {@code false}. Trades CBO correctness for fast-path
+   * hit rate; only enable when stale estimates are acceptable.
+   */
+  public static final String SPARK_CONF_CBO_COLUMN_STATS_ALLOW_STALE =
+      "spark.lance.cbo.column.stats.allow.stale";
+
+  private LanceStatsKeys() {}
+
+  /** Maximum length of a column name surfaced in error messages, to bound exception size. */
+  private static final int MAX_NAME_IN_ERROR = 256;
+
+  /**
+   * True when {@code name} produces a non-ambiguous TBLPROPERTIES key under {@link #COLUMN_PREFIX}.
+   * Ambiguity sources today: dots in the name (collide with future nested-leaf paths) and the empty
+   * string (would emit a key like {@code lance.stats.column..min} that orphan-cleanup would later
+   * misattribute).
+   */
+  public static boolean isStatsCompatibleColumnName(String name) {
+    return name != null && !name.isEmpty() && name.indexOf('.') < 0;
+  }
+
+  /**
+   * Reject column names that would produce ambiguous TBLPROPERTIES keys: dotted names collide with
+   * future nested-leaf paths (and would mis-attribute under the prefix-based orphan-cleanup sweep,
+   * which splits the column segment on the first '.'), empty names produce a malformed {@code
+   * lance.stats.column..<suffix>} key. Forbidding at write time is more conservative than
+   * introducing wire-format escaping.
+   *
+   * @throws IllegalArgumentException if the name is null, empty, or contains a dot.
+   */
+  public static void validateColumnName(String name) {
+    if (name == null) {
+      throw new IllegalArgumentException("Column name must not be null");
+    }
+    if (name.isEmpty()) {
+      throw new IllegalArgumentException(
+          "ANALYZE TABLE does not support empty column names. Persisted-stats keys would be "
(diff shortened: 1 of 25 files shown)


LuciferYang:
cc @hamersaw @fangbo FYI

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceAnalyzeTableExec.scala
```diff
@@ -0,0 +1,383 @@
...
+    changes += TableChange.setProperty(LanceStatsKeys.SCHEMA_HASH, schemaHash)
+
+    // Per-column stats via Spark's CatalogColumnStat codec: serialize each ColumnStat (internal
+    // catalyst rep) to Spark's `<col>.{version,distinctCount,min,max,nullCount,avgLen,maxLen}`
+    // string map, under our column prefix. Re-ANALYZE correctness: write exactly the keys this run
+    // produced and REMOVE any prior key for the column this run did NOT produce (e.g. min/max gone
+    // after the column became all-NULL), so the reader can't serve a stale value under the fresh
+    // manifest version. (Columns not analyzed at all are cleared by staleKeys.)
+    //
+    // The histogram is intentionally dropped (when spark.sql.statistics.histogram.enabled is on,
```
summaryzb: Maybe it's better to skip histogram computing ahead, rather than skipping persist the histogram result

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceAnalyzeTableExec.scala
```diff
@@ -0,0 +1,383 @@
...
+          }
+          .getOrElse(Map.empty[String, String])
+      produced.foreach { case (k, v) => changes += TableChange.setProperty(k, v) }
+      existingProps.keys
+        .filter(k => k.startsWith(keyPrefix) && !produced.contains(k))
+        .foreach(k => changes += TableChange.removeProperty(k))
+    }
+
+    // Step 4: complete-marker LAST — only valid stats payloads have lance.stats.complete=true.
+    changes += TableChange.setProperty(LanceStatsKeys.COMPLETE, "true")
```
summaryzb: Lance table atomically alterTable properties, `valid stats` is over-design

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatsKeys.java
```diff
@@ -0,0 +1,253 @@
...
+   * Per-column key prefix; a full key is {@code COLUMN_PREFIX + name + <suffix>} where {@code
+   * <suffix>} is one of the {@code COLUMN_SUFFIX_*} values below. Those suffixes intentionally
+   * mirror Spark's own {@code CatalogColumnStat} map keys (e.g. {@code "<col>.min"}, {@code
+   * "<col>.distinctCount"}), because the writer serializes each stat with {@code
+   * CatalogColumnStat.toMap} and the reader deserializes with {@code CatalogColumnStat.fromMap} —
+   * the per-column payload IS Spark's native string form, prefixed with {@link #COLUMN_PREFIX}.
+   */
+  public static final String COLUMN_PREFIX = "lance.stats.column.";
+
+  public static final String COLUMN_SUFFIX_VERSION = ".version";
```
summaryzb: It seems all the suffix variant only used in test cases

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceAnalyzeTableExec.scala
LuciferYang: Makes sense, building it just to throw it away is wasted work. `LanceNativeColumnStats` now turns `spark.sql.statistics.histogram.enabled` off around the aggregation (and restores it after), so Spark never computes a histogram in the first place. Dropped the post-serialization strip too. The test that sets `histogram.enabled=true` still passes, now because nothing gets produced rather than produced-then-dropped.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceAnalyzeTableExec.scala
LuciferYang: Agreed. The whole change set goes in via a single `updateConfig`, so there's no half-written window for the marker to guard. I removed `lance.stats.complete` end to end: the two writes, the read-side gate, and the constant. Validity now keys off `version` + `computedAtVersion` + `schemaHash`, which already covered everything the flag did.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatsKeys.java
LuciferYang: Yeah. The read and write paths both go through Spark's `CatalogColumnStat.toMap`/`fromMap`, so prod never references the suffixes directly. Moved them into a test-only `LanceStatsTestKeys` and left `LanceStatsKeys` with just the keys production actually reads or writes.

fangbo:
Hi, @LuciferYang There some discussions about column statistics in lance format . See https://github.com/lance-format/lance/discussions/4540.

LuciferYang:
Thanks @fangbo, this is the context I was missing.

Agreed that column statistics really belong in the format itself. Per-fragment collection consolidated into a mergeable, eventually default-on store is a much better long-term home than recomputing in Spark and parking the result in TBLPROPERTIES.

Since @HaochengLIU's native work (lance #5639, #5967) is already underway and covers the same ground, I'd rather not ship an interim Spark-side mechanism that we'd only have to migrate off later. So I'll put this PR on hold and wait for the native column-statistics API to land, then wire lance-spark's CBO up to `Dataset.column_statistics()` directly.

Happy to help review/test on the native side in the meantime, and to align on the fields Spark's CBO needs (min/max/nullCount/distinctCount/avgLen/maxLen) so they're covered when it lands.

PR #640, opened 2026-06-18
fix: preserve target column metadata on update columns from command

### LINKED ISSUE #641: preserve target column metadata on update columns from command
## What changes are proposed in this pull request?
When running `ALTER TABLE ... UPDATE COLUMNS ... FROM <source>`, the write schema was taken directly from the source query's schema. The source query, such as a temp view or SELECT, does not carry the Arrow metadata defined on the target table columns, most notably `arrow:large-var-char`, which marks a column as Arrow `LargeUtf8`.

As a result, updating a large-varchar column could lose its `large-var-char` metadata on write, causing the column to be treated as regular `Utf8` and producing a type mismatch against the existing target column.

This change adds `UpdateColumnsBackfillExec.withTargetMetadataForUpdateColumns`, which merges each target column's metadata into the corresponding field of the source query schema before building the write `LanceDataset`, so updated columns keep their original metadata, including `arrow:large-var-char`.

zhangyue19921010:
Closes: https://github.com/lance-format/lance-spark/issues/641


zhangyue19921010:
Hi @fangbo Would u mind to take a look? Thanks!

fangbo:
+1 LGTM

PR #654, opened 2026-06-24
feat: support ddl for tag operation

### LINKED ISSUE #651: Support tag DDL
This PR https://github.com/lance-format/lance-spark/pull/576 has implements branch DDL. The tag DDL should inherit the DDL grammar. 

 # Create Tag

```sql
ALTER TABLE <table_name> CREATE TAG (IF NOT EXISTS) <tag_name> (AS OF VERSION <main_version>
ALTER TABLE <table_name> CREATE TAG (IF NOT EXISTS) <tag_name> AS OF BRANCH <branch_name> (VERSION <branch_version>
```

# Drop Tag

```sql
ALTER TABLE <table_name> DROP TAG (IF EXISTS) <tag_name>
```

# Show Tags

```sql
SHOW (TAGS | TAG) (FROM | IN) <table_name>
```

fangbo:
Close https://github.com/lance-format/lance-spark/issues/651

### DIFF
--- a/docs/src/operations/ddl/create-tag.md
+++ b/docs/src/operations/ddl/create-tag.md
@@ -91,7 +91,6 @@ The `CREATE TAG` command returns:
 - `CREATE TAG` is implemented as a Spark SQL extension command.
 - The referenced table must be a Lance table.
 - Creating a tag from a non-existent branch, tag, or version returns an error.
-- The tag name is returned even when the command result contains only a single row.
 
 ## See Also
 
--- a/docs/src/operations/ddl/create-tag.md
+++ b/docs/src/operations/ddl/create-tag.md
@@ -1,0 +1,99 @@
+# CREATE TAG
+
+Create a named tag for a Lance table.
+
+!!! warning "Spark Extension Required"
+    This feature requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+## Overview
+
+The `CREATE TAG` command creates a stable named reference to a specific table state. A tag can be created from:
+
+- the latest version of `main`
+- a specific version on `main`
+- the head of another branch
+- a specific version on another branch
+- another tag
+
+## Syntax
+
+=== "SQL"
+
+    ```sql
+    ALTER TABLE <table> CREATE TAG [IF NOT EXISTS] <tag_name>;
+
+    ALTER TABLE <table> CREATE TAG [IF NOT EXISTS] <tag_name>
+    AS OF VERSION <version>;
+
+    ALTER TABLE <table> CREATE TAG [IF NOT EXISTS] <tag_name>
+    AS OF BRANCH <source_branch>;
+
+    ALTER TABLE <table> CREATE TAG [IF NOT EXISTS] <tag_name>
+    AS OF BRANCH <source_branch> VERSION <version>;
+
+    ALTER TABLE <table> CREATE TAG [IF NOT EXISTS] <tag_name>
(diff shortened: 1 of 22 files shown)


fangbo:
@LuciferYang This PR is ready. Could you please look at it ? Thanks a lot.

LuciferYang:
quite busy today and will provide feedback tomorrow.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceShowTags.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+}
+
+object LanceShowTagsOutputType {
+  val SCHEMA = StructType(
+    Array(
+      StructField("name", DataTypes.StringType, nullable = false),
+      StructField("branch", DataTypes.StringType, nullable = true),
+      StructField("version", DataTypes.LongType, nullable = false),
+      StructField("created_at", DataTypes.StringType, nullable = true),
```
LuciferYang: `SHOW BRANCHES` exposes its timestamp as a numeric `created_at LongType`, see `LanceShowBranches.scala` `StructField("created_at", DataTypes.LongType, nullable = false)`. `SHOW TAGS` instead emits a `String`. 

The divergence is rooted in the Lance Java API: `Tag.getCreatedAt()` returns `Optional<Instant>`, whereas `Branch.getCreateAt()` returns a primitive `long` (unix seconds), so the exec renders the tag value via `Instant.toString()`. 

That string is itself a stable, JDK-documented ISO-8601 form (`DateTimeFormatter.ISO_INSTANT`, e.g. `2026-06-24T10:15:30Z`), the concern is not that the format is unstable, but that

1. two sibling commands now return the same conceptual field in two different representations (ISO-8601 string vs numeric epoch), 
2. the connector's own docs don't pin which form `SHOW TAGS` returns. 

Align `SHOW TAGS` to the numeric `LongType` epoch for parity with `SHOW BRANCHES`, or, if the ISO-8601 string is intentional, state that format explicitly in `show-tags.md` so downstream parsers can rely on it.

docs/src/operations/ddl/create-tag.md
```diff
@@ -0,0 +1,99 @@
...
+| Column | Type   | Description            |
+|--------|--------|------------------------|
+| `name` | String | The name of the new tag |
+
+## Notes and Limitations
+
+- `CREATE TAG` is implemented as a Spark SQL extension command.
+- The referenced table must be a Lance table.
+- Creating a tag from a non-existent branch, tag, or version returns an error.
+- The tag name is returned even when the command result contains only a single row.
```
LuciferYang: Reword to describe the actual behavior, e.g. *"The command returns a single row containing the name of the created tag."* or delete it because the `## Output` table already states the schema.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/plans/logical/LanceShowTags.scala
fangbo: Good advice. I have fixed it. The show-tag's output fields: `created_at` and `update_at` are consistent with show-branch.

docs/src/operations/ddl/create-tag.md
fangbo: "_The tag name is returned even when the command result contains only a single row_" is deleted. The `create-branch.md` is also updated.

LuciferYang (approved):
LGTM

PR #658, opened 2026-06-25
fix: reuse dataset during update backfill

zhangyue19921010:
reuse dataset per task during update backfill


zhangyue19921010:
Hi @fangbo PTAL~ It has been fully verified in the test environment and confirmed to take effect

fangbo:
+1, Great optimization

PR #660, opened 2026-06-29
feat: add use_scalar_index in storage options

### LINKED ISSUE #659: Add parameter useScalarIndex for LanceSparkReadOptions
# Background

We have a large table about 1 billion rows. Some columns in this table have B-Tree indexes. When we execute a query, each scan node needs to load the index data. This leads to two undesirable consequences: 

- the memory pressure on each scan executor is very high; 
- the B-Tree index data is read repeatedly, which puts significant pressure on object storage.

# Proposal

Add a parameter `useScalarIndex` to the LanceSparkReadOptions. This parameter will be assigned to `ScanOptions.useScalarIndex`

fangbo:
Close https://github.com/lance-format/lance-spark/issues/659

# Description
This PR adds support for the use_scalar_index read option in LanceSparkReadOptions and wires it through to scan execution.

# What changed

- Added use_scalar_index as a supported read option in LanceSparkReadOptions
- Parsed the option from storage/read properties and exposed it through the builder/getter
- Propagated the option to ScanOptions in both fragment scanning and count(*) execution paths
- Added serialization and option parsing tests for use_scalar_index
- Updated the documentation for read options in select.md
- Fixed an integration test by explicitly enabling spark.sql.catalog.lance.storage.use_scalar_index

# Why
This change allows users to control whether scalar indices are used during scan-time filtering, making the read behavior configurable through Spark read/storage options.

### DIFF
--- a/docs/src/operations/dql/select.md
+++ b/docs/src/operations/dql/select.md
@@ -237,7 +237,7 @@ These options control how data is read from Lance datasets. They can be set usin
 | Option                | Type    | Default | Description                                                                                                       |
 |-----------------------|---------|---------|-------------------------------------------------------------------------------------------------------------------|
 | `batch_size`          | Integer | `8192`  | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
-| `use_scalar_index`    | Boolean | `true`  | Whether to use scalar indices (e.g. zonemap) for filter acceleration during scanning.                             |
+| `use_scalar_index`    | Boolean | `false` | Whether to use scalar indices (e.g. btree) for filter acceleration during scanning.                               |
 | `version`             | Integer | Latest  | Specific dataset version to read. If not specified, reads the latest version.                                     |
 | `block_size`          | Integer | -       | Block size in bytes for reading data.                                                                             |
 | `index_cache_size`    | Integer | -       | Size of the index cache in number of entries.                                                                     |
--- a/docs/src/operations/dql/select.md
+++ b/docs/src/operations/dql/select.md
@@ -236,7 +236,8 @@ These options control how data is read from Lance datasets. They can be set usin
 
 | Option                | Type    | Default | Description                                                                                                       |
 |-----------------------|---------|---------|-------------------------------------------------------------------------------------------------------------------|
-| `batch_size`          | Integer | `512`   | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
+| `batch_size`          | Integer | `8192`  | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
+| `use_scalar_index`    | Boolean | `true`  | Whether to use scalar indices (e.g. zonemap) for filter acceleration during scanning.                             |
 | `version`             | Integer | Latest  | Specific dataset version to read. If not specified, reads the latest version.                                     |
 | `block_size`          | Integer | -       | Block size in bytes for reading data.                                                                             |
 | `index_cache_size`    | Integer | -       | Size of the index cache in number of entries.                                                                     |
--- a/docs/src/operations/dql/select.md
+++ b/docs/src/operations/dql/select.md
@@ -237,7 +237,7 @@ These options control how data is read from Lance datasets. They can be set usin
 | Option                | Type    | Default | Description                                                                                                       |
 |-----------------------|---------|---------|-------------------------------------------------------------------------------------------------------------------|
 | `batch_size`          | Integer | `8192`  | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
-| `use_scalar_index`    | Boolean | `false` | Whether to use scalar indices (e.g. btree) for filter acceleration during scanning.                               |
+| `use_scalar_index`    | Boolean | `true`  | Whether to use scalar indices (e.g. btree) for filter acceleration during scanning.                               |
 | `version`             | Integer | Latest  | Specific dataset version to read. If not specified, reads the latest version.                                     |
 | `block_size`          | Integer | -       | Block size in bytes for reading data.                                                                             |
 | `index_cache_size`    | Integer | -       | Size of the index cache in number of entries.                                                                     |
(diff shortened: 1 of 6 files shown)


fangbo:
@LuciferYang This PR is ready. Could you please look at it ? Thank you.

LuciferYang:
tied‑up today, will review this one tomorrow.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
```diff
@@ -93,6 +94,7 @@ public class LanceSparkReadOptions implements Serializable {
   private static final boolean DEFAULT_PUSH_DOWN_FILTERS = true;
   // Changed from 512 to 8192 for better OLAP scan performance (33x improvement)
   private static final int DEFAULT_BATCH_SIZE = 8192;
+  private static final boolean DEFAULT_USE_SCALAR_INDEX = false;
```
LuciferYang: Seems Lance core's `ScanOptions.useScalarIndex` defaults to `true`. Flipping it to `false` here silently disables scalar-index usage for every existing deployment with a BTree index on a hot predicate column potentially  performance regression on upgrade? So maybe we should match Lance core. WDYT?

lance-spark-base_2.12/src/test/java/org/lance/spark/LanceSparkReadOptionsSerializationTest.java
```diff
@@ -144,4 +144,34 @@ public void testPerReadOptionOverridesCatalogDefaults() {
         options.isExecutorCredentialRefresh(),
         "per-read .option(...) must override the catalog-level default");
   }
+
+  @Test
+  public void testUseScalarIndexFromOptions() {
```
LuciferYang, line 169: The integration test exercises `spark.sql.catalog.lance.storage.use_scalar_index` — i.e., the `withCatalogDefaults` → `parseTypedFlags` path — but no unit test covers it. Mirror `testExecutorCredentialRefreshFromCatalogDefaults` (lines 109-125) with two cases: catalog default reaches the typed field, and per-read overrides the catalog default. Use the non-default value, otherwise a no-op `parseTypedFlags` would still pass.

lance-spark-base_2.12/src/test/java/org/lance/spark/LanceSparkReadOptionsSerializationTest.java
```diff
@@ -144,4 +144,34 @@ public void testPerReadOptionOverridesCatalogDefaults() {
...
+    Map<String, String> properties = new HashMap<>();
+    properties.put(LanceSparkReadOptions.CONFIG_DATASET_URI, "s3://bucket/path");
+    properties.put(LanceSparkReadOptions.CONFIG_USE_SCALAR_INDEX, "false");
```
LuciferYang, line 172: Both tests set `false` and assert `false`, and `DEFAULT_USE_SCALAR_INDEX` is also `false`,  so a no-op `parseTypedFlags` or a no-op `.useScalarIndex(...)` setter still passes.
- `testUseScalarIndexFromOptions`: set/assert the non-default value.
- `testUseScalarIndexSerialization`: round-trip BOTH `true` and `false` (covers field-drop and builder-no-op together, immune to any future default flip).

docs/src/operations/dql/select.md
```diff
@@ -236,7 +236,8 @@ These options control how data is read from Lance datasets. They can be set usin
 
 | Option                | Type    | Default | Description                                                                                                       |
 |-----------------------|---------|---------|-------------------------------------------------------------------------------------------------------------------|
-| `batch_size`          | Integer | `512`   | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
+| `batch_size`          | Integer | `8192`  | Number of rows to read per batch during scanning. Larger values may improve throughput but increase memory usage. |
```
LuciferYang, line 239: This row (`512 → 8192`) is a late doc catch-up for `DEFAULT_BATCH_SIZE` (`LanceSparkReadOptions.java:96`, already `8192` pre-PR) — unrelated to `use_scalar_index`. Not blocking; a one-liner in the PR description can avoids reviewer confusion.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSparkReadOptions.java
fangbo: Good advice. I have update default value to `true` .

lance-spark-base_2.12/src/test/java/org/lance/spark/LanceSparkReadOptionsSerializationTest.java
fangbo: The two cases have added. See `testUseScalarIndexFromCatalogDefaults` and `testPerReadOptionOverridesCatalogDefaults`

lance-spark-base_2.12/src/test/java/org/lance/spark/LanceSparkReadOptionsSerializationTest.java
fangbo: The default value is configured `true`, so the two test cases may make sense.

docs/src/operations/dql/select.md
```diff
@@ -236,7 +236,8 @@ These options control how data is read from Lance datasets. They can be set usin
...
+| `use_scalar_index`    | Boolean | `false` | Whether to use scalar indices (e.g. btree) for filter acceleration during scanning.                               |
```
LuciferYang: change to `true`？

LuciferYang (approved):

docs/src/operations/dql/select.md
fangbo: Thank you. I have fixed it.

LuciferYang (approved):

LuciferYang:
Merged into main. Thanks @fangbo

PR #661, opened 2026-06-29
fix: render pushed-down timestamp literals in UTC

### LINKED ISSUE #662: Pushed-down TimestampType filter literals are rendered in JVM-local time zone
## Summary

Filter pushdown renders `TimestampType` literals using `java.sql.Timestamp.toString()`, which formats in the **JVM default time zone**. Spark stores `TimestampType` as micros-since-epoch **UTC**, so on any driver whose JVM isn't UTC the pushed `timestamp '...'` literal is shifted by the driver's UTC offset, silently dropping or including the wrong rows.

## Affected code

`lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java` (`compileLiteral`, the `TimestampType` branch):

```java
java.time.Instant instant =
    java.time.Instant.ofEpochSecond(micros / 1_000_000, (micros % 1_000_000) * 1_000);
return "timestamp '" + Timestamp.from(instant) + "'";
```

Two related issues in the same expression:

- `Timestamp.toString()` uses the JVM default time zone → literal shifted on non-UTC JVMs.
- `micros / 1_000_000` and `micros % 1_000_000` round toward zero → wrong wall-clock components for pre-1970 (negative) micros.
- `Timestamp.toString()` carries the default `Locale`, which is inappropriate for an SQL literal.

The sibling `Date` branch (`LocalDate.ofEpochDay`) is already TZ-independent, so the bug is timestamp-specific.

## Repro sketch

1. Start a driver with `-Duser.timezone=America/Los_Angeles`.
2. Query a Lance table with `WHERE ts >= TIMESTAMP '2024-01-01 00:00:00'`.
3. The pushed literal is rendered as `2023-12-31 16:00:00` (LA wall clock) and Lance returns rows from the wrong UTC window.

## Suggested fix

Format the `Instant` with a `DateTimeFormatter` pinned to `ZoneOffset.UTC` + `Locale.ROOT`, and switch the micros split to `Math.floorDiv` / `Math.floorMod` so pre-epoch values produce the correct UTC wall clock.

## Notes

I have a fix and tests ready — will link a PR.

LuciferYang:
Closes #662

## Summary

Fix wrong-row drift in pushed-down `TimestampType` predicates on non-UTC drivers.

## Root cause

> `FilterPushDown.compileLiteral` rendered the literal via `java.sql.Timestamp.toString()`, which formats in the JVM default time zone — but Spark stores `TimestampType` as micros-since-epoch **UTC**.

**Before** — `lance-spark-base_2.12/.../read/FilterPushDown.java`:

```java
java.time.Instant instant =
    java.time.Instant.ofEpochSecond(micros / 1_000_000, (micros % 1_000_000) * 1_000);
return "timestamp '" + Timestamp.from(instant) + "'";
```

Problems:

- `Timestamp.toString()` uses the JVM default time zone → literal shifted by the driver's UTC offset on non-UTC JVMs, silently dropping or including the wrong rows.
- `/` and `%` round toward zero → wrong wall-clock components for pre-1970 micros.
- `Timestamp.toString()` carries the default `Locale`, which is inappropriate for an SQL literal.

The sibling `Date` branch (`LocalDate.ofEpochDay`) was already TZ-independent, so the bug was timestamp-specific.

## Fix

**After**:

```java
private static final DateTimeFormatter UTC_TIMESTAMP_FORMATTER =
    new DateTimeFormatterBuilder()
        .appendPattern("uuuu-MM-dd HH:mm:ss")
        .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)
        .toFormatter(Locale.ROOT);

// ...

Instant instant =
    Instant.ofEpochSecond(
        Math.floorDiv(micros, 1_000_000L),
        Math.floorMod(micros, 1_000_000L) * 1_000L);
return "timestamp '"
    + UTC_TIMESTAMP_FORMATTER.format(instant.atOffset(ZoneOffset.UTC))
    + "'";
```

- Pin a `DateTimeFormatter` to `ZoneOffset.UTC` + `Locale.ROOT` so the literal is independent of the driver's JVM time zone and locale.
- Switch the micros split to `Math.floorDiv` / `Math.floorMod` for pre-epoch correctness.
- Emit fractional seconds only when non-zero (`appendFraction(..., 0, 9, true)`).
- `Date` branch unchanged.

## Tests

In `FilterPushDownTest`, using raw `Long`-micros helpers added to `TestPredicates` so the literal isn't pre-shifted by `Timestamp.valueOf` before reaching the converter:

- [x] TZ-independence under `user.timezone=America/Los_Angeles`
- [x] Trailing fractional-zero trimming
- [x] Pre-epoch negative micros via `floorDiv` / `floorMod`
- [x] Range predicates (`>=`, `<`) routed through the same `compileLiteral` chokepoint

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/FilterPushDown.java
@@ -27,12 +27,30 @@ import org.apache.spark.sql.types.DataTypes;
 import java.math.BigDecimal;
 import java.sql.Date;
 import java.sql.Timestamp;
+import java.time.Instant;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeFormatterBuilder;
+import java.time.temporal.ChronoField;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Locale;
 import java.util.Set;
 import java.util.stream.Collectors;
 
 public class FilterPushDown {
+  /**
+   * Renders a Spark {@code TimestampType} (micros-since-epoch UTC) as a {@code uuuu-MM-dd HH:mm:ss}
+   * literal in UTC. Fractional seconds are emitted only when non-zero. Pinned to {@link
+   * ZoneOffset#UTC} and {@link Locale#ROOT} so the pushed literal is independent of the driver
+   * JVM's default time zone and locale.
+   */
+  private static final DateTimeFormatter UTC_TIMESTAMP_FORMATTER =
+      new DateTimeFormatterBuilder()
+          .appendPattern("uuuu-MM-dd HH:mm:ss")
+          .appendFraction(ChronoField.NANO_OF_SECOND, 0, 9, true)
+          .toFormatter(Locale.ROOT);
+
   /**
    * Create SQL 'where clause' from Spark V2 predicates.
    *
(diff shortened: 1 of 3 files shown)


LuciferYang:
cc @fangbo

fangbo:
+1,  Great fix !

LuciferYang:
> +1, Great fix !

Thanks @fangbo , could you please approve ?

LuciferYang:
Thank you @fangbo

Issue #665, opened 2026-06-30
Per-read .option() storage credentials are dropped on path-based reads, causing Azure MSI fallback


sezruby:
### Summary

When reading a Lance dataset by path with storage credentials supplied via the
DataFrame reader `.option(...)` API, the credentials are silently dropped before the
dataset is opened. The native object store then falls back to its default credential
chain — for Azure, the IMDS / Managed Identity endpoint — and the read fails on any
environment without a managed identity.

### Repro

```scala
spark.read.format("lance")
  .option("azure_storage_sas_token", "<sas>")
  .option("azure_storage_account_name", "<account>")
  .load("abfss://<fs>@<account>.dfs.core.windows.net/path/to/ds.lance")
  .show()
```

No `spark.sql.catalog.*` configured, and the SAS is **not** present as an OS env var.

### Observed

```
LanceError(IO): Generic MicrosoftAzure error: Error performing token request:
Error performing GET http://169.254.169.254/metadata/identity/oauth2/token
  ?api-version=2019-08-01&resource=https%3A%2F%2Fstorage.azure.com ... after 3 retries
    at org.lance.Dataset.openNative(Native Method)
    at org.lance.spark.utils.Utils$OpenDatasetBuilder.build(Utils.java:...)
    at org.lance.spark.BaseLanceNamespaceSparkCatalog.loadTableFromPath(...)
```

### Root cause

`LanceDataSource` implements `SupportsCatalogOptions`. On a path-based read, Spark calls
`extractIdentifier(options)` and then `catalog.loadTable(identifier)` — and **only the
`Identifier` is forwarded; the per-read option map is not.**

- `extractIdentifier` builds `new LanceIdentifier(readOptions.getDatasetUri())`, keeping
  only the location and discarding the options.
- `loadTableFromPath` therefore rebuilds storage options from `catalogConfig` alone,
  which is empty when credentials were supplied per-read.
- The resulting credential-less `Dataset.open` falls through to object_store's default
  Azure credential chain → IMDS/MSI → the error above.

This is not Azure-specific — the same drop affects any per-`.option()` storage credential
(S3 keys, GCS, etc.) on the path-based read. It is masked if the credential also happens
to be present as an OS environment variable, since lance core reads those as a fallback.

### Scope / not covered by existing PRs

- #522 propagates **namespace/REST-vended** storage options (from `describeTable`) into
  read options, and renames the path-based calls to `createPathBasedReadOptions(...)`, but
[...]

PR #666, opened 2026-06-30
fix(spark): clarify auth errors when storage .option()s are dropped on path-based access

sezruby:
## What

Give a clear, actionable error when a **path-based** Lance read fails on storage authentication because credentials were supplied via `.option()` — instead of surfacing a misleading downstream auth error (e.g. an Azure MSI/IMDS token failure) far from the real cause.

## Why

Path-based access (`spark.read.format("lance").load("abfss://…")`) routes through `SupportsCatalogOptions`, which forwards only the table `Identifier` to the catalog's `loadTable` — never the per-read `.option()` map. Any storage credentials passed that way (e.g. an Azure SAS token) are therefore dropped before reaching the native object store, and the open falls back to the default credential chain. On clusters with no managed identity that surfaces as a confusing Azure MSI/IMDS token error:

```
LanceError(IO): Generic MicrosoftAzure error: Error performing token request:
Error performing GET http://169.254.../metadata/identity/oauth2/token ... after 3 retries
  at org.lance.Dataset.openNative(Native Method)
  at org.lance.spark.BaseLanceNamespaceSparkCatalog.loadTableFromPath(...)
```

Per the discussion here, per-read `.option()` credentials are **not** something we want to plumb through the `SupportsCatalogOptions` path — catalog-managed credentials are the supported way, consistent with other Spark connectors. Thanks @hamersaw for the direction. This PR drops the earlier credential-propagation approach and instead makes the failure self-explanatory.

## What changed

- `BaseLanceNamespaceSparkCatalog.loadTableFromPath`: when the native open fails with what looks like a storage authentication/authorization error, it's wrapped with a message that explains `.option()` credentials aren't applied to path-based access and points at the supported alternatives (catalog `storage.*` config, environment variables, or a managed identity). The original native error is preserved as the cause.
- Detection is best-effort and message-based (`looksLikeStorageAuthFailure`); a miss just yields today's raw error. Non-auth failures — including genuine not-found (`IllegalArgumentException` → `NoSuchTableException`) — are rethrown unchanged, so there is no behavior change for valid reads.

## Testing

- `PathAccessErrorTest` (base module, no cloud account required): the Azure MSI/IMDS message and common auth markers (`403`, `SignatureDoesNotMatch`, "not authorized") are detected, including nested causes; non-auth errors pass through unchanged; the enriched message names the URI, `.option(...)`, and the catalog `storage.*` alternative and preserves the cause.
- Existing `read` (16) and `write` (25) connector suites pass; `./mvnw` spotless + checkstyle clean on the touched modules.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -74,6 +74,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
 import java.util.regex.Pattern;
@@ -1742,10 +1743,79 @@ public abstract class BaseLanceNamespaceSparkCatalog
       tableProperties = dataset.getConfig();
     } catch (IllegalArgumentException e) {
       throw new NoSuchTableException(ident);
+    } catch (RuntimeException e) {
+      throw describePathAccessError(datasetUri, e);
     }
 
     return createDataset(
         readOptions, schema, null, null, null, false, fileFormatVersion, tableProperties, null);
+  }
+
+  // Substrings (matched case-insensitively) that indicate a native open failed on storage
+  // authentication/authorization rather than on a missing or corrupt dataset. Best-effort: a miss
+  // simply yields the raw native error, i.e. today's behavior.
+  private static final String[] STORAGE_AUTH_ERROR_MARKERS = {
+    "identity/oauth2/token",
+    "token request",
+    "managed identity",
+    "credential",
+    "authenticat",
+    "authoriz",
+    "forbidden",
+    "access denied",
+    "signaturedoesnotmatch",
+    "invalidaccesskeyid",
+    "403",
+  };
+
+  /**
+   * Returns {@code true} if {@code error} or any of its causes carries a message that looks like a
+   * storage authentication/authorization failure (e.g. a dropped SAS token falling back to the
+   * Azure MSI/IMDS credential chain). Best-effort and message-based, since the native object-store
+   * error is surfaced only as text.
+   */
+  static boolean looksLikeStorageAuthFailure(Throwable error) {
+    for (Throwable cause = error; cause != null; cause = cause.getCause()) {
+      String message = cause.getMessage();
+      if (message == null) {
+        continue;
+      }
+      String lower = message.toLowerCase(Locale.ROOT);
+      for (String marker : STORAGE_AUTH_ERROR_MARKERS) {
+        if (lower.contains(marker)) {
+          return true;
+        }
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Enriches a native open failure on path-based access with actionable guidance when it looks like
+   * a storage authentication error, and returns it unchanged otherwise.
+   *
+   * <p>Path-based access ({@code spark.read.format("lance").load(uri)}) routes through {@code
+   * SupportsCatalogOptions}, which forwards only the table identifier to the catalog — never the
+   * per-read {@code .option(...)} map. Storage credentials passed that way (e.g. an Azure SAS
+   * token) are therefore dropped before reaching the object store, and the open falls back to the
+   * default credential chain, surfacing a confusing error far from the real cause (e.g. an Azure
+   * MSI/IMDS token request failure). This points the user at the supported alternatives.
+   */
+  static RuntimeException describePathAccessError(String datasetUri, RuntimeException error) {
+    if (!looksLikeStorageAuthFailure(error)) {
+      return error;
+    }
+    return new IllegalStateException(
+        "Failed to open Lance dataset at "
+            + datasetUri
+            + " and the underlying error looks like a storage authentication failure. Note that "
+            + "storage credentials passed via .option(...) are not applied to path-based access: "
+            + "only the table identifier is forwarded to the catalog, so a SAS token or account "
+            + "key set that way is dropped before reaching the object store. Configure storage "
+            + "credentials on the catalog instead — e.g. spark.sql.catalog.<name>.storage.<key> — "
+            + "or supply them via environment variables or a managed identity. Underlying error: "
+            + error.getMessage(),
+        error);
   }
 
   public abstract LanceDataset createDataset(
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/PathAccessErrorTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/PathAccessErrorTest.java
@@ -1,0 +1,99 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark;
+
+import org.junit.jupiter.api.Test;
+
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+/**
+ * Unit tests for the path-based-access error enrichment in {@link BaseLanceNamespaceSparkCatalog}.
+ *
+ * <p>Storage credentials passed via {@code .option(...)} are dropped on path-based access (only the
+ * table identifier is forwarded to the catalog), so the native open falls back to the default
+ * credential chain and surfaces a confusing auth error far from the real cause. These tests cover
+ * the message-based detection and enrichment; no cloud account is required.
+ */
+public class PathAccessErrorTest {
+
+  private static final String URI = "abfss://fs@acct.dfs.core.windows.net/t.lance";
+
+  /** The Azure MSI/IMDS failure users actually hit when a SAS token is dropped. */
+  private static final String AZURE_MSI_ERROR =
+      "LanceError(IO): Generic MicrosoftAzure error: Error performing token request: "
+          + "Error performing GET http://169.254.169.254/metadata/identity/oauth2/token "
+          + "after 3 retries";
+
+  @Test
+  public void detectsAzureMsiTokenFailure() {
+    assertTrue(
+        BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(
+            new RuntimeException(AZURE_MSI_ERROR)));
+  }
+
+  @Test
+  public void detectsCommonAuthMarkers() {
+    assertTrue(
+        BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(
+            new RuntimeException("403 Forbidden")));
+    assertTrue(
+        BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(
+            new RuntimeException("<Code>SignatureDoesNotMatch</Code>")));
+    assertTrue(
+        BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(
+            new RuntimeException("The request is not authorized to perform this operation.")));
+  }
+
+  @Test
+  public void detectsAuthMarkerInNestedCause() {
+    RuntimeException wrapped =
+        new RuntimeException("failed to open dataset", new RuntimeException(AZURE_MSI_ERROR));
+    assertTrue(BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(wrapped));
+  }
+
+  @Test
+  public void ignoresNonAuthErrors() {
+    assertFalse(
+        BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(
+            new RuntimeException("Not a Lance dataset: missing _versions directory")));
+    assertFalse(BaseLanceNamespaceSparkCatalog.looksLikeStorageAuthFailure(new RuntimeException()));
+  }
+
+  @Test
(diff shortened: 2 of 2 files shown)


sezruby:
Gentle ping 🙂 this one's been sitting without a reviewer. It's green (all checks pass, spotless/checkstyle clean) and mergeable.

Quick recap: path-based reads that pass storage creds via `.option(...)` (e.g. an Azure SAS token) were silently dropped before the native open, so Azure fell back to Managed Identity (IMDS `169.254.169.254`) and failed on clusters without one. `SupportsCatalogOptions` forwards only the `Identifier` to `loadTable`, and `LanceIdentifier` didn't carry the options, so the fix threads them through (and covers `tableExistsAtPath` too). Verified with a real abfss read that fails on `main` and passes here. Fixes #665.

@geruh @LuciferYang @fangbo would any of you have a few minutes to take a look? Happy to adjust anything. Thanks!

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -167,6 +167,18 @@ private CreateTableSpec resolveCreateSpec(StructType schema, Map<String, String>
     return CreateTableSpec.resolve(schema, properties, catalogConfig.getFileFormatVersion());
   }
 
+  /**
+   * Per-read options carried on a path-based identifier (e.g. storage credentials supplied via the
+   * DataFrame {@code .option(...)} call). Empty for non-{@link LanceIdentifier} identifiers. These
+   * override catalog-level storage options so path-based access can supply credentials without a
+   * catalog being configured.
+   */
+  private static Map<String, String> getPathOptions(Identifier ident) {
```
LuciferYang: getPathOptions(ident) is only consumed on the read path (loadTableFromPath, tableExistsAtPath). Since extractIdentifier is shared by reads and writes, the other path-based sites that also receive the credential-carrying Identifier still pass Optional.empty() and use catalogConfig.getStorageOptions() only:
  - createTableAtPath (native write driven by readOptions.getStorageOptions())
  - stageCreateAtPath / stageReplaceAtPath / stageCreateOrReplaceAtPath (StagedCommitOptions.pathBased(catalogConfig.getStorageOptions(), ...))
  - dropTableAtPath (Dataset.drop(uri, catalogConfig.getStorageOptions()))
  - resolveIdentifier (reached on the path-based flow via alterTable; stageReplace early-returns to stageReplaceAtPath before it)
So .option()-supplied credentials (e.g. an Azure SAS) are dropped on the write/drop/alter path-based flows and the native open falls back to the default credential chain (Azure MSI). This is pre-existing behavior, not a regression from this PR. But this PR does introduce an intra-operation inconsistency: stageCreateOrReplaceAtPath now authenticates the existence probe (fixed tableExistsAtPath) with the SAS, then opens the same dataset without it — the probe passes and the open fails, which is hard to diagnose.

Please either thread getPathOptions(ident) into all path-based open sites in this PR (for lanceDatasetExists / registerExistingTable, which take a bare String location, pass the options down from the caller), or explicitly declare the write/drop/alter paths out of scope here and file a follow-up issue.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Utils.java
```diff
@@ -190,11 +191,45 @@ public static LanceSparkReadOptions createReadOptions(
...
+
+    // Seed with per-read options first so withCatalogDefaults treats catalog storage options as
+    // defaults underneath them (catalog options fill gaps; per-read options win on conflict).
+    // Drop the dataset-URI key ("path"): it is the URI delivery channel, not a storage option, and
+    // must not leak into the map forwarded to the native object store. Other recognized typed
+    // reader flags are promoted to builder fields by fromOptions; stripping them from the storage
+    // map is handled separately (see lance-format/lance-spark#520).
+    if (pathOptions != null && !pathOptions.isEmpty()) {
+      Map<String, String> seedOptions = new HashMap<>(pathOptions);
+      seedOptions.remove(LanceSparkReadOptions.CONFIG_DATASET_URI);
```
LuciferYang: This strips only the "path" key. builder.fromOptions(seedOptions) then stores the whole seed map into storageOptions (parseTypedFlags only reads keys, never removes them), so per-read typed flags like version / batch_size / block_size remain in the map forwarded to the native object store via getStorageOptions(). The comment above acknowledges the cleanup is deferred to #520, but this PR widens the leak surface through the new pathOptions channel. Fine to defer — please just call out the known limitation in the PR description (or strip the recognized typed keys here too).

lance-spark-3.5_2.12/src/test/java/org/lance/spark/AbfssSasReadIntegrationTest.java
```diff
@@ -0,0 +1,93 @@
...
+ * Integration test for path-based abfss reads authenticated with an Azure SAS token passed via the
+ * DataFrame {@code .option(...)} API. This reproduces the reported failure where a SAS supplied on
+ * the reader was dropped before reaching the native object store (the catalog {@code loadTable}
+ * path only received the {@link org.apache.spark.sql.connector.catalog.Identifier}).
+ *
+ * <p>Disabled unless both env vars are set, since it needs a real Azure account:
+ *
+ * <pre>
+ *   AZURE_ABFSS_URI       e.g. abfss://&lt;fs&gt;@&lt;account&gt;.dfs.core.windows.net/path/to/ds.lance
+ *   AZURE_STORAGE_SAS_TOKEN   the SAS query string (no leading '?')
```
LuciferYang: This javadoc block has two bugs that both make the regression guard silently no-op:
1. The env-var usage block (this line) documents AZURE_STORAGE_SAS_TOKEN, but the test is gated by @EnabledIfEnvironmentVariable(named="LANCE_IT_SAS") and reads System.getenv("LANCE_IT_SAS"). A developer following the doc exports AZURE_STORAGE_SAS_TOKEN, LANCE_IT_SAS stays unset, and the test is silently skipped — they think the guard ran when it did not. (This is exactly what the in-body comment "Do not simplify this back to AZURE_STORAGE_SAS_TOKEN" warns against.)
2. The "Run with" command uses -pl lance-spark-base_2.12, but this test lives in lance-spark-3.5_2.12, so that module contains no such test and -Dtest=AbfssSasReadIntegrationTest matches nothing and reports success with 0 tests run.
Please fix both: use LANCE_IT_SAS in the usage block and the run command, and change the module to -pl lance-spark-3.5_2.12.

sezruby:
Thanks for the careful review @LuciferYang! Addressed all three in 66a723f:

1. **Write/drop/alter paths dropping creds.** Agreed this shouldn't balloon the PR beyond the read path. Fixed the one intra-operation inconsistency this PR introduced: `stageCreateOrReplaceAtPath` now threads `getPathOptions(ident)` into both the open and the staged commit, so they use the same credentials as the (now SAS-authed) existence probe. The remaining path-based write/drop/alter sites (and the bare-`String` `lanceDatasetExists` / `registerExistingTable` helpers) are declared out of scope here and tracked in #683.

2. **Typed-flag leak via pathOptions.** Left the stripping to #520 as you suggested; called out the known limitation explicitly in the PR description (Scope and known limitations).

3. **Test javadoc bugs.** Fixed both: usage block and run command now use `LANCE_IT_SAS`, and the run command targets `-pl lance-spark-3.5_2.12`.

All checks green, spotless/checkstyle clean.

sezruby:
Thanks for the detailed review @LuciferYang. Pushed in 66a723f:

- **stageCreateOrReplaceAtPath** now threads the per-read `.option()` credentials into the open *and* the staged commit, so they match the existence probe (`tableExistsAtPath`) — no more "probe authenticates with the SAS, then the open falls back to the default chain." The remaining write/drop/alter path-based sites are out of scope for this read-path PR; happy to file a follow-up issue to track them.
- **Test javadoc + run command** now use `LANCE_IT_SAS` (not `AZURE_STORAGE_SAS_TOKEN`, which core's `with_env_azure()` fallback would mask) and the correct `lance-spark-3.5_2.12` module.
- The typed-flag leak (`version`/`batch_size`/`block_size` riding the shared options map) is the pre-existing #520 cleanup — left deferred, noting it now also flows through the new `pathOptions` channel.

CI is green across all Spark/Scala matrices. Mind another look when you have time? cc @hamersaw for maintainer visibility.

hamersaw:
@sezruby thanks for the PR! Looking into this a bit I have some concerns about piggybacking on the `LanceIdentifier` to pass credentials. It seems we're doing this because by implementing `SupportsCatalogOptions` means we only pass an identifier implementation (ie. `LanceIdentifier`) between the API and actual opening. In the background we rely on the catalog to manage credentials. So essentially means that auth through the `.option()` is unsupported. This is the precedent across all the OSS Spark connectors I can find. The alternatives are (1) auth through a catalog (even if basically a placeholder) (2) use env vars (ie. `AZURE_STORAGE_SAS_TOKEN`) or (3) instance identifies (ex. MSI / IMDS) - the native default credential chain.

Is there something specific about your use-case where we can't make one of ^^^ work? A code snippet for basically an ephemeral catalog could be something like:
```
spark = SparkSession.builder()
    .appName("abfss-sas-read-it")
    .master("local")
    // Register a catalog named "lance" so path-based .load() routes through it,
    // and put the SAS in its storage config — where SupportsCatalogOptions expects creds.
    .config("spark.sql.catalog.lance", "org.lance.spark.LanceNamespaceSparkCatalog")
    .config("spark.sql.catalog.lance.storage.azure_storage_sas_token", sas)
    .getOrCreate();
// ...
Dataset<Row> df = spark.read().format("lance").load(uri);   // no .option() creds
```

sezruby:
Thanks @hamersaw — fair framing, and you're right about the mechanics: `SupportsCatalogOptions` only carries the `Identifier` into `loadTable`, so per-read `.option()`s aren't delivered by the framework, and catalog-managed creds are the norm for catalog-routed connectors. No argument there.

The gap your three alternatives don't cover is **per-read / per-dataset credentials within a single session**:

- **Catalog config** is static — one SAS per catalog per `SparkSession`. Our case reads multiple Lance datasets living in different storage accounts/tenants (with short-lived SAS that rotate) in the *same* job; that would need a separate catalog per credential set, or reconfiguring the session between reads.
- **Env var** is process-global with the same limitation — and it's actually what *masked* this gap, since core's `with_env_azure()` picks up `AZURE_STORAGE_SAS_TOKEN` regardless of whether the connector propagated anything.
- **MSI/IMDS** — the target clusters have no managed identity, which is what surfaced the original failure.

So the ask isn't "auth without a catalog" in general — it's a per-read credential channel for the multi-tenant / no-managed-identity case, which mirrors the standard Spark idiom for storage creds on file sources (`.option("fs.azure...", …)`).

I'd split this into two questions:

1. **Should per-read creds be supported at all?** If serving that multi-tenant/no-MSI case isn't a direction we want, I'm happy to close this and document catalog-config as the supported path.
2. **If yes, what's the cleanest mechanism?** Carrying them on `LanceIdentifier` was the minimal way to work *within* `SupportsCatalogOptions`. I'm not attached to it — if the second credential channel is the concern, I'm open to other plumbing (e.g. delivering `.option()`s through the DSv2 `getTable` path for path-based reads). Happy to rework whichever way fits the connector's direction.

Which would you prefer? I'll adjust accordingly.

hamersaw:
@sezruby I guess I'm leaning towards saying we don't support per-read creds. This seems to be the precedent set by all other Spark connectors using `SupportsCatalogOptions`. Obviously, the goal is to get your workflow working. So I do think the correct question is whether any of these options will work for you. I think we could probably solve this a little quicker without AI responses.

sezruby:
@hamersaw Thanks, catalog-config works for our immediate case, so we'll use that for now. On whether one of the three always suffices:

  1. For registered tables, a namespace/REST catalog that vends per-table storageOptions (what #522 already propagates via describeTable) can give different creds on the same account that case is covered. The gap is path-based / ad-hoc reads: a bare `spark.read.format("lance").load("abfss://…")` has no catalog entry to hang creds on, and dynamic/short-lived SAS can't be pre-registered as catalogs or session config. Env var and MSI are account/identity-global, so they can't express it either.
  2. Lance's native object_store already accepts different creds per `Dataset.open` (it's not authority-cached), so the capability is there underneath — the connector dropping the .option() map before the open is the only barrier.
  3. Precedent is mixed, not uniform: Iceberg/Hudi don't take per-read storage .option()s (catalog/session-level), but BigQuery, Snowflake, Delta, and the built-in JDBC source do. So per-read creds aren't unusual. It's specific to the SupportsCatalogOptions routing.
  4. The failure is also misleading — instead of "storage options ignored," it surfaces as an Azure MSI/IMDS error, which sends you debugging the wrong layer:
```
  LanceError(IO): Generic MicrosoftAzure error: Error performing token request:
  Error performing GET http://IP/metadata/identity/oauth2/token
    ?api-version=2019-08-01&resource=https%3A%2F%2Fstorage.azure.com ... after 3 retries
      at org.lance.Dataset.openNative(Native Method)
      at org.lance.spark.BaseLanceNamespaceSparkCatalog.loadTableFromPath(...)
```

  So: catalog for now, but complex workloads read different datasets on the same account/container with distinct short-lived creds by path, and there's no way to express that through catalog/env/MSI. I'm flexible on mechanism, the identifier carry, or delivering path reads through the DSv2 getTable options path. Would you be open to supporting the per-read case or, at minimum, failing fast with a clear message instead of the MSI red herring?

--
This is written and reviewed by me — just got some help polishing and formatting 😅

hamersaw:
> Precedent is mixed, not uniform: Iceberg/Hudi don't take per-read storage .option()s (catalog/session-level), but BigQuery, Snowflake, Delta, and the built-in JDBC source do. So per-read creds aren't unusual. It's specific to the SupportsCatalogOptions routing.

I agree, it's the fact that we use `SupprotsCatalogOptions` that makes this difficult.

> Would you be open to supporting the per-read case or, at minimum, failing fast with a clear message instead of the MSI red herring?

If this can work using the existing catalog approach, that's the cleanest solution I think. A clear error message would certainly help this!

sezruby:
@hamersaw Went with the clear error message. On path-based reads, when the open fails with what looks like a storage auth error (the MSI/IMDS red herring you get when a SAS is dropped), it now wraps it — says `.option()` creds aren't applied on path-based access and points at catalog `storage.*` / env var / MSI. Original error kept as the cause.

Only touches the path-based branch, so catalog-cred failures are unchanged, and non-auth/not-found errors pass through as before. Added a unit test.

PR #667, opened 2026-07-01
fix: null LanceColumnarPartitionReader.fragmentReader before releasing

LuciferYang:
## Summary

Fix M3 from an internal audit: `LanceColumnarPartitionReader.next()` releases the previous fragment's scanner before opening the next one, but leaves the `fragmentReader` field pointing to the just-closed scanner. If `LanceFragmentColumnarBatchScanner.create(...)` then throws, Spark's task-completion listener calls `close()` on this reader, which re-closes the already-closed scanner and raises `ArrowArrayStream is already closed` (NPE from `ArrowArrayStream.memoryAddress()` on a second `release()` inside `ArrowArrayStreamReader.closeReadSource`, wrapped as IOException).

## Root cause

`lance-spark-base_2.12/.../read/LanceColumnarPartitionReader.java` **Before**:

```java
while (fragmentIndex < inputPartition.getLanceSplit().getFragments().size()) {
  if (fragmentReader != null) {
    fragmentReader.close();          // fragment 0 released
  }
  fragmentReader =                   // if this throws, field still points at
      LanceFragmentColumnarBatchScanner.create(...);   // the released scanner
  fragmentIndex++;
  ...
}
```

The `LanceFragmentScanner.create()` path can throw for a missing fragment (`getFragment == null` → `IllegalStateException`, wrapped in `RuntimeException`), a storage/IO error during dataset open, or any scan-option builder failure. Any of these leaves the field dangling.

The Lance JNI handles themselves are guarded and idempotent, but the enclosing `arrow-java` `ArrowArrayStream.release()` is not — its second call hits `checkNotNull(data, "ArrowArrayStream is already closed")` in `ArrowArrayStream.memoryAddress()`.

## Fix

Null-first in `next()` — release the field's reference before invoking `close()`, so the follow-up `close()` short-circuits:

```java
while (fragmentIndex < inputPartition.getLanceSplit().getFragments().size()) {
  // Null-first so if create(...) below throws, the subsequent close() sees fragmentReader ==
  // null and short-circuits, rather than re-closing the already-closed previous scanner and
  // raising `ArrowArrayStream is already closed`.
  if (fragmentReader != null) {
    LanceFragmentColumnarBatchScanner toClose = fragmentReader;
    fragmentReader = null;
    toClose.close();
  }
  fragmentReader = LanceFragmentColumnarBatchScanner.create(...);
  ...
}
```

Apply the same shape to `close()` so the reader is idempotent — `PartitionReader` extends `Closeable`, whose contract requires `close()` to have no effect on repeat calls (JDK `Closeable.close` Javadoc; `AutoCloseable.close` calls out this distinction explicitly).

## Tests

Two regression tests (`LanceColumnarPartitionReaderTest`):

- **`testCloseAfterMidIterationCreateFailure`** — split `[0, 999999]`. Drains fragment 0, then `next()` attempts to open fragment 999999 which throws. `close()` must be a no-op. Uses `assertThrows(RuntimeException.class, drain)` + `assertDoesNotThrow(close)`, with a `finally reader.close()` guard so an `AssertionError` mid-lambda still releases the Arrow allocator.

- **`testCloseIsIdempotent`** — single fragment, full drain, then two consecutive `close()` calls. Both must succeed.

Both fail on `origin/main` (`ArrowArrayStream is already closed` NPE) and pass with the fix.

## Verification

| Command | Result |
|---|---|
| `mvn -pl lance-spark-base_2.12 test -Dtest=LanceColumnarPartitionReaderTest` | ✅ 5/5 |
| `mvn -pl lance-spark-base_2.12 test` (base module) | ✅ 460, 0 fail |
| `make test` (default 3.5 / 2.12) | ✅ 988, 0 fail, 48 skipped |
| `make lint` (Checkstyle + Spotless, full matrix) | ✅ 15 modules SUCCESS |

Full Spark × Scala matrix (3.4/3.5 × 2.12/2.13, 4.0/4.1 × 2.13) delegated to CI.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
@@ -41,8 +41,13 @@ public class LanceColumnarPartitionReader implements PartitionReader<ColumnarBat
       return true;
     }
     while (fragmentIndex < inputPartition.getLanceSplit().getFragments().size()) {
+      // Null-first so if create(...) below throws, the subsequent close() sees fragmentReader ==
+      // null and short-circuits, rather than re-closing the already-closed previous scanner and
+      // raising `ArrowArrayStream is already closed`.
       if (fragmentReader != null) {
-        fragmentReader.close();
+        LanceFragmentColumnarBatchScanner toClose = fragmentReader;
+        fragmentReader = null;
+        toClose.close();
       }
       fragmentReader =
           LanceFragmentColumnarBatchScanner.create(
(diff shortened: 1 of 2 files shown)


LuciferYang:
cc @fangbo FYI

fangbo:
+1, Great Fix !

LuciferYang:
Thank you @fangbo

PR #675, opened 2026-07-03
fix: preserve timestamp timezones across schema roundtrips

yyzhao2025:
## Summary

This PR preserves Arrow timestamp timezones when schemas roundtrip through Spark metadata.

Before this change, an Arrow schema like `timestamp[us, Asia/Shanghai]` could lose its original timezone after:

- Arrow schema -> Spark `StructType`
- Spark `StructType` -> Arrow schema

In write paths that rebuild Arrow schema with a caller-provided timezone (for example `"UTC"`), this could change an existing field from `timestamp:us:Asia/Shanghai` to `timestamp:us:UTC` and fail schema validation during commit.

## Root cause

`LanceArrowUtils.fromArrowSchema` converted timezone-aware Arrow timestamps into Spark `TimestampType`, but did not preserve the original Arrow timezone anywhere.

Later, `LanceArrowUtils.toArrowSchema` rebuilt Arrow timestamps from Spark types using the provided `timeZoneId`, which could overwrite the original logical type.

## What this PR changes

- store the original Arrow timestamp timezone in Spark field metadata during `fromArrowSchema`
- restore that timezone during `toArrowSchema` when the metadata is present
- avoid writing this internal metadata key back into Arrow field metadata

## Regression coverage

This PR adds regression coverage for:

- Arrow schema roundtrip through Spark metadata
- `ADD COLUMNS` backfill
- `UPDATE COLUMNS` backfill
- overwrite write path
- staged commit for an existing table

## Compatibility

This change is backward compatible:

- it only affects Arrow timestamps that already carry a timezone
- existing Spark-originated schemas without this metadata keep their previous behavior
- timestamp types without a timezone are unchanged

## Testing

Verified locally:

- `./mvnw -pl lance-spark-base_2.12 -Dtest=LanceArrowUtilsTest -Dsurefire.failIfNoSpecifiedTests=false test`

Added native-backed regression tests for add/update backfill, overwrite, and staged commit flows as part of this PR.

Note: I could not fully execute the native-backed write-path tests on my local machine because the available local JNI/JDK setup was architecture-mismatched (`darwin-aarch64` native library vs `x86_64` JDK). The tests are included so they can run in CI / a matching environment.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
@@ -50,6 +50,7 @@ object LanceArrowUtils {
   val ARROW_DATE_MILLISECOND_KEY = DateMilliUtils.ARROW_DATE_MILLISECOND_KEY
   val ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY =
     FixedSizeBinaryUtils.ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY
+  val ARROW_TIMESTAMP_TIMEZONE_KEY = "lance.arrow.timestamp.timezone"
 
   // Namespaced keys used to embed child Spark Metadata on a parent StructField when the
   // child sits inside an ArrayType/MapType — Spark has no per-element metadata slot of its
@@ -249,6 +250,8 @@ object LanceArrowUtils {
         // Preserve FixedSizeBinary byte width so a subsequent write reproduces
         // FixedSizeBinary(n) instead of falling back to variable-length Binary.
         builder.putLong(ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY, fsb.getByteWidth.toLong)
+      case timestamp: ArrowType.Timestamp if timestamp.getTimezone != null =>
+        builder.putString(ARROW_TIMESTAMP_TIMEZONE_KEY, timestamp.getTimezone)
       case _ =>
     }
   }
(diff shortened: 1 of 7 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -50,6 +50,7 @@ object LanceArrowUtils {
   val ARROW_DATE_MILLISECOND_KEY = DateMilliUtils.ARROW_DATE_MILLISECOND_KEY
   val ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY =
     FixedSizeBinaryUtils.ARROW_FIXED_SIZE_BINARY_BYTE_WIDTH_KEY
+  val ARROW_TIMESTAMP_TIMEZONE_KEY = "lance.arrow.timestamp.timezone"
```
hamersaw, line 53: Can we update this to something like:
```
val LANCE_TIMESTAMP_TIMEZONE_KEY = "_lance.timestamp.timezone"
```

I think this is basically an ephemeral key to handle timezone passing in the Spark -> Arrow -> Spark and not traditional arrow metadata. Lance core already correctly handles timezone storage, so this is Spark connector specific and feels like we should be careful not to leak this into the metadata.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
```diff
@@ -342,7 +345,9 @@ object LanceArrowUtils {
         .readValue(metadata.json, classOf[java.util.LinkedHashMap[_, _]])
         .asScala
         .collect {
-          case (k, v) if !LANCE_INTERNAL_METADATA_KEYS.contains(k.toString) =>
+          case (k, v)
+              if !LANCE_INTERNAL_METADATA_KEYS.contains(k.toString) &&
+                k.toString != ARROW_TIMESTAMP_TIMEZONE_KEY =>
```
hamersaw, line 350: I think we should just add the `LANCE_TIMESTAMP_TIMEZONE_KEY` into `LANCE_INTERNAL_METADATA_KEYS` so we don't need this special check.

hamersaw (changes requested):
Thanks for the fix! Just some nit naming and then I think we can merge.

PR #697, opened 2026-07-16
fix: fix List child field name inconsistent between pylance and lance-spark

### LINKED ISSUE #687: Inconsistent List child field name between PyLance and lance-spark (`item` vs `element`) causes schema drift for `List<String>`
## Summary

There is an inconsistency in how Arrow `List` / `LargeList` child field names are generated across the Lance stack.

PyLance / core Lance code commonly uses `item` as the child field name for list types, while `lance-spark` writes Spark `ArrayType` back to Arrow using `element`.

For example, a `List<String>` field may appear as:

- PyLance / Lance-native schema: `list<item: string>`
- lance-spark round-tripped schema: `list<element: string>`

Semantically these schemas are equivalent in Arrow, but the child field name difference creates avoidable schema drift across engines and can surface in schema equality checks, debugging output, tests, metadata round-trips, and schema evolution logic that compares field names strictly.

## Evidence

In `lance-spark`, `ArrayType` is converted back to Arrow `List` using `"element"` as the child field name:

- [lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala](https://github.com/lance-format/lance-spark/blob/main/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala#L387-L398)

```scala
val fieldType = new FieldType(nullable, ArrowType.List.INSTANCE, null, meta.asJava)
new Field(
  name,
  fieldType,
  Seq(
    toArrowField(
      "element",
      elementType,
      containsNull,
      timeZoneId,
      elementMetadata,
      largeVarTypes)).asJava)
```

In Lance-native Rust code, list child fields are commonly created as `"item"`:

- [rust/lance-file/src/reader.rs](https://github.com/lance-format/lance/blob/main/rust/lance-file/src/reader.rs#L2622)

```rust
let categories_type = DataType::List(Arc::new(Field::new("item", DataType::Utf8, true)));
```

This `item` naming pattern also appears broadly across Lance core / Rust / JNI paths.

### Step 1: Create a dataset with PyLance

```python
import pyarrow as pa
import lance

uri = "/tmp/list_string_example.lance"

table = pa.table(
    {
        "id": pa.array([1, 2], type=pa.int64()),
        "tags": pa.array(
            [
                ["cat", "animal"],
                ["red", "round", "fruit"],
            ],
            type=pa.list_(pa.string()),
        ),
    }
)

ds = lance.write_dataset(table, uri)
print(ds.schema)
```

Expected schema shape from PyLance is typically:

```text
id: int64
tags: list<item: string>
```

### Step 2: Read and write the same dataset through lance-spark

After Spark reads the dataset and writes it back, the equivalent field may become:

```text
tags: list<element: string>
```

## Expected Behavior

One of the following should hold consistently across the stack:

1. All Lance components use the same canonical child field name for `List` / `LargeList` / `FixedSizeList`, preferably `item` to match the broader Arrow/Lance-native convention.
2. Schema comparison / schema evolution logic explicitly normalizes list child names so that `item` and `element` are treated as equivalent and do not create schema drift.

## Actual Behavior

Different components generate different child field names for the same logical type:

- Lance-native / PyLance: `item`
- lance-spark: `element`

## Why This Matters

Although Arrow list child names are usually not semantically important, the inconsistency still creates real friction:

- schema diffs are noisier than necessary
- round-trip tests can fail or require special handling
- debugging output is confusing across tools
- schema evolution / equality code may treat logically identical schemas as different
- downstream tools that serialize or compare the full Arrow schema may observe avoidable drift

## Suggested Direction

There are two reasonable options:

### Option A: Align lance-spark with the Lance-native convention

Change `lance-spark` to emit `item` instead of `element` when converting Spark `ArrayType` back to Arrow `List` / `FixedSizeList`.

Pros:

- matches the dominant naming convention in Lance core / PyLance / Rust
- reduces cross-engine schema drift

### Option B: Keep current output but normalize in comparisons

If `lance-spark` intentionally follows Spark-side conventions, then schema comparison / schema evolution code should normalize child field names for list-like types and treat `item` / `element` as equivalent.

Pros:

- less disruptive for Spark-specific code
- still avoids false schema conflicts

## Questions

- Is `element` in `lance-spark` an intentional Spark compatibility choice, or just a local convention?
- Is there already a preferred canonical child field name for list-like Arrow types in the broader Lance ecosystem?
- Would maintainers prefer alignment at schema generation time, or equivalence at comparison time?

fangbo:
Close https://github.com/lance-format/lance-spark/issues/687

## Why

This PR fixes schema drift caused by inconsistent `List` child field names between PyLance and lance-spark. In the reported case, PyLance used `item` while lance-spark rewrote the same list field with `element`, which could change the schema after an Arrow ↔ Spark roundtrip and break compatibility for existing datasets<cite-inline data-index="1" data-url="https://github.com/lance-format/lance-spark/pull/697" data-title="fix: fix List child field name inconsistent between pylance and lance-spark by fangbo · Pull Request #697 · lance-format/lance-spark · GitHub"></cite-inline>.

## How

This change preserves the original Arrow list child field name in Spark metadata and restores it when converting back to Arrow. It introduces `ListChildUtils`, stores the list child name under `_lance.list.child.name`, uses that metadata during writeback, and falls back to `item` when no metadata is present<cite-inline data-index="1" data-url="https://github.com/lance-format/lance-spark/pull/697" data-title="fix: fix List child field name inconsistent between pylance and lance-spark by fangbo · Pull Request #697 · lance-format/lance-spark · GitHub"></cite-inline>.

## Tests

The test coverage includes:
- preserving list child names when reading Arrow schema into Spark
- defaulting Spark `ArrayType` writes to `item` when no metadata exists
- writing back the stored child field name from metadata
- preserving nested `Array<Array<String>>` child names
- roundtripping nested list schemas without losing inner child field names

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/ListChildUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/ListChildUtils.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.utils;
+
+import org.apache.spark.sql.types.Metadata;
+
+/** Utility methods for preserving Arrow List child field names via Spark metadata. */
+public class ListChildUtils {
+
+  public static final String LANCE_LIST_CHILD_NAME_METADATA_KEY = "_lance.list.child.name";
+  public static final String LIST_CHILD_NAME_DEFAULT = "item";
+
+  /** Returns the persisted Arrow List child name, or the default if metadata is absent. */
+  public static String listChildName(Metadata metadata) {
+    if (metadata == null || !metadata.contains(LANCE_LIST_CHILD_NAME_METADATA_KEY)) {
+      return LIST_CHILD_NAME_DEFAULT;
+    }
+    return metadata.getString(LANCE_LIST_CHILD_NAME_METADATA_KEY);
+  }
+}
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/util/LanceArrowUtils.scala
@@ -260,7 +260,16 @@ object LanceArrowUtils {
         val children = field.getChildren
         if (!children.isEmpty) {
           val childField = children.get(0)
-          builder.putString(ListChildUtils.LANCE_LIST_CHILD_NAME_METADATA_KEY, childField.getName)
+          // Arrow permits unnamed List children; a null here would survive into the Metadata map
+          // and only blow up later when it is serialized to JSON.
+          val childName =
+            Option(childField.getName).getOrElse(ListChildUtils.LIST_CHILD_NAME_DEFAULT)
+          // Only record a non-default name. Writeback resolves an absent key to the same default,
+          // so stamping it would add nothing while putting an internal key on the Spark schema of
+          // every array column, breaking equality against an equivalent plain Spark schema.
+          if (childName != ListChildUtils.LIST_CHILD_NAME_DEFAULT) {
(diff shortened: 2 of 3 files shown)


hamersaw (approved):
Thanks for the fix! I made a small update because Arrow actually allows nullable field names. This should be impossible to reach under normal circumstances (ie. dataset written by Lance Spark / Lance Core), but I want to make sure we're safe against other Lance writers that could be incorrect. Let me know your thoughts and once you sign off lets get it merged!

fangbo:
> Thanks for the fix! I made a small update because Arrow actually allows nullable field names. This should be impossible to reach under normal circumstances (ie. dataset written by Lance Spark / Lance Core), but I want to make sure we're safe against other Lance writers that could be incorrect. Let me know your thoughts and once you sign off lets get it merged!

@hamersaw Thanks for your modification. I think it's sensible.

PR #700, opened 2026-07-17
test: add end-to-end append test for legacy `element`-child datasets

### LINKED ISSUE #687: Inconsistent List child field name between PyLance and lance-spark (`item` vs `element`) causes schema drift for `List<String>`

### LINKED ISSUE #699: Follow-up to #697: add end-to-end append test for legacy `element`-child datasets
### Background

PR #697 (closes #687) fixed the `item` vs `element` List child field name inconsistency between PyLance/lance-core and lance-spark: it stores a non-default Arrow list child name in Spark `StructField` metadata under `_lance.list.child.name` (via the new `ListChildUtils`), restores it on writeback, and defaults to `item` to match lance-core. The design is sound and the driver→executor write chain that keeps append working against a legacy `element`-child dataset is correct.

This is a follow-up to capture two gaps that a review of #697 surfaced. Neither blocks the merged fix; both are about hardening.

### 1. No end-to-end test for appending into a legacy `element`-child dataset (MEDIUM)

The core guarantee of #697 is that appending into an existing dataset whose List child is named `element` no longer fails lance-core's by-name schema validation. That guarantee depends on three hops:

1. `BaseLanceNamespaceSparkCatalog.loadTableInternal` derives the table schema via `LanceArrowUtils.fromArrowSchema(dataset.getSchema())` (`BaseLanceNamespaceSparkCatalog.java:1660` / `:1740`), stamping `_lance.list.child.name=element` onto the array column's Spark metadata.
2. `LanceDataset.newWriteBuilder` passes the table's `sparkSchema` (not `logicalWriteInfo.schema()`) into `SparkWrite.SparkWriteBuilder` (`LanceDataset.java:399-401`).
3. The executor write buffers (`QueuedArrowBatchWriteBuffer.java:168`, `SemaphoreArrowBatchWriteBuffer.java:144`) convert that `sparkSchema` back through `LanceArrowUtils.toArrowSchema(...)`, where `elementName = ListChildUtils.listChildName(metadata)` (`LanceArrowUtils.scala:368`) restores `element`, satisfying `validate_schema` (lance-core `field.rs` `compare_with_options` → `schema.rs` `compare_fields`, by-name).

All three hops are correct, but the 8 new tests in `LanceArrowUtilsSuite.scala:500-625` are all in-process `toArrowSchema` / `fromArrowSchema` roundtrips with no `SparkSession` — they only exercise hop 1 (metadata record/restore). No test actually writes into an `element`-child dataset and lets lance-core run append validation. The existing end-to-end array test (`BaseSparkDataTypeRoundtripTest.testArrayOfStringRoundtrip`) builds a fresh `item`-child table and self-reads/writes, so it doesn't cover the #687 scenario either.

**Risk:** if a later refactor changes the schema source of the executor write buffers (e.g. switches to `logicalWriteInfo.schema()`, or a DataFrame transform drops the array column's metadata before the write), append-compat silently regresses to writing `item` and fails lance-core with `element != item` — and the current test suite stays green because nothing covers the driver→executor write path.

**Suggested fix:** add an end-to-end test that creates (externally / by hand) a Lance dataset whose List child is named `element`, then uses Spark to append a batch and asserts the append succeeds with the correct row count. This makes any change to the `newWriteBuilder` / executor-buffer schema source turn the suite red instead of surfacing on a real legacy table in production.

### 2. Silent fallback to `item` when the child-name key is absent — path-based write edge (LOW)

`ListChildUtils.listChildName` returns the default `item` when metadata is `null` or lacks `_lance.list.child.name` (`ListChildUtils.java:26`). On the normal catalog-routed append path this is fine — the table schema flows from `fromArrowSchema` and carries the key. But `LanceDataSource implements SupportsCatalogOptions` with `supportsExternalMetadata()` returning `true` (`LanceDataSource.java:54` / `:99`), and its `getTable(StructType schema, …)` (`:84-86`) adopts the caller-provided schema directly.

**Risk:** a path-based write (`df.write.format("lance").save(path)`) into a legacy `element`-child table, where the incoming DataFrame's array-column metadata was stripped by an upstream transform, would fall back to writing `item` and fail lance-core's by-name check. This is not reachable via the documented append APIs (schema flows from the table side via `fromArrowSchema`, carrying the key), so it's an edge/theoretical risk.

**Suggested fix:** no code change needed on its own; if the end-to-end test from item 1 also covers the path-based `.save()` branch, this edge is guarded as a side effect.

---

_Filed as a follow-up to the review of #697. Anchors reference the merged code (PR #697 head `a2866b4`)._

LuciferYang:
## What

Follow-up to #697 (which closed #687). #697 fixed the `item` vs `element` Arrow List child field name inconsistency between PyLance/lance-core and lance-spark: it records a non-default list child name in Spark `StructField` metadata under `_lance.list.child.name` on read and restores it on writeback, defaulting to `item` to match lance-core.

The design is sound, but all 8 of #697's new tests are in-process `toArrowSchema` / `fromArrowSchema` roundtrips (no `SparkSession`) — they cover the metadata record/restore step but nothing exercises the full driver→executor write path that actually keeps `append` working against a dataset whose List child is named `element` (i.e. the pre-#697 lance-spark convention, or a dataset written by an older lance-spark / a hand-built producer). That guarantee is the whole point of #687, so it deserves an end-to-end regression.

## Change

Adds `BaseSparkConnectorWriteTest.appendIntoLegacyElementChildDataset` (test-only, no production changes):

1. Seeds a dataset directly through the Lance JNI (`Dataset.create`) with a List child explicitly named `element`, bypassing the Spark write path (which names it `item` today), and asserts the seed really uses `element` so the test can't go vacuous.
2. Appends a Spark DataFrame via the catalog-routed `writeTo(...).append()` path and asserts the row count.
3. Appends again via the path-based `df.write.format("lance").mode(Append).save(path)` path — covering the `getTable(StructType, …)` branch that adopts the caller-provided schema (the LOW-severity edge from the review).
4. Asserts the stored List child name is still `element` after both appends.

## Test plan

- `make test SPARK_VERSION=3.5 SCALA_VERSION=2.12` — new test passes; full `SparkConnectorWriteTest` (26 tests) green.
- **Negative check:** temporarily reverting the child-name restore in `LanceArrowUtils.toArrowField` (forcing `elementName = "item"`) makes the new test fail with lance-core's `Append with different schema: tags had mismatched children: missing=[tags.element], unexpected=[tags.item]` — confirming the test guards the #697 fix rather than passing trivially.

Closes #699.

### DIFF
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/write/BaseSparkConnectorWriteTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/write/BaseSparkConnectorWriteTest.java
@@ -14,12 +14,19 @@
 package org.lance.spark.write;
 
 import org.lance.Version;
+import org.lance.WriteParams;
 import org.lance.spark.LanceDataSource;
 import org.lance.spark.LanceRuntime;
 import org.lance.spark.LanceSparkReadOptions;
 import org.lance.spark.LanceSparkWriteOptions;
 import org.lance.spark.TestUtils;
 
+import org.apache.arrow.memory.BufferAllocator;
+import org.apache.arrow.memory.RootAllocator;
+import org.apache.arrow.vector.types.pojo.ArrowType;
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.arrow.vector.types.pojo.FieldType;
+import org.apache.arrow.vector.types.pojo.Schema;
 import org.apache.spark.sql.Dataset;
 import org.apache.spark.sql.Row;
 import org.apache.spark.sql.RowFactory;
@@ -40,6 +47,7 @@ import java.io.File;
 import java.nio.file.Path;
 import java.time.format.DateTimeFormatter;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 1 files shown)


hamersaw (approved):
Probably a good test to have, thanks!

LuciferYang:
Thank you @hamersaw

PR #701, opened 2026-07-17
feat: build FTS indexes as independent segments

ddupg:
## Background

Distributed FTS builds currently give every fragment task the same caller-provided UUID. Executors write partial indexes into that shared physical index, and the driver calls `mergeIndexMetadata` before committing it. This differs from the independent segment workflow in Lance and allows retries or speculative attempts to target the same physical index directory.

## Implementation

- Batch fragments according to `num_segments` and let each Spark task build one uncommitted FTS segment with its own Lance-generated UUID.
- Publish the task-produced segments in one driver-side `commitExistingIndexSegments` transaction, removing the FTS dependency on a shared UUID and `mergeIndexMetadata`.
- Add integration coverage for distinct segment UUIDs, complete non-overlapping fragment coverage, consistent index details, and term and phrase queries across segments.
- Document the `num_segments` option and the FTS physical segment build workflow.

## Validation

- `./mvnw test -pl lance-spark-3.5_2.13 -Dtest=AddIndexTest#testCreateIndexFailureLeavesTableUsable+testCreateFtsSegmentIndexSupportsTermAndPhraseQueries`
- `./mvnw spotless:check -pl lance-spark-3.5_2.13 -am`

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -36,8 +36,7 @@ import org.lance.spark.arrow.LanceArrowWriter
 import org.lance.spark.utils.{CloseableUtil, FieldPathUtils, Utils}
 import org.lance.spark.write.SingleBatchArrowReader
 
-import java.time.Instant
-import java.util.{Collections, Locale, Optional, UUID}
+import java.util.{Collections, Locale, UUID}
 
 import scala.collection.JavaConverters._
 import scala.reflect.ClassTag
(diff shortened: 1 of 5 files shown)


ddupg:
@LuciferYang @hamersaw @fangbo Could help review this please? thanks

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -959,6 +770,125 @@ case class ZonemapIndexTask(
...
+      case e: Exception =>
+        throw new RuntimeException(
+          "FTS segment build failed. Uncommitted segments are not visible to readers and " +
+            "will not affect query correctness.",
+          e)
+    }
+  }
+
+  // Intentionally mirrors the existing Zonemap grouping semantics without refactoring that path.
+  private def batchFragments(
```
fangbo: This is duplicated with `ZoneMapIndexJob.batchFragments`. Could this method is refactored to a helper util ?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -959,6 +770,125 @@ case class ZonemapIndexTask(
   }
 }
 
+/**
+ * Builds FTS index segments on fragment batches. Each task attempt creates an independent,
+ * uncommitted segment whose UUID is generated by Lance.
+ */
+class FtsIndexJob(
```
fangbo: `ZoneMapIndexJob` and `FtsIndexJob` is very similar. Could these two classes are abstracted to a `SegmentedIndexJob` ?

ddupg: @fangbo Thanks for the suggestion. I agree that these scalar segment-index jobs should share a common build abstraction. I’m addressing that in [#740](https://github.com/lance-format/lance-spark/pull/740), which consolidates the scalar segment build flow and extends it to more index types.

ddupg:
The current CI failure is caused by a Lance core query bug in fragmented FTS scans: when an FTS index has multiple physical segments, the prefilter coverage is derived from only one segment's fragment bitmap. As a result, matches from other segments can be filtered out.

This is unrelated to the distributed index-build implementation in this PR, but this PR makes it possible to create multiple physical FTS segments and therefore exposes the issue in the catalog-only/local query fallback tests.

I have opened lance-format/lance#8211 to fix the upstream issue. It may be best to wait for that fix to land and for Lance Spark to consume the updated Lance core version before moving this PR forward.

PR #714, opened 2026-07-22
fix: merge write-time storage options into staged commits

### LINKED ISSUE #712: CREATE/REPLACE (staged) commits drop write-time storage options
# CREATE/REPLACE (staged) commits drop write-time storage options

## Problem

Storage credentials supplied via `.option(...)` on a `CREATE TABLE AS SELECT` /
`REPLACE TABLE AS SELECT` are used for the executor-side fragment writes but never reach the
final commit. The commit falls back to the default credential chain (e.g. Azure MSI/IMDS) and
fails on any environment without one.

Plain append/overwrite to an existing table does not have this problem.

## Root cause

`StagedCommit` freezes `storageOptions` in its constructor, from whatever was known at *stage*
time (`stageCreate`/`stageCreateAtPath`, called before the write happens):

```java
// StagedCommit.java
this.storageOptions = options.getStorageOptions();   // set once, no setter exists
```

`LanceBatchWrite.commit(...)` runs later, after the write, and is where write-time
`.option(...)` values are actually available (via `writeOptions.getStorageOptions()`). Its
non-staged branch merges them in before committing — its staged branch doesn't:

```java
if (stagedCommit != null) {
  stagedCommit.setFragments(fragments);
  stagedCommit.setSchema(arrowSchema);
  // storage options are never touched here
} else {
  CommitBuilder commitBuilder =
      new CommitBuilder(ds)
          .writeParams(
              LanceRuntime.mergeStorageOptions(
                  writeOptions.getStorageOptions(), initialStorageOptions));
  ...
}
```

`StagedCommit` has no setter for `storageOptions`, so there's no way to get the write-time
options into it after construction.

## Repro

```scala
spark.range(10).toDF("id")
  .writeTo("lance.`s3://bucket/t.lance`")
  .using("lance")
  .option("access_key_id", "...")
  .option("secret_access_key", "...")
  .option("endpoint", "http://127.0.0.1:9000")
  .create()
```

Fragment writes succeed; the manifest commit fails (or authenticates with the wrong
credentials), because `CommitBuilder.writeParams(storageOptions)` uses the value frozen at stage
time, before `.option(...)` was consulted.

## Precedent

`StagedCommit`/`LanceBatchWrite.commit()` has had this kind of gap before:
- **#443** fixed `use_large_var_types` being silently dropped on the `createOrReplace()` /
  staged-commit path — a different option, same two files.
- **#616** fixed exactly this — missing storage-option merge before commit — in five places
  (`LanceBatchWrite`'s non-staged branch, both backfill writers, `SparkPositionDeltaWrite` x2).
  It didn't touch `StagedCommit`/the staged branch. This is the missing sixth site.

## Not a duplicate of #683

#683 is about `extractIdentifier`/`LanceIdentifier` dropping per-`.option()` values used to
*resolve* the table before `stageCreate*` runs. This issue is about `newWriteBuilder`'s
write-time options, consumed later in `LanceBatchWrite.commit()`. Different Spark API, different
point in the pipeline — fixing #683 doesn't touch `StagedCommit`/`LanceBatchWrite.commit()`, and
fixing this doesn't touch `extractIdentifier`. Both are needed.

## Possible direction

Give `StagedCommit` a way to receive options after construction, and call it from
`LanceBatchWrite.commit()`'s staged branch the same way the non-staged branch already merges
them:

```java
// StagedCommit.java
public void mergeStorageOptions(Map<String, String> extra) {
  if (extra != null && !extra.isEmpty()) {
    this.storageOptions.putAll(extra);
  }
}
```

```java
// LanceBatchWrite.java, staged branch of commit(...)
stagedCommit.mergeStorageOptions(
    LanceRuntime.mergeStorageOptions(
        writeOptions.getStorageOptions(), initialStorageOptions));
```

One thing to watch: `StagedCommit`'s constructor currently aliases the map from
`StagedCommitOptions` directly, and for path-based staged creates that's often
`catalogConfig.getStorageOptions()` — `Collections.unmodifiableMap(...)`. `putAll` on it throws.
Whatever the fix, `StagedCommit` would need its own copy of the map before mutating it.

ivscheianu:

Fixes #712.

`StagedCommit` captures storage options once at stage-create time from the catalog's static config. For path-based staged creates/replaces (no namespace), that is the only storage-options source available at that point, so write-time `.option(...)` values and namespace-vended credentials obtained afterward were silently dropped from the final manifest commit, falling back to ambient credential discovery instead.

This PR:

- Adds `StagedCommit.mergeStorageOptions(Map<String, String>)` to layer in additional options after construction.
- Calls it from `LanceBatchWrite.commit()`'s staged branch with `LanceRuntime.mergeStorageOptions(writeOptions.getStorageOptions(), initialStorageOptions)` — the same merge and precedence the non-staged branch and `LanceSparkWriteOptions.toWriteParams()` (executor-side fragment writes) already use.
- Makes `StagedCommit`'s constructor defensively copy the incoming storage options map, since callers (e.g. `LanceSparkCatalogConfig.getStorageOptions()`) can hand it an unmodifiable map, which would otherwise make the new merge throw `UnsupportedOperationException` on every real path-based staged create.

All six staged entry points (`stageCreate[AtPath]`, `stageReplace[AtPath]`, `stageCreateOrReplace[AtPath]`) funnel through this single `LanceBatchWrite.commit()` staged branch, so the fix closes the gap uniformly without touching call sites individually. `AddColumnsBackfillBatchWrite`/`UpdateColumnsBackfillBatchWrite`/`SparkPositionDeltaWrite` are unaffected — they never use `StagedCommit` and already perform their own correct merge.

## Are these changes tested?

- New tests in `StagedCommitTest`: `mergeStorageOptions` adds new keys, overrides existing keys on conflict, is a no-op for null/empty input, does not mutate the caller's map, and accepts an unmodifiable base map (regression test for the `UnsupportedOperationException` risk above).
- New tests in `LanceBatchWriteTest`: `LanceBatchWrite.commit()` merges `initialStorageOptions` into a `StagedCommit` that started with empty options (simulating `stageCreateAtPath`), including a full round-trip through an actual dataset commit; and a separate test confirming write-time-only options survive the merge while `initialStorageOptions` still wins on key conflict, matching `LanceRuntime.mergeStorageOptions`'s documented precedence.
- `./mvnw test -pl lance-spark-3.5_2.12 -am` — 1238 tests, 0 failures, 0 errors.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
@@ -178,6 +178,14 @@ public class LanceBatchWrite implements BatchWrite {
       if (enableStableRowIds != null) {
         stagedCommit.setEnableStableRowIds(enableStableRowIds);
       }
+      // StagedCommit.storageOptions was captured at stage-create time from the catalog's static
+      // config only; it does not include write-time options or namespace-vended credentials
+      // obtained afterward (e.g. describeTable()/declareTable() results in initialStorageOptions).
+      // Merge both in now so the manifest commit below authenticates with them instead of falling
+      // back to ambient credential discovery. Same merge semantics as the non-staged branch below.
+      stagedCommit.mergeStorageOptions(
+          LanceRuntime.mergeStorageOptions(
+              writeOptions.getStorageOptions(), initialStorageOptions));
     } else {
       // For non-staged tables, commit immediately
       long version =
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
@@ -178,11 +178,9 @@ public class LanceBatchWrite implements BatchWrite {
       if (enableStableRowIds != null) {
         stagedCommit.setEnableStableRowIds(enableStableRowIds);
       }
-      // StagedCommit.storageOptions was captured at stage-create time from the catalog's static
-      // config only; it does not include write-time options or namespace-vended credentials
-      // obtained afterward (e.g. describeTable()/declareTable() results in initialStorageOptions).
-      // Merge both in now so the manifest commit below authenticates with them instead of falling
-      // back to ambient credential discovery. Same merge semantics as the non-staged branch below.
+      // For a path-based staged create, StagedCommit only has the catalog's static storage
+      // options at this point. Merge in write-time and namespace-vended options now so
+      // StagedCommit.commit() uses them. Mirrors the non-staged merge below.
       stagedCommit.mergeStorageOptions(
           LanceRuntime.mergeStorageOptions(
               writeOptions.getStorageOptions(), initialStorageOptions));
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/write/StagedCommit.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/write/StagedCommit.java
@@ -31,6 +31,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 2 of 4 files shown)


ivscheianu:
Hello @hamersaw, small one. I'd appreciate it if you could have a look when you get a chance. Thanks a lot!

hamersaw (approved):
Thanks for the PR! Looks like a good fix. I was trying to figure my way around resetting the catalog configuration on this twice, because IIUC the options that we pass to `mergeStorageOptions` will already exist in the base options. They just include additional per-write options as well. I don't think there's a sane way to get around this.

PR #720, opened 2026-07-24
fix: preserve case-sensitive option keys in BaseLanceNamespaceSparkCatalog

### LINKED ISSUE #717: BaseLanceNamespaceSparkCatalog passes lowercased option keys to LanceNamespace.connect()
## Description

`CaseInsensitiveStringMap` is Spark's wrapper that lowercases all config keys. In `BaseLanceNamespaceSparkCatalog.initialize()`, there are two places extracting options from it:

| Line | Code | Result |
|------|------|--------|
| 225 | `new HashMap<>(options.asCaseSensitiveMap())` | ✅ Keys preserve original case |
| 254 | `new HashMap<>(options)` | ❌ Keys are lowercased |

Line 254 produces `namespaceOptions` with all-lowercase keys, which is then passed to `LanceNamespace.connect()` and eventually to the storage backend. Any backend that expects case-sensitive config keys will fail to find them.

## Fix

One-line change at line 254:

```diff
- Map<String, String> namespaceOptions = new HashMap<>(options);
+ Map<String, String> namespaceOptions = new HashMap<>(options.asCaseSensitiveMap());
```

Other paths (`LanceDataset`, `LanceDataSource`) already correctly use `asCaseSensitiveMap()`.

Yohahaha:
CaseInsensitiveStringMap lowercases all keys when used directly, causing case-sensitive storage backend configs to fail. Use asCaseSensitiveMap() to preserve original key casing, consistent with other code paths.

Closes #717

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
@@ -234,7 +234,7 @@ public abstract class BaseLanceNamespaceSparkCatalog
     }
 
     // Initialize the namespace with proper configuration
-    Map<String, String> namespaceOptions = new HashMap<>(options);
+    Map<String, String> namespaceOptions = new HashMap<>(options.asCaseSensitiveMap());
 
     // Save namespace impl and properties for serialization to workers
     this.namespaceImpl = impl;
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseLanceNamespaceSparkCatalogInitializeTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseLanceNamespaceSparkCatalogInitializeTest.java
@@ -1,0 +1,96 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark;
+
+import org.lance.memwal.ShardingSpec;
+import org.lance.spark.write.StagedCommit;
+
+import org.apache.spark.sql.types.StructType;
+import org.apache.spark.sql.util.CaseInsensitiveStringMap;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+import java.nio.file.Path;
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+public class BaseLanceNamespaceSparkCatalogInitializeTest {
+
+  @TempDir private Path tempDir;
+
+  @Test
+  public void testInitializePreservesCaseSensitiveOptionKeys() {
+    Map<String, String> rawOptions = new HashMap<>();
+    rawOptions.put("impl", "dir");
+    rawOptions.put("root", tempDir.toString());
+    rawOptions.put("MyMixedCaseKey", "value1");
+    rawOptions.put("ALLCAPS_KEY", "value2");
+    rawOptions.put("camelCaseOption", "value3");
+
+    CaseInsensitiveStringMap options = new CaseInsensitiveStringMap(rawOptions);
+
+    TestCatalog catalog = new TestCatalog();
+    catalog.initialize("test", options);
+
+    Map<String, String> properties = catalog.getNamespaceProperties();
+
+    assertTrue(
+        properties.containsKey("MyMixedCaseKey"),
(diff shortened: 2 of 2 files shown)


Yohahaha:
@LuciferYang @ivscheianu could you help review this PR? thank you!

ivscheianu:
> @LuciferYang @ivscheianu could you help review this PR? thank you!

hey @Yohahaha, thanks for the fix! Looks good to me, opened a PR targeting this branch to also add a unit test, just to cover this case: https://github.com/Yohahaha/lance-spark/pull/1

Feel free to merge it if you think it's fine.

cc: @hamersaw, probably good to have a look on this

lance-spark-base_2.12/src/main/java/org/lance/spark/BaseLanceNamespaceSparkCatalog.java
```diff
@@ -234,7 +234,7 @@ public void initialize(String name, CaseInsensitiveStringMap options) {
     }
 
     // Initialize the namespace with proper configuration
-    Map<String, String> namespaceOptions = new HashMap<>(options);
+    Map<String, String> namespaceOptions = new HashMap<>(options.asCaseSensitiveMap());
```
LuciferYang, line 237: I'm on board with the direction: `namespaceOptions` and `storageOptions` at line 208 should have been the same key shape all along. What I want to check is that this drops the normalization wholesale instead of keeping it for the keys that actually need case preserved. Two things. First, the backend matches keys exactly (`properties.get("root")` and `k.strip_prefix("storage.")` in `dir.rs`, `strip_prefix("headers.")` in `rest.rs`). So `ROOT=/tmp` now fails with `Missing required property 'root'`, wrapped in `Failed to construct namespace impl …`, while `Headers.x-api-key`, `credential_vendor.Enabled` and `Storage.region` are dropped with no error at all. Case in the suffix alone is still fine: `storage.Region` works because both `as_s3_options` and opendal lowercase the key again. Second, this method still reads `impl`, `parent` and `single_level_ns` through `options.get`, which is case-insensitive, so you can end up with the catalog accepting a config that the backend then reports as missing. My preference is to keep the behavior this PR has and add two small things: a warn when a key differs from its canonical form only by case, scoped to the keys the connector owns (`impl`, `parent`, `parent_delimiter`, `single_level_ns`) plus the first-party prefixes, and a note in `docs/src/config.md` that keys are case-sensitive. I was going to suggest keeping the original keys and adding a lowercase copy, but I tested it and it does not work. opendal's `Configurator::from_iter` lowercases keys before handing them to the serde-derived config, and a duplicate field comes back as `ConfigInvalid`. On 0.57.0, the version this PR pins, passing both `Region` and `region` gives ``duplicate field `region` ``. OSS and TOS pass the user's keys straight into `from_iter` (`options.clone()` at `oss.rs:68` and `tos.rs:74`), and S3/GCS/Azure do the same under `use_opendal=true` (`aws.rs:122`), so a doubled map turns keys that work today into hard failures, for example `storage.Endpoint` on S3 with `use_opendal=true`. COS, GooseFS and HuggingFace always go through opendal but build their own canonical config map and read exact keys, so they are unaffected. That route needs the duplicate-field problem solved first, or narrowing down to normalizing only the suffixes under the first-party prefixes (`storage.`, `credential_vendor.`, `headers.`, `header.`, `tls.`).

LuciferYang: On the other hand, keys in `docs/src/config.md` now have to be copied with their exact case, but the docs don't say so. `storage.*` at `:70`, `root` at `:126` and `headers.*` at `:241` each show one spelling only. When it goes wrong the error is `Missing required property 'root'`, and someone looking at their own `ROOT` won't connect that to casing. A sentence in the docs plus the warn scoped to connector-owned keys should cover it. Worth a look at `pushDownFilters` and `topN_push_down` in `select.md:245-246` while you're there: they're the two documented keys whose canonical form isn't all lowercase, they go through exact `containsKey`, and getting the case wrong just silently disables them. Both were already case-sensitive before this change, so that part is a pre-existing trap rather than something this PR introduces.

LuciferYang (commented):
https://github.com/lance-format/lance-spark/blob/f0afd08275bbf7bb22b33cee8987097de745214b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/catalyst/optimizer/LanceBlobSourceContextRule.scala#L58

There's one more copy of this pattern left at `LanceBlobSourceContextRule.scala:58`: `new java.util.HashMap[String, String](relation.options)` normalizes the keys the same way, then line 60 wraps them back into a `CaseInsensitiveStringMap` and writes them onto the relation. `AbstractLanceFtsPredicateRule.scala:96` in the same package already uses `rel.options.asCaseSensitiveMap()`, so the two differ by that one call. The blast radius is smaller than the catalog case: it needs a blob target, a relation carrying options, and a user key with uppercase in it, and the `V2Writes.mergeOptions` assert only exists on Spark 4.x. Given one-PR-one-concern, a separate issue seems cleaner. Want me to file one?

Yohahaha (commented):
Thanks @LuciferYang! Created #737 to track the same bug in `LanceBlobSourceContextRule` separately.

hamersaw:
@Yohahaha is there any specific use-case that is failing right now? It looks like lance core is inconsistent on applying case sensitivity to options. Without a compelling reason this feels like something we should just leave to lance core to handle IMO.

Yohahaha:
> @Yohahaha is there any specific use-case that is failing right now? It looks like lance core is inconsistent on applying case sensitivity to options. Without a compelling reason this feels like something we should just leave to lance core to handle IMO.

our internal storage backend needs all the storage options must be case sensitive, but lance spark pass case in-sensitive options when call namespace#connect.
I have described the case in issue #717

PR #724, opened 2026-07-24
fix: fall back to local scan for FTS on catalog-only namespaces

wjones127:
Previously, a full-text query was routed to the namespace `queryTable` endpoint whenever a namespace was configured (`LanceScanBuilder.build()` gated only on `namespaceImpl != null`). `LanceNamespace.queryTable` is a default interface method that throws, and catalog-only implementations such as Glue, Hive, and Iceberg never override it, so `lance_match`, `lance_match_phrase`, and `lance_multi_match` failed with `Not supported: queryTable`. `DirectoryNamespace` implements it natively, which is why only the Glue job failed:

https://github.com/lance-format/lance-spark/actions/runs/29837726315/job/88658279987

```
UnsupportedOperationException{code=0, message='Not supported: queryTable'}
	at org.lance.namespace.LanceNamespace.queryTable(LanceNamespace.java:391)
	at org.lance.spark.search.LanceSearchColumnarPartitionReader.openArrowReader(LanceSearchColumnarPartitionReader.java:87)
```

The routing is now gated on whether the namespace implementation actually overrides `queryTable` (`LanceRuntime.supportsQueryTable`, cached per impl); otherwise the query falls through to the local per-fragment scan, which pushes the `FullTextQuery` into `ScanOptions`.

## Tests

- `FtsCatalogOnlyNamespaceTest` — a namespace that delegates every operation the connector uses to `DirectoryNamespace` but leaves `queryTable` unimplemented, i.e. the Glue shape. Without the fix it reproduces the CI failure exactly; with it, `lance_match(body, 'hello')` returns 15 rows and `lance_match_phrase(body, 'hello world')` returns 10 — the local path applies structured queries correctly.
- `LanceRuntimeQueryTableSupportTest` — catalog-only, query-capable, `dir`, and null/unknown impls.

## Not addressed here

On `dir`-backed catalogs FTS still routes through `queryTable`, which ignores `structured_query` and returns all rows — hence the `Assumptions.assumeTrue(false)` in `BaseFtsMatchPushdownTest`, `BaseFtsPhraseMatchPushdownTest`, `BaseFtsAdvancedOptionsTest`, and the `xfail` on the SEARCH TVF test. The local scan path already handles phrase and slop correctly today, so preferring it whenever the dataset is directly reachable would let those tests be re-enabled without waiting on the lance-core `dir.rs` bump. That is a larger behavioral change and left for a follow-up.

### DIFF
--- a/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/FtsCatalogOnlyNamespaceTest.java
+++ b/lance-spark-3.4_2.12/src/test/java/org/lance/spark/read/FtsCatalogOnlyNamespaceTest.java
@@ -1,0 +1,16 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.read;
+
+public class FtsCatalogOnlyNamespaceTest extends BaseFtsCatalogOnlyNamespaceTest {}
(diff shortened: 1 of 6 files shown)


ivscheianu (approved):
Thanks for the fix @wjones127! Sorry for the trouble

hamersaw (approved):
Thanks Will! This is certainly the right fix.

hamersaw (approved):

PR #727, opened 2026-07-28
fix: write wrong offset for fixed size list with nulls

j7nhai:
### Problem

When writing a `FixedSizeList` column (e.g. a 2048-dim embedding vector) that contains null rows **before** non-null rows, subsequent rows are read back with corrupted data — either all-null or shifted values. The bug is non-deterministic in the sense that it only manifests when nulls precede non-null entries.

### Root Cause

Arrow's `FixedSizeList` stores data in two layers:

```
Parent vector (FixedSizeListVector)          Child vector (e.g. Float4Vector)
┌──────────────────────────────┐            ┌─────────────────────────────────────────┐
│ Row 0: validity=1 (non-null)  │  ───────>  │ [0] 1.0  [1] 2.0  [2] 3.0  [3] 4.0    │
│ Row 1: validity=0 (null)      │  ───────>  │ [4] ___  [5] ___  [6] ___  [7] ___    │  ← placeholder
│ Row 2: validity=1 (non-null)  │  ───────>  │ [8] 5.0  [9] 6.0 [10] 7.0 [11] 8.0   │
└──────────────────────────────┘            └─────────────────────────────────────────┘
```

The child vector is a flat contiguous array. Row N's data is located at `[N * listSize, (N+1) * listSize)`. Even when a row is null, `listSize` slots **must** be reserved in the child vector to keep offsets aligned for subsequent rows.

`FixedSizeListWriter.setNull()` only marked the parent vector's validity bit as null but did **not** advance the child writer's position:

```scala
// Before fix
override def setNull(): Unit = {
  valueVector.setNull(count)  // parent marked null, child position unchanged
}
```

This caused every null row to shift all subsequent non-null data backward by `listSize` positions, producing corrupted reads.

### Offset Drift Example (listSize=4)

| Row | Data | elementWriter.count | Written at | Read offset | Result |
|-----|------|---------------------|-----------|-------------|--------|
| 0 | null | 0 (not advanced) | — | 0–3 | null ✅ |
| 1 | [1,2,3,4] | 0 → 4 | 0–3 | **4–7** | ❌ shifted |
| 2 | null | 4 (not advanced) | — | 8–11 | null ✅ |
| 3 | [5,6,7,8] | 4 → 8 | 4–7 | **12–15** | ❌ shifted |

### Fix

Advance the child writer's count by `listSize` on null rows, keeping offsets aligned. This is consistent with how `StructWriter.setNull()` handles null propagation to child writers.

```scala
// After fix
override def setNull(): Unit = {
  elementWriter.count += valueVector.getListSize()
  valueVector.setNull(count)
}
```

### Testing

Added `testFixedSizeListNullMixed` — writes 4 rows with interleaved nulls (`[non-null, null, null, non-null]`) and asserts the non-null row **after** nulls reads back correctly.

### Impact

- **Trigger**: any `FixedSizeList` column where null rows appear before non-null rows
- **Not triggered**: all-non-null columns, or nulls only at the tail
- **Severity**: each null row shifts all subsequent rows by `listSize` positions — data corruption scales with the number of preceding nulls

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
@@ -208,6 +208,9 @@ private[arrow] class FixedSizeListWriter(
     val elementWriter: LanceArrowFieldWriter) extends LanceArrowFieldWriter {
 
   override def setNull(): Unit = {
+    // Child vector must reserve listSize slots even for null rows,
+    // otherwise subsequent rows write to wrong offsets.
+    elementWriter.count += valueVector.getListSize()
     valueVector.setNull(count)
   }
 
--- a/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
+++ b/lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
@@ -624,4 +624,38 @@ public abstract class BaseSparkDataTypeRoundtripTest {
     assertEquals(v1, got.get(0));
     assertEquals(v2, got.get(1));
   }
+
+  // ---------------- fixed-size-list with null ----------------
+
+  @Test
+  public void testFixedSizeListNullMixed() {
+    org.apache.spark.sql.types.Metadata meta =
+        new org.apache.spark.sql.types.MetadataBuilder()
+            .putLong("arrow.fixed-size-list.size", 4L)
+            .build();
+    StructType schema =
+        new StructType()
+            .add("id", DataTypes.IntegerType, false)
+            .add(
+                org.apache.spark.sql.types.DataTypes.createStructField(
+                    "vec", DataTypes.createArrayType(DataTypes.FloatType, false), true, meta));
+    List<Row> data =
+        Arrays.asList(
+            RowFactory.create(0, Arrays.asList(1.0f, 2.0f, 3.0f, 4.0f)),
+            RowFactory.create(1, null),
+            RowFactory.create(2, null),
+            RowFactory.create(3, Arrays.asList(5.0f, 6.0f, 7.0f, 8.0f)));
+    List<Row> out = writeAndRead(schema, data, "fsl_null").orderBy("id").collectAsList();
+    assertEquals(4, out.size());
+    // Row 0: non-null
+    assertFalse(out.get(0).isNullAt(1));
+    assertEquals(Arrays.asList(1.0f, 2.0f, 3.0f, 4.0f), out.get(0).getList(1));
+    // Row 1: null
+    assertTrue(out.get(1).isNullAt(1));
+    // Row 2: null
+    assertTrue(out.get(2).isNullAt(1));
+    // Row 3: non-null AFTER nulls - this is the critical assertion
+    assertFalse(out.get(3).isNullAt(1), "Row after null should not be null");
+    assertEquals(Arrays.asList(5.0f, 6.0f, 7.0f, 8.0f), out.get(3).getList(1));
+  }
 }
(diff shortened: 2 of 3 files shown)


j7nhai:
@LuciferYang @ivscheianu could you help review this PR? thank you!

LuciferYang:
I’m a bit occupied and will take a look later.

j7nhai:
@jackye1995 could you help review this PR? thank you!

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -208,6 +208,9 @@ private[arrow] class FixedSizeListWriter(
     val elementWriter: LanceArrowFieldWriter) extends LanceArrowFieldWriter {
 
   override def setNull(): Unit = {
+    // Child vector must reserve listSize slots even for null rows,
+    // otherwise subsequent rows write to wrong offsets.
+    elementWriter.count += valueVector.getListSize()
```
LuciferYang, line 213: This only fixes the write path, so tables already on disk are still wrong.
I wrote one with main and read it back with your build: identical to what main itself reads.
Rewriting doesn't help either.
`read(old).write(new)` carries the bad values straight through, because the read side is already picking them up from the shifted layout.

lance-spark-base_2.12/src/test/java/org/lance/spark/BaseSparkDataTypeRoundtripTest.java
```diff
@@ -624,4 +624,38 @@ public void testUDTAsArrayElementRoundtrip() {
     assertEquals(v1, got.get(0));
     assertEquals(v2, got.get(1));
   }
+
+  // ---------------- fixed-size-list with null ----------------
+
+  @Test
+  public void testFixedSizeListNullMixed() {
```
LuciferYang, line 631: This fix covers more paths than the test does.
Besides the top-level FSL column I tried four more shapes, all broken on main and all fixed by your one line:
a fully null struct row,
a non-null struct whose FSL field is null,
`ARRAY<ARRAY<FLOAT>>` with FSL metadata on the inner type (a null outer element corrupts the elements after it),
and `MAP<INT, ARRAY<FLOAT>>` with an FSL value.

Start with the fully null struct row:
every other shape goes through the generic `LanceArrowFieldWriter.write` dispatch,
and only that one goes through the loop in `StructWriter.setNull()` that calls `setNull()` on each child.

That loop looks like dead code (24 of the 28 `setNull()` overrides in this file are empty),
and when I deleted it and kept only `count += 1`,
the nested case read back all nulls while `testFixedSizeListNullMixed` stayed green.

Array elements and Map values are worth adding if you have the budget;
note a Map column needs an explicit `file_format_version=2.2`,
since the 2.1 default doesn't support Map.

The `ADD/UPDATE COLUMNS ... FROM` backfill path (`AbstractBackfillWriter`) uses the same writer too.
I measured it broken on main and fixed on your build, so it deserves a case as well.

`AddIndexExec` also uses the writer,
but a BTREE index on an FSL column is rejected by Lance core at `createIndex` (non-nested field),
so that path is moot.

`ShardingBatchKeyEvaluator`'s writer only computes shard keys and never lands on disk,
so it matters less.

For schema construction see `BaseSparkConnectorReadTest#testFixedSizeArrayInNestedStruct`.

LuciferYang:
The line "Vectors must be non-nullable" in `docs/src/operations/ddl/create-table.md:98` doesn't match this PR.
The code has never rejected nullable vector columns (`LanceArrowUtils` passes the field's nullable straight into the FSL `FieldType`);
the data it wrote was simply wrong, and you've now made it correct.

Enforcing the doc instead isn't practical:
a check on top-level columns would reject this PR's own new test (`:641`'s `vec` is nullable)
and would break the existing `BaseFloat16VectorTest` table creation,
and the nested case can't be blocked at all,
because `LanceArrowUtils.scala:431`'s `nullable || field.nullable` force-relaxes child fields whenever the parent struct is nullable.

I checked: a child declared NOT NULL still comes out nullable.
That sentence came in with the old commit that added vector column support and was never revisited,
so either drop it or rewrite it to say nullable vectors are supported now.
`AGENTS.md:61` also asks for docs to be updated alongside user-facing changes.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
j7nhai: You're right — this PR only fixes the write path going forward. Existing tables written with the old code have a corrupted on-disk layout that can't be recovered by re-reading (the read side interprets the shifted layout as actual values).

j7nhai:
@LuciferYang I have fixed the doc and enhanced tests

hamersaw (approved):
Thanks for the fix! Once CI passes we can merge.

PR #733, opened 2026-07-29
fix: support Arrow LargeList columns

### LINKED ISSUE #732: Spark fails to read Arrow LargeList columns
## Problem

  Lance datasets containing Arrow `LargeList` columns cannot be loaded through
  the Spark catalog. Schema resolution fails with:

  ```text
  [UNSUPPORTED_ARROWTYPE] Unsupported arrow type LargeList
  ```

  This reproduces with lance-spark 0.1.3-beta.8 and 0.6.1.

  ## Root cause

  PR #155 added a `LargeListVector` accessor, but
  `LanceArrowUtils.fromArrowField` still delegates `ArrowType.LargeList` to
  Spark's standard Arrow converter. Spark does not support that conversion.

  The failure therefore occurs during `spark.table(...)` schema resolution,
  before any rows are read.

  ## Proposed fix

  - Map Arrow `LargeList` to Spark `ArrayType`.
  - Preserve nested element metadata.
  - Use the existing `LanceLargeArrayAccessor`.
  - Reject 64-bit offsets or lengths that Spark's 32-bit array indexing cannot
    represent.
  - Add tests for top-level, nested, null, and empty `LargeList` values.

  I have a tested implementation and can submit a PR.

skycutter:
Closes #732.

Spark's Arrow converter does not support `ArrowType.LargeList`, causing Lance tables containing LargeList columns to fail during catalog schema resolution.

This change:

- Maps Arrow `LargeList` fields to Spark `ArrayType`.
- Preserves nested element metadata for LargeList fields.
- Uses the existing `LanceLargeArrayAccessor` for vector materialization.
- Adds checked conversion of 64-bit offsets and lengths to Spark's 32-bit array indexing, producing a clear error when a value cannot be represented.
- Adds focused tests for top-level and nested schemas, vector reads, null and empty arrays, and unrepresentable offsets.

## Are these changes tested?

- Spark 4.0 / Scala 2.13: `LargeListSupportTest` — 3 tests passed.
- Spark 3.5 / Scala 2.12: `LargeListSupportTest` — 3 tests passed.
- Spark 4.0 runtime bundle built successfully.
- The patched bundle was validated against an existing Lance dataset with LargeList columns, including a 1,000-row materialization test.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/LanceLargeArrayAccessor.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/LanceLargeArrayAccessor.java
@@ -42,7 +42,18 @@ public class LanceLargeArrayAccessor {
   public ColumnarArray getArray(int rowId) {
     long start = accessor.getElementStartIndex(rowId);
     long end = accessor.getElementEndIndex(rowId);
-    return new ColumnarArray(arrayData, (int) start, (int) (end - start));
+    return new ColumnarArray(
+        arrayData, toSparkArrayInt(start, "offset"), toSparkArrayInt(end - start, "length"));
+  }
+
+  static int toSparkArrayInt(long value, String description) {
+    if (value < 0 || value > Integer.MAX_VALUE) {
+      throw new UnsupportedOperationException(
+          String.format(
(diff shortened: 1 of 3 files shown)


beinan (approved):

jackye1995 (approved):
looks good!

PR #734, opened 2026-07-31
fix: hide system indexes from SHOW INDEXES

hfutatzhanghb:
## Summary

- Read raw Lance index metadata with `Dataset#getIndexes` so system indexes can be identified before any plugin-backed description or statistics call.
- Filter the reserved `__lance_frag_reuse` and `__lance_mem_wal` indexes from `SHOW INDEXES`.
- Keep `getIndexStatistics` for user-created indexes so indexed fragment and row counts remain available.
- Add a regression test backed by a real bucket-partitioned Lance table, which creates the MemWAL system index.
- Document that internal fragment-reuse and MemWAL indexes are excluded from command output.

## Problem

Lance datasets can contain internal indexes such as `__lance_frag_reuse` and `__lance_mem_wal`. These are implementation details and must not appear in Spark's `SHOW INDEXES` output.

`ShowIndexesExec` previously called `Dataset#describeIndices` for the complete index list and then called `Dataset#getIndexStatistics` for every returned entry. With Lance Core 9.0.0, describing a dataset containing MemWAL can fail before any `IndexDescription` is returned:

```text
No scalar index plugin found for name 'memwal'
  at org.lance.Dataset.nativeDescribeIndices
  at org.lance.Dataset.describeIndices
  at ShowIndexesExec.run
```

This means filtering by `IndexDescription#indexType` after `describeIndices()` is too late. The command fails before Spark can inspect or filter the descriptions.

## Solution

`ShowIndexesExec` now uses `Dataset#getIndexes` to read raw manifest metadata. That API does not open or resolve the index plugin, so Spark can safely:

1. Remove entries named `__lance_frag_reuse` or `__lance_mem_wal`.
2. Group raw segments by logical index name and preserve deterministic name ordering.
3. Resolve field names from the remaining user-index metadata.
4. Call `getIndexStatistics` only for those user-created indexes.

This keeps the existing user-facing statistics behavior instead of dropping all statistics calls, while ensuring system indexes never reach `getIndexStatistics`.

## Tests
Have tested in my production envirnment.

The regression test:

1. Creates a bucket-partitioned Lance table and inserts data, producing the MemWAL system index.
2. Verifies `SHOW INDEXES` succeeds and does not expose the system index.
3. Creates a user B-tree index on the same table.
4. Verifies the B-tree index remains visible with its type and indexed fragment/row statistics.

The first CI revision demonstrated the regression consistently across the Spark 3.4, 3.5, 4.0, 4.1, and 4.2 unit-test matrix, all failing at `Dataset#describeIndices` with the same MemWAL plugin error. The updated implementation avoids that failing API path for index enumeration.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ShowIndexesExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ShowIndexesExec.scala
@@ -24,6 +24,13 @@ import org.lance.spark.utils.{FieldPathUtils, Utils}
 
 import scala.collection.JavaConverters._
 
+object ShowIndexesExec {
+  private val SystemIndexTypes = Set("FragmentReuse", "MemWal")
+
+  private def isSystemIndex(indexType: String): Boolean =
+    indexType != null && SystemIndexTypes.exists(_.equalsIgnoreCase(indexType))
+}
+
 /**
  * Physical execution of SHOW INDEXES for Lance datasets.
  *
@@ -46,7 +53,9 @@ case class ShowIndexesExec(
 
     val dataset = Utils.openDatasetBuilder(readOptions).build()
     try {
-      val indexes = dataset.describeIndices().asScala.toSeq
+      val indexes = dataset.describeIndices().asScala.toSeq.filterNot { idx =>
+        ShowIndexesExec.isSystemIndex(idx.getIndexType)
+      }
       val lanceSchema = dataset.getLanceSchema()
 
       indexes.map { idx =>
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ShowIndexesExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ShowIndexesExec.scala
@@ -25,10 +25,10 @@ import org.lance.spark.utils.{FieldPathUtils, Utils}
 import scala.collection.JavaConverters._
 
 object ShowIndexesExec {
-  private val SystemIndexTypes = Set("FragmentReuse", "MemWal")
+  private val SystemIndexNames = Set("__lance_frag_reuse", "__lance_mem_wal")
 
-  private def isSystemIndex(indexType: String): Boolean =
-    indexType != null && SystemIndexTypes.exists(_.equalsIgnoreCase(indexType))
+  private def isSystemIndex(indexName: String): Boolean =
+    indexName != null && SystemIndexNames.exists(_.equalsIgnoreCase(indexName))
 }
 
 /**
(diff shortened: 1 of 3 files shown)


hfutatzhanghb:
@majin1102  @fangbo  Hi, could you please help review this PR have free time? Thanks very much !

hfutatzhanghb:
Hi, @LuciferYang . could you please help review this PR have free time? Thanks very much !

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseShowIndexesTest.java
```diff
@@ -131,4 +131,31 @@ public void testShowIndexes() {
     long numIndexedRows = row.getLong(4);
     Assertions.assertTrue(numIndexedRows >= 1L, "num_indexed_rows should be at least 1");
   }
+
+  @Test
+  public void testShowIndexesFiltersSystemIndexes() {
```
fangbo: The index `__lance_frag_reuse` test case maybe missed ? We can create btree index using `defer_index_remap=true` to create fragment reuse index.

hfutatzhanghb: Have added!  Thanks for reviewing.

hfutatzhanghb:
Hi, @fangbo  @LuciferYang . All were ready for reviewing. Please help check it when have free time. Thanks very much !

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/ShowIndexesExec.scala
```diff
@@ -46,11 +53,16 @@ case class ShowIndexesExec(
 
     val dataset = Utils.openDatasetBuilder(readOptions).build()
     try {
-      val indexes = dataset.describeIndices().asScala.toSeq
+      val indexes = dataset.getIndexes.asScala.toSeq
+        .filterNot(idx => ShowIndexesExec.isSystemIndex(idx.name()))
+        .groupBy(_.name())
+        .toSeq
+        .sortBy(_._1)
+        .map(_._2.head)
```
fangbo, line 61: Is this `groupBy` necessary ?

hfutatzhanghb: `groupBy` is needed because getIndexes() returns per-segment physical Index entries — for a zonemap index with 3 segments, we get 3 Index objects all named test_index_zonemap. Without groupBy, SHOW INDEXES would show duplicate rows. The existing test testCreateZonemapIndexWithNumSegments validates exactly this: segments.size() == expectedSegmentCount > 1.

fangbo:
Great fix. LGTM

hfutatzhanghb:
@fangbo Thanks for reviewing and merging!

PR #739, opened 2026-08-03
fix: create index succeeds without creating an index on empty tables

### LINKED ISSUE #738: create index succeeds without creating an index on empty tables
### Description

When CREATE INDEX is executed on a table with no fragments, AddIndexExec returns success before validating the index type or processing the train option:

https://github.com/lance-format/lance-spark/blob/main/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala#L104

This causes:

- BTREE, ZONEMAP, and FTS creation to report success without committing index metadata.
- The index to be missing from SHOW INDEXES.
- Unsupported index types such as IVF-PQ to potentially report success instead of throwing an exception.

Lance 9.0 supports creating empty scalar indices with zero fragment coverage, so scalar index creation does not need to be skipped.

### Expected behavior

For BTREE, ZONEMAP, and FTS:

- Validate the index definition.
- Commit index metadata with zero fragment coverage.
- Return fragments_indexed = 0.
- Make the index visible through SHOW INDEXES.

For IVF-PQ and other unsupported index types:

- Throw UnsupportedOperationException, regardless of whether the table contains fragments.

Empty vector index creation is out of scope because Lance Spark does not currently support creating vector indices.

### Testing

Extend the empty-table integration test to verify that:

- Scalar index metadata is created and visible through SHOW INDEXES.
- Unsupported index types fail instead of silently succeeding.

majin1102:
## Summary

- register empty scalar index metadata when `CREATE INDEX` runs on an empty table
- validate index options before taking the empty-table path
- rely on Lance's atomic `createIndex` commit and avoid the previous duplicate metadata commit
- document the behavior and add unit/integration coverage for empty scalar indexes

## Root cause

`AddIndexExec` returned early when a table had no fragments, so the command reported success without registering index metadata. The deferred-index helper also manually committed metadata after `Dataset.createIndex`, even though the SDK call without fragment IDs already commits atomically.

## Validation

- `AddIndexTest`: 30 tests passed on Spark 3.5 / Scala 2.12
- `spotless:check` passed for `lance-spark-base_2.12`
- `git diff --check` passed
- Python integration test file parsed successfully

Closes #738

### DIFF
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -193,6 +193,10 @@ to scanning the data until it is populated. There are two ways to populate it:
 index performs no segmented build at creation time, `num_segments` cannot be combined with
 `train = false` — pass it on the eager build that populates the index instead.
 
+Creating a scalar index on an empty table also registers an empty index with zero fragment
+coverage. The index is immediately visible through `SHOW INDEXES`. After data is appended, populate
+it by re-running `CREATE INDEX` or calling `Dataset.optimizeIndices`.
+
 ## Output
 
 The `CREATE INDEX` command returns the following information about the operation:
--- a/integration-tests/test_lance_spark.py
+++ b/integration-tests/test_lance_spark.py
@@ -1000,7 +1000,7 @@ class TestDDLIndex:
         assert metadata["index_version"] == 2
 
     def test_create_index_empty_table(self, spark):
-        """Test CREATE INDEX on empty table."""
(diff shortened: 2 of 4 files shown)


majin1102:
@codex review

majin1102:
@hamersaw Hi Dan, could you take a look when you have time?
This is a minor bug fix for creating index.

hamersaw (approved):
Thanks for the fix!

PR #741, opened 2026-08-04
feat: support otel for spark connector

yanghua:
## Summary

  Add opt-in OpenTelemetry metrics integration for Lance Spark using the Java `LanceMetrics` bridge provided by `lance-core`.

  This updates Lance Spark to `lance-core` `10.1.0-beta.2`, wires OTel initialization into driver and executor read/write/search paths, and exposes
  configuration through `spark.lance.otel.enabled` or `LANCE_SPARK_OTEL_ENABLED`. It also includes the required OpenTelemetry dependencies in shaded bundle
  whitelists and documents the new configuration.

  ## Testing

  - `./mvnw spotless:apply`
  - `./mvnw compile -pl lance-spark-3.5_2.13 -am`
  - `./mvnw test -pl lance-spark-3.5_2.13 -Dtest=org.lance.spark.LanceRuntimeQueryTableSupportTest,org.lance.spark.read.SparkConnectorReadTest`

### DIFF
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -72,6 +72,23 @@ and namespace-specific options:
 | `spark.sql.catalog.{name}.parent`           | String | ✗        | Parent prefix for multi-level namespaces. See [Note on Namespace Levels](#note-on-namespace-levels).                             |
 | `spark.sql.catalog.{name}.parent_delimiter` | String | ✗        | Delimiter for parent prefix (default: `.`). See [Note on Namespace Levels](#note-on-namespace-levels).                           |
 
+## OpenTelemetry Metrics
+
+Lance Spark can bridge Lance's native Java metrics into the JVM OpenTelemetry
+pipeline. Enable it with:
+
+```shell
+--conf spark.lance.otel.enabled=true
+```
+
+or set `LANCE_SPARK_OTEL_ENABLED=true` in each JVM environment.
+
+The bridge uses the OpenTelemetry `GlobalOpenTelemetry` meter provider, so
+configure your exporter, resource, and reader with the standard OpenTelemetry
+Java SDK settings. Spark driver and executor JVMs are independent processes;
+set the Spark conf or environment variable for every process that should emit
+Lance metrics.
+
 ## Example Namespace Implementations
 
 ### Directory Namespace
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -81,7 +81,13 @@ pipeline. Enable it with:
 --conf spark.lance.otel.enabled=true
 ```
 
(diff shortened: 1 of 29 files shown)


yanghua:
@copilot resolve the merge conflicts in this pull request

docs/src/config.md
```diff
@@ -72,6 +72,23 @@ and namespace-specific options:
 | `spark.sql.catalog.{name}.parent`           | String | ✗        | Parent prefix for multi-level namespaces. See [Note on Namespace Levels](#note-on-namespace-levels).                             |
 | `spark.sql.catalog.{name}.parent_delimiter` | String | ✗        | Delimiter for parent prefix (default: `.`). See [Note on Namespace Levels](#note-on-namespace-levels).                           |
 
+## OpenTelemetry Metrics
+
+Lance Spark can bridge Lance's native Java metrics into the JVM OpenTelemetry
+pipeline. Enable it with:
+
+```shell
+--conf spark.lance.otel.enabled=true
```
fangbo, line 81: Configuration precedence should be explicitly documented
The current lookup order is:

1. System.getProperty("spark.lance.otel.enabled")
2. SparkEnv.get().conf().get("spark.lance.otel.enabled")
3. System.getenv("LANCE_SPARK_OTEL_ENABLED")

This means that if the JVM system property is explicitly set to false, OpenTelemetry will ultimately be disabled even if the Spark conf is set to true.

This is not necessarily wrong, but I would suggest:

1. Clearly documenting the precedence order.
2. Alternatively, consider giving Spark conf higher priority than the system property, since the documentation primarily recommends enabling this via --conf spark.lance.otel.enabled=true.
3. Adding a debug or warning log for invalid values. For example, spark.lance.otel.enabled=yes would be silently parsed as false by Boolean.parseBoolean, which makes troubleshooting less straightforward.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceRuntime.java
```diff
@@ -200,6 +211,42 @@ public static Session session(String catalogName) {
...
 
+  /**
+   * Installs Lance's JNI-backed OpenTelemetry metrics bridge when enabled.
+   *
+   * <p>The bridge is process-global and {@link LanceMetrics#instrument()} is idempotent, so Spark
+   * driver and executor code paths can call this before Lance work starts in each JVM.
+   *
+   * @return true if OpenTelemetry was enabled and the bridge was installed, false otherwise
+   */
+  public static boolean enableOpenTelemetry() {
```
fangbo, line 236: Each partition's reader or writer will invoke this method. Is it necessary to cache the result ?

fangbo (approved):

PR #748, opened 2026-08-06
feat: build btree fragments with scalar segments

ddupg:
## Summary
- Route BTREE fragment builds through the shared scalar segment index workflow.
- Keep BTREE range mode on the existing Spark preprocessing path.
- Allow `num_segments` for BTREE fragment builds and apply Lance's single-column index constraint consistently.

## CI concurrency note

This PR also fixes a pre-existing CI instability exposed by the migration. In local tests, four concurrent BTREE sorts (about 40 MB each) can exhaust Lance's shared 150 MB DataFusion pool. Tests now use `local[3]` while keeping `spark.default.parallelism=10`, limiting concurrent execution without changing the default 10-segment planning. The same failure reproduces before the migration, so it is not a segment-flow regression.

## Test plan
- `./mvnw spotless:check -pl lance-spark-3.5_2.13 -am`
- `./mvnw -pl lance-spark-3.4_2.12 -am -Dsurefire.failIfNoSpecifiedTests=false -Dtest=org.lance.spark.update.AddIndexTest#testIndexesRejectMultipleColumns test`
- `JAVA_HOME=/Library/Java/JavaVirtualMachines/temurin-17.jdk/Contents/Home ./mvnw -pl lance-spark-4.1_2.13 -am -Dsurefire.failIfNoSpecifiedTests=false -Dtest=org.lance.spark.update.AddIndexTest,org.lance.spark.update.CreateIndexStandardSyntaxTest test` (79 tests)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -161,53 +164,35 @@ case class AddIndexExec(
     val (nsImpl, nsProps, tableId, initialStorageOpts) =
       extractNamespaceInfo(lanceDataset, readOptions)
 
-    // Scalar segment indexes use the logical segment commit path.
-    if (scalarSegmentIndexType.isDefined) {
-      val scalarSegmentJob = new ScalarSegmentIndexJob(
+    // Range-mode BTree uses preprocessed data from Spark and keeps its dedicated path.
+    if (btreeBuildMode.contains("range")) {
+      val segments = new RangeBasedBTreeIndexJob(
```
majin1102, line 167: Since BTREE sorting cost depends on the amount of data rather than fragment count, assigning fragments to the selected numSegments purely by count may create substantial task skew. 

Could we use fragment row counts for batching? After selecting numSegments, sort fragments by row count and greedily assign each one to the least-loaded segment to reduce BTREE sort skew.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -703,6 +583,7 @@ object IndexUtils extends Logging {
     "inverted" -> IndexType.INVERTED)
 
   private val scalarSegmentIndexTypes: Set[IndexType] = Set(
+    IndexType.BTREE,
```
majin1102, line 584: Adding BTREE here also changes the resolution at lines 96–99 from resolveLeafField to resolveField, allowing non-leaf container fields to pass driver-side validation. Could we decouple segment-job selection from container-field support and preserve resolveLeafField for BTREE?

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
```diff
@@ -773,15 +788,60 @@ public void testZonemapDeferredRejectsNumSegments() {
...
-                    fullTable)));
+    Dataset<Row> result =
+        spark.sql(
+            String.format(
+                "alter table %s create index idx_btree_seg using btree (id) with (num_segments=3)",
+                fullTable));
+
+    Assertions.assertEquals("idx_btree_seg", result.collectAsList().get(0).getString(1));
+
+    org.lance.Dataset lanceDataset = org.lance.Dataset.open().uri(tableDir).build();
```
majin1102: I guess we should use Utils.openDatasetBuilder(LanceSparkReadOptions.from(tableDir)).build() here. AGENTS.md requires datasets to be opened through this helper so storage-option merging and runtime-session management are preserved.

majin1102 (commented):
left some comments, PTAL

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
ddupg: Good catch, fixed.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
ddupg: Thanks, this is a good point. The main purpose of this PR is to migrate BTREE fragment builds onto the shared scalar segment flow and remove the duplicated BTREE-specific execution path.

Row-count-aware fragment grouping is a broader improvement to the shared scalar segment planner, rather than something specific to BTREE. To keep this PR focused, I opened #757 to track it separately. Lance-Ray’s greedy balancing strategy can serve as a reference there.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
ddupg: Updated. I decoupled field-shape validation from segment-job selection and preserved leaf-only validation for BTREE, BITMAP, NGRAM, and BLOOM_FILTER.

My original thought was to keep Lance Spark permissive and rely on Lance Core for type validation to reduce rule drift—an approach informed by my experience with Lance-Ray. However, Lance Spark only distinguishes leaf fields from container fields here and leaves detailed type compatibility to Lance Core, so this structural validation should remain stable. I have updated it as suggested.

PR #749, opened 2026-08-06
fix: preserve LargeBinary across a Spark schema round trip

beinan:
## Problem

Spark has a single `BinaryType` that covers both Arrow `Binary` (32-bit offsets) and Arrow `LargeBinary` (64-bit offsets). #413 fixed the **write** path so `use_large_var_types` produces `LargeBinary`, but the **read** path still drops the distinction: a Lance `LargeBinary` column comes back as a plain `BinaryType` with no marker, so writing that DataFrame back narrows the column to `Binary` and the append is rejected by Lance's schema validation.

This bites any read → transform → write round trip, which is the common shape for backfills:

```python
df = spark.read...load()                       # LARGE_BINARY -> BINARY
df = df.withColumn("md5", md5("image_bytes"))  # image_bytes untouched
df.writeTo(table).append()                     # BINARY != LARGE_BINARY -> fail
```

It also affects `UPDATE` and `ALTER TABLE ... ADD COLUMNS FROM`, which round-trip the table schema through Spark internally.

The asymmetry is visible in `LanceArrowUtils`:

- write (`toArrowType`): `case BinaryType if largeVarTypes => ArrowType.LargeBinary.INSTANCE` ✅
- read (`convertArrowFieldType`): `case _: ArrowType.LargeBinary => BinaryType`, and `augmentTypeMarkers` had **no `LargeBinary` case** — so the 64-bit-offset information was discarded.

`LargeUtf8` already gets a marker (`arrow:large-var-char`), and `FixedSizeBinary` already preserves its byte width. `LargeBinary` was the one that got missed.

## Fix

Record the 64-bit-offset marker in Spark field metadata on read and honor it on writeback, mirroring what `LargeVarCharUtils` already does for `LargeUtf8`.

- New `LargeVarBinaryUtils` holds the key, alongside the existing `LargeVarCharUtils` / `FixedSizeBinaryUtils`.
- The writeback check is gated on `dt == BinaryType`, so a stray copied metadata key cannot widen an unrelated column.
- The read-side marker skips blob columns (`if !isBlobField(field)`) — those are already `LargeBinary`-backed via their own marker, which `toArrowField` honors on its own.

## Second commit: Spark 3.4 read path

Adding the end-to-end test surfaced a second, pre-existing instance of the same omission one layer down. Spark 3.4's `ArrowUtils.fromArrowType` rejects `LargeBinary` with `UNSUPPORTED_ARROWTYPE`, and `LanceArrowColumnVector` was delegating `LargeVarBinaryVector` to Spark's `ArrowColumnVector`:

```
UNSUPPORTED_ARROWTYPE: Unsupported arrow type LargeBinary
  at org.apache.spark.sql.util.ArrowUtils$.fromArrowType(ArrowUtils.scala:82)
  at org.apache.spark.sql.vectorized.ArrowColumnVector.<init>(ArrowColumnVector.java:140)
  at org.lance.spark.vectorized.LanceArrowColumnVector.<init>(LanceArrowColumnVector.java:152)
```

`LanceArrowColumnVector` already routes `LargeVarCharVector` to a dedicated accessor for exactly this reason on `LargeUtf8` — `LargeBinary` was missing the equivalent. Added `LargeVarBinaryAccessor`, mirroring `LargeVarCharAccessor`.

This gap broke **any** read of a LargeBinary Lance table on Spark 3.4, independent of the metadata change here; this branch's test is simply the first coverage to exercise that path on 3.4.

## Tests

Three new tests, each verified to **fail without the fix and pass with it**:

| Test | Without fix |
|---|---|
| `LargeBinary roundtrip preserves 64-bit offsets` | FAILED |
| `LargeBinary inside Array roundtrips to LargeBinary` | FAILED |
| `appendAfterTransformPreservesLargeBinary` (end-to-end) | `RuntimeException: Failed to write data to lance` |

The unit test also asserts that a **sibling plain `Binary` column is not widened**, guarding the main regression risk. The end-to-end test reproduces the reported scenario (read → `withColumn` → `append`) and its failure without the fix matches the reported error exactly.

Verified on Temurin JDK 17 (matching CI). Removing just the accessor wiring reproduces the CI
error exactly, confirming that line is what fixes it. Full suites green: `lance-spark-base_2.12`
(573 tests), `lance-spark-3.4_2.12` (1127 tests), `lance-spark-3.5_2.12` (1266 tests, including
all blob tests), and checkstyle.

Reported by Ryan Stuart, who also root-caused the read/write asymmetry and supplied the initial patch that this PR builds on — thank you!

Co-authored-by: Ryan Stuart <ryanstuart@microsoft.com>

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LargeVarBinaryUtils.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/LargeVarBinaryUtils.java
@@ -1,0 +1,88 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.lance.spark.utils;
+
+import org.apache.arrow.vector.types.pojo.Field;
+import org.apache.spark.sql.types.BinaryType;
+import org.apache.spark.sql.types.StructField;
+
+import java.util.Map;
+
+/**
+ * Utility class for LargeBinary Arrow type metadata handling.
+ *
+ * <p>Spark has a single {@code BinaryType} that covers both Arrow {@code Binary} (32-bit offsets)
+ * and Arrow {@code LargeBinary} (64-bit offsets). When reading a Lance table containing a
+ * LargeBinary column, the distinction is preserved in Spark field metadata so a subsequent write
+ * reproduces LargeBinary instead of silently narrowing the column to Binary.
+ *
+ * <p>This mirrors {@link LargeVarCharUtils}, which does the same for LargeUtf8.
+ */
+public class LargeVarBinaryUtils {
+
+  public static final String ARROW_LARGE_VAR_BINARY_KEY = "arrow:large-var-binary";
+  public static final String ARROW_LARGE_VAR_BINARY_VALUE = "true";
+
+  /**
+   * Check if a Spark field is a large binary field based on its metadata.
+   *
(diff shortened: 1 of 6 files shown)


LuciferYang (approved):

LuciferYang:
Merged into main. Thank @beinan

PR #751, opened 2026-08-07
feat: preserve lance schema for dataframe overwrite

j7nhai:
### Background

In the regular Spark-to-Arrow write path, schema is reconstructed using Spark-expressible types. This can lose original Lance schema semantics, such as unsigned types, `FixedSizeList`, `FixedSizeBinary`, nullability, and field metadata.

### Semantic Guarantees in This Fix

- **Schema preservation applies only to explicit overwrite**: only Spark truncate-overwrite (`df.write.mode("overwrite")`) prefers the original Lance Arrow schema.
- **No schema-preservation constraint for replace/create-or-replace**: `REPLACE TABLE` and `CREATE OR REPLACE TABLE` continue to rebuild the target schema under staged replace semantics.
- **Safety boundary for overwrite**: on explicit overwrite, if original schema is missing or structurally incompatible, fail fast to avoid silent schema drift.
- **Metadata policy**: explicit overwrite does not merge Spark metadata; Lance metadata is kept unchanged.

### Implementation Highlights

- In `SparkWriteBuilder`, decouple "native write mode is OVERWRITE" from "explicit overwrite semantics":
  - `truncate()` is used only for explicit overwrite (schema-preserving path).
  - staged replace uses `forceOverwriteWriteMode()`, which affects native write mode only and does not trigger schema preservation.
- In `LanceBatchWrite`, schema preservation and compatibility checks are controlled only by the explicit-overwrite flag, not by `write_mode=OVERWRITE` as a fallback trigger.

### Added Tests

- `OverwriteSchemaPreservationTest` now covers explicit-overwrite scenarios including unsigned types, nullability, `FixedSizeList`, `FixedSizeBinary`, metadata retention, repeated overwrite, and large-batch writes.

### Compatibility and Impact

- Append semantics remain unchanged.
- `REPLACE TABLE` and `CREATE OR REPLACE TABLE` return to expected staged-replace behavior and are no longer affected by explicit-overwrite schema-preservation guards.


j7nhai:
@LuciferYang @hamersaw  could you help review this PR? thank you!

j7nhai:
Gentle ping @beinan @LuciferYang — this is part of a series of PRs I'm preparing. Would appreciate a review when you have a moment. Thanks! 🙏

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
```diff
@@ -165,7 +185,22 @@ public void commit(WriterCommitMessage[] messages) {
 
     Schema arrowSchema =
         LanceArrowUtils.toArrowSchema(schema, "UTC", true, writeOptions.isUseLargeVarTypes());
-    boolean isOverwrite = overwrite || writeOptions.isOverwrite();
+    boolean isOverwrite = overwrite;
```
LuciferYang, line 188: **HIGH** · correctness

`LanceBatchWrite.java:188` narrows `overwrite || writeOptions.isOverwrite()` to just `overwrite`, taking the documented `write_mode=overwrite` option (`insert-into.md:287`) out of play on the commit path. Measured: 2 rows in the table, then that option with `mode("append")` and 1 row gives 3 rows on your build and 1 on main. The executor still builds fragments under native OVERWRITE, so fragment and manifest schemas diverge; with `use_large_var_types=true` you can commit an unreadable table. After this change `isOverwrite()` has no callers left in the repo, so the removal looks unintentional. Either reject the option up front or keep routing it through the Overwrite operation.

lance-spark-base_2.12/src/main/scala/org/lance/spark/arrow/LanceArrowWriter.scala
```diff
@@ -292,6 +296,42 @@ private[arrow] class UnsignedLongWriter(val valueVector: UInt8Vector)
   }
 }
 
+/** Writes Spark ShortType data into a UInt1Vector (uint8). */
+private[arrow] class ShortToUnsignedByteWriter(val valueVector: UInt1Vector)
+  extends LanceArrowFieldWriter {
+  override def setNull(): Unit = {}
+  override def setValue(input: SpecializedGetters, ordinal: Int): Unit = {
+    valueVector.setSafe(count, input.getShort(ordinal).toByte)
```
LuciferYang, line 304: **HIGH** · correctness

All four new unsigned writers are unchecked narrowing casts (`.toByte` at `:304`, `.toChar` at `:313`, `.toInt` at `:331`), so out-of-range values wrap silently. Measured via `mode("overwrite")`: uint8 given `255, 300, -1` reads back `255, 44, 255`; uint32 given `42, 5000000000, -7` reads back `42, 705032704, 4294967289`. main keeps the values by rebuilding the column as signed, so this trades schema drift for value corruption. Since `isTypeCompatible` whitelists those three widenings, the writers need range checks that throw. Do not extend that to uint64, since the test at `:185` relies on `-1L` meaning uint64 max.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
```diff
@@ -165,7 +185,22 @@ public void commit(WriterCommitMessage[] messages) {
...
+    boolean isExplicitOverwrite = overwrite && stagedCommit == null;
+
+    // In explicit non-staged overwrite mode, original schema must exist and remain compatible.
+    if (isExplicitOverwrite) {
+      if (originalArrowSchema == null) {
+        throw new IllegalStateException(
+            "Overwrite requires existing Lance schema, but none was found.");
+      }
+      if (!isTypeCompatible(originalArrowSchema, arrowSchema)) {
```
LuciferYang, line 197: **HIGH** · compatibility

`isFieldCompatible` covers only four families (Int, List/FixedSizeList, Utf8/LargeUtf8, Binary); everything else falls through to `ot.equals(st)`, and since `:187` (pre-existing) always derives the schema with `"UTC"`, non-UTC timestamp columns are guaranteed incompatible. I measured six types failing here that succeed on main. To be fair, those successes come from rewriting the schema, which is the drift you are removing. The cost: `Float16`, non-UTC `Timestamp` and `LargeList` tables cannot be appended to on either side, so overwrite was their only write path and closing it makes them read-only. The check also sits in `commit()`, after data is written; the constructor already has the dataset open.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
```diff
@@ -165,7 +185,22 @@ public void commit(WriterCommitMessage[] messages) {
...
+        throw new IllegalArgumentException(
+            "Overwrite schema is incompatible with existing Lance schema. "
+                + "Overwrite must not change schema type families.");
+      }
+      arrowSchema = originalArrowSchema;
```
LuciferYang, line 202: **MEDIUM** · compatibility

`LanceBatchWrite.java:202` swaps in `originalArrowSchema` outright, bypassing the rule at `LanceArrowUtils.scala:424-431` that relaxes struct children when the parent is nullable. Measured with `p Struct<x int32 NOT NULL>` and one `p = NULL` row: your build fails on the executor with `contained null values even though the field is marked non-null`, main succeeds. Preserving NOT NULL is the better semantics, so the real defect is that the error never mentions overwrite schema preservation. I would avoid both obvious fixes: relaxing erases an externally declared NOT NULL, and rejecting at plan time kills the still-working no-null-parent case. Wrap the executor error with an attributable message and document the limit.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceDataWriter.java
```diff
@@ -313,6 +322,53 @@ private BufferAndTask buildBufferAndTask(BlobReferenceResolver resolver) {
...
+        boolean useQueuedBuffer,
+        int batchSize,
+        boolean useLargeVarTypes,
+        long maxBatchBytes,
+        BlobReferenceResolver resolver,
+        Schema originalSchema) {
+      if (useQueuedBuffer) {
+        int queueDepth = writeOptions.getQueueDepth();
+        if (originalSchema != null) {
+          return new QueuedArrowBatchWriteBuffer(
```
LuciferYang, line 357: **MEDIUM** · design

Neither `originalSchema != null` branch in `createWriteBuffer` (`LanceDataWriter.java:357` and `:365`) passes `useLargeVarTypes` down, even though the signature accepts it. Measured on a plain Utf8 table: `.option("use_large_var_types","true").mode("overwrite")` succeeds but leaves manifest and data as `Utf8`, where main produces `LargeUtf8`. This has a much wider trigger surface than the other findings, needing no externally created table and no native types. The option exists solely to get past the 2GB offset limit, so losing it silently means users hit overflow again with nothing to point at.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceDataWriter.java
```diff
@@ -313,6 +322,53 @@ private BufferAndTask buildBufferAndTask(BlobReferenceResolver resolver) {
...
+              originalSchema, schema, batchSize, queueDepth, maxBatchBytes, resolver);
+        }
+        return new QueuedArrowBatchWriteBuffer(
+            schema, batchSize, queueDepth, useLargeVarTypes, maxBatchBytes, resolver);
+      }
+
+      if (originalSchema != null) {
```
LuciferYang, line 364: **MEDIUM** · design

`LanceDataWriter.java:364` only uses the preserved schema for overwrite. The append branch and both version-specific `SparkPositionDeltaWrite` classes (`:311` on 3.5, `:279` on 3.4) still build from the Spark-derived schema. So after a uint32 table is overwritten and keeps `Int(32,false)`, `mode("append")`, `UPDATE` and `DELETE` all fail with `Append with different schema`. All three work on main because its overwrite converts the column to signed. Appending into a natively created uint32 table was already broken, so you did not introduce that, but removing the one workaround makes "overwrite-only table" a permanent state.

docs/src/operations/dml/insert-overwrite.md
```diff
@@ -2,6 +2,17 @@
 
 Replace all existing data in a table with new data. This operation removes all existing rows and inserts the new data atomically.
 
+## Schema Preservation Semantics
+
+For explicit DataFrame overwrite paths (for example, `mode("overwrite")` and `writeTo(...).overwritePartitions()`), Lance-Spark prefers the existing table's original Arrow schema instead of rebuilding from Spark-expressible types.
```
LuciferYang, line 7: **MEDIUM** · docs

Both examples in `insert-overwrite.md:7` are off. `writeTo(...).overwritePartitions()` cannot run against Lance: it needs `SupportsDynamicOverwrite` plus `OVERWRITE_DYNAMIC`, and `LanceDataset.java:59-61` only declares `BATCH_READ`, `BATCH_WRITE` and `TRUNCATE`. It fails at analysis with `does not support dynamic overwrite`, and `BaseBlobV2CopyFailureTest.java:35` already asserts that. `mode("overwrite")` splits in two: `save()` preserves, but `saveAsTable()` compiles to `ReplaceTableAsSelect` and takes the staged-replace path, which does not (measured: `id INT NOT NULL` came back nullable), and `saveAsTable` is what this section demonstrates.

lance-spark-base_2.12/src/test/java/org/lance/spark/write/OverwriteSchemaPreservationTest.java
```diff
@@ -0,0 +1,1087 @@
...
+    InternalRow row = new GenericInternalRow(new Object[] {1, payload});
+
+    assertThrows(
+        RuntimeException.class,
+        () -> overwriteAndGetSchemaWithSparkSchema(uri, incompatibleSparkSchema, writeOptions, row),
+        "Overwrite must reject incompatible nested child types");
+  }
+
+  @Test
+  public void testOverwriteLocksUnsignedOverflowWrapBehavior(TestInfo testInfo) throws IOException {
```
LuciferYang, line 753: **MEDIUM** · test coverage

I checked this file out onto main and ran it: 27 tests, 20 failures, 7 passes. The 7 that pass (nullability, `FixedSizeList`, `FixedSizeBinary`, Decimal, nested struct, append) already round-trip on main, so they are snapshots rather than regression guards, yet the PR description lists the first three as covered. The real gap is value-level coverage for the four unsigned writers: changing `.toChar` to `.toByte`, or making `UnsignedIntWriter` throw outright, leaves all 27 green. The two type-reject tests are actually caught on the executor, and deleting the whole driver-side check keeps the suite green too.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/QueuedArrowBatchWriteBuffer.java
```diff
@@ -155,25 +150,36 @@ public QueuedArrowBatchWriteBuffer(
...
+        schema,
         sparkSchema,
         batchSize,
         queueDepth,
         maxBatchBytes,
         resolver);
   }
 
-  public QueuedArrowBatchWriteBuffer(
+  QueuedArrowBatchWriteBuffer(
```
LuciferYang, line 170: **MEDIUM** · compatibility

This PR also changes constructor visibility in the two write buffers. Per `javap`: `QueuedArrowBatchWriteBuffer`'s `public (StructType,int,int)` is deleted, three more (`:170`/`:182`/`:198`) drop to package-private, `SemaphoreArrowBatchWriteBuffer`'s `:133` and `:155` drop as well, and `WriterFactory`'s 8-arg constructor is replaced rather than overloaded. Everything in the repo stays green since the callers share the package, but `lance-spark-base_2.12` is published: downstream callers passing their own `BufferAllocator` will fail to compile or hit `NoSuchMethodError`, and there is no japicmp or revapi gate. Keep the old signatures as `@Deprecated` delegates, or mark this breaking in the description.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java
```diff
@@ -232,6 +267,98 @@ public void abort(WriterCommitMessage[] messages) {
...
   }
 
+  // ==================== Schema compatibility helpers ====================
+
+  /**
+   * Checks whether the original schema is structurally compatible with the Spark-derived schema.
+   * Compatible means same number of fields and each field pair is in the same "type family" (e.g.
+   * int32 signed and int32 unsigned, List and FixedSizeList, Utf8 and LargeUtf8).
+   */
+  static boolean isTypeCompatible(Schema original, Schema spark) {
```
LuciferYang, line 277: **LOW** · maintainability

The 92 lines of `isTypeCompatible` push `LanceBatchWrite` from 251 to 378 lines, while the same package already has `LanceWriteSchemaValidator` comparing table schema against input schema, with field-path error messages and 13 existing tests. They are not drop-in interchangeable (one compares Arrow `Schema`, the other Spark `StructType`, and the latter is only called from the blob v2 branch), so a new `LanceArrowSchemaCompat` alongside it with table-driven unit tests is the better shape: right now those 92 lines have no direct test at all. The field-count check is also implemented twice, and the driver-side copy is effectively dead code.

LuciferYang (commented):
Ran every scenario below on both this branch and the parent commit `71237c2c`, in side-by-side worktrees.

The direction is right, and splitting `truncate()` from staged replace is clean: `REPLACE TABLE AS SELECT` and `saveAsTable` still rebuild the schema, and the staged-commit and write test suites pass. My concern is the wiring, not the goal.

Three blockers:

1. `LanceBatchWrite.java:188` drops the `writeOptions.isOverwrite()` fallback. The documented `write_mode=overwrite` option now silently appends, and with `use_large_var_types=true` it commits a table that cannot be read back.
2. The four new unsigned writers narrow without range checks. Overwriting a uint column with out-of-range values corrupts them silently; the parent commit kept the values.
3. `isTypeCompatible` covers four type families, so overwrite hard-fails for six Lance-native types. Three of those shapes reject append on both builds, so overwrite was their only write path and they end up read-only. The check also runs inside `commit()`, after the data files are written.

Six more inline. The widest one is `use_large_var_types` being dropped under overwrite: a plain Utf8 table plus one documented option, and the fix is passing an argument that is already in the signature.

Three things I ruled out, so you can skip them: no OCC race (`getSchema()` and `version()` read from one snapshot, commit still goes through `readVersion`), blob v2 unaffected (its tests pass, Arrow schema JSON round-trips losslessly), and positional field matching is not exploitable (Spark rejects renamed columns at analysis).

Pre-existing on both builds, unrelated to this PR: Map overwrite fails, `Timestamp(us, "+00:00")` crashes the JVM, Float16 needs Arrow 18+.

j7nhai:
resolved all @LuciferYang @hamersaw

PR #752, opened 2026-08-07
feat: support schema evolution DDLs (ADD/DROP/RENAME/ALTER COLUMN)

### LINKED ISSUE #62: schema evolution DDLs
ALTER TABLE ADD COLUMN
ALTER TABLE RENAME COLUMN
ALTER TABLE ALTER COLUMN
ALTER TABLE DROP COLUMN

puchengy:
Closes #62.

Adds support for the schema-evolution DDLs requested in #62 by extending `BaseLanceNamespaceSparkCatalog.alterTable` to handle Spark `TableChange.ColumnChange` requests, delegating to the underlying Lance dataset schema-evolution API via a new `LanceSchemaEvolution` helper.

## Supported operations

| DDL | Lance API |
|-----|-----------|
| `ALTER TABLE ... ADD COLUMN[S]` | `Dataset.addColumns` |
| `ALTER TABLE ... DROP COLUMN [IF EXISTS]` | `Dataset.dropColumns` |
| `ALTER TABLE ... RENAME COLUMN` | `Dataset.alterColumns` (rename) |
| `ALTER TABLE ... ALTER COLUMN ... DROP/SET NOT NULL` | `Dataset.alterColumns` (nullability) |

## Atomic, all-or-nothing semantics

To honor Spark's `alterTable` contract, an accepted request is validated in full against the current schema and then committed through **exactly one** Lance core operation (`addColumns` / `dropColumns` / `alterColumns` — each commits its whole batch atomically):

- Same-kind changes are batched into that single call (e.g. `ADD COLUMNS (a, b)`; rename + nullability edits to the same column are merged into one `ColumnAlteration`).
- Requests that would require more than one core mutation are rejected **before the first write**: mixing column additions, drops, and alterations in one statement, or combining any column change with `TBLPROPERTIES` changes (which commit separately). This keeps a rejected `ALTER TABLE` from leaving the table partially mutated. Heterogeneous batching in one commit is not yet supported by the core.

## Deliberately unsupported (rejected before any mutation)

- **`ALTER COLUMN ... TYPE`** — throws `UnsupportedOperationException`. The `lance-core:10.0.0-rc.3` JNI `create_column_alteration` reads the cast target from the Java `ArrowType.toString()` (e.g. `"Int(64, true)"`) and parses it with Rust `DataType::from_str`, swallowing the parse failure with `.ok()` — so the cast target is dropped and the type change becomes a silent no-op (the commit lands but the stored type is unchanged; verified for both integer- and float-widening). This is fixed upstream in lance-format/lance#8417 (carries the cast type across FFI via the Arrow C Data Interface), but that fix is not yet in a published `org.lance:lance-core` artifact — the newest published version (`11.0.0-beta.3`) predates it, and Lance does not publish nightly/snapshot Java builds. Once a lance release containing #8417 ships, this can be enabled by bumping `<lance.version>` and replacing the rejection with the `castTo` path.
- **ADD COLUMN with a `DEFAULT` value** — Lance's all-null add cannot backfill a default, so it is rejected rather than silently filling `NULL`.
- **ADD COLUMN on a legacy-format (`file_format_version='LEGACY'`) table** — the core all-null add path rejects legacy datasets; rejected up front with a clear message instead of a raw core error.
- Positional (`FIRST`/`AFTER`) adds, nested-column adds, and column-comment updates.

## Tests

- Java unit tests in `SparkLanceNamespaceTestBase` (run against both directory and REST namespaces): ADD/DROP/RENAME COLUMN and ALTER COLUMN DROP NOT NULL happy paths; batched multi-column ADD; and rejection cases for mixed-kind requests, column-change-plus-`TBLPROPERTIES`, `DROP COLUMN IF EXISTS` on a missing column, `DEFAULT` values, legacy format, and unsupported type changes.
- PySpark integration tests in `TestDDLAlterTableColumns`.
- `TestSparkDirectoryNamespace` full suite passes locally (66/66); `make lint` clean.

Docs updated in `docs/src/operations/ddl/alter-table.md`.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
@@ -15,7 +15,6 @@ package org.lance.spark;
 
 import org.lance.Dataset;
 import org.lance.schema.ColumnAlteration;
-import org.lance.spark.utils.FieldPathUtils;
 
 import org.apache.arrow.vector.types.pojo.Field;
 import org.apache.arrow.vector.types.pojo.Schema;
@@ -31,24 +30,30 @@ import org.apache.spark.sql.types.StructType;
 import org.apache.spark.sql.util.LanceArrowUtils;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.LinkedHashMap;
+import java.util.HashSet;
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 /**
  * Translates a Spark {@link ColumnChange} schema-evolution request (produced by {@code ALTER TABLE
  * ADD/DROP/RENAME/ALTER COLUMN}) into a single atomic {@link Dataset} mutation.
  *
- * <p>To preserve Spark's all-or-nothing {@code alterTable} contract, an accepted request is
- * committed through exactly one core operation: {@code addColumns}, {@code dropColumns}, or {@code
- * alterColumns} (each of which commits its whole batch atomically). The request is fully validated
- * against the current schema, and requests that would require more than one core mutation are
- * rejected, <em>before</em> anything is written — so a rejected request never leaves the table
- * partially mutated. Heterogeneous batching (mixing additions, drops, and alterations in one
- * commit) is not supported by the core yet, so such requests must be issued as separate statements.
+ * <p>To preserve Spark's ordered, all-or-nothing {@code alterTable} contract, an accepted request
+ * is committed through exactly one core operation ({@code addColumns}, {@code dropColumns}, or
+ * {@code alterColumns} — each commits its whole batch atomically <em>against the current
+ * schema</em> with no ordering between the batched entries). Because the batch cannot express
+ * changes that depend on an earlier change in the same request, the request is validated against
+ * the current schema and the following are rejected <em>before</em> anything is written, so a
+ * rejected request never leaves the table partially mutated:
+ *
+ * <ul>
+ *   <li>requests that would require more than one core mutation (mixing additions, drops, and
+ *       alterations, since heterogeneous batching is not supported by the core yet);
+ *   <li>requests where two changes target the same column (their order would be lost when collapsed
+ *       into one batch);
+ *   <li>nested (multi-part) column paths, since validation is top-level only.
+ * </ul>
  */
 final class LanceSchemaEvolution {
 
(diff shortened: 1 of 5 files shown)


lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
```diff
@@ -0,0 +1,304 @@
...
+ *
+ * <ul>
+ *   <li>requests that would require more than one core mutation (mixing additions, drops, and
+ *       alterations, since heterogeneous batching is not supported by the core yet);
+ *   <li>requests where two changes target the same column (their order would be lost when collapsed
+ *       into one batch);
+ *   <li>nested (multi-part) column paths, since validation is top-level only.
+ * </ul>
+ */
+final class LanceSchemaEvolution {
```
fangbo, line 60: Great implementation. I have a problem to discuss. LanceSchemaEvolution operates directly on Dataset . But LanceNamespace also has some interfaces for schema evolution, such as [alterTableAddColumns](https://github.com/lance-format/lance-namespace/blob/main/java/lance-namespace-core/src/main/java/org/lance/namespace/LanceNamespace.java#L646) . Do you think it is necessary to call the relevant interfaces of LanceNamespace?

puchengy: Thanks for the comment, let me take a look. Thanks!

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
```diff
@@ -0,0 +1,106 @@
...
+ * Translates Spark {@link ColumnChange} schema evolution requests (produced by {@code ALTER TABLE
+ * ADD/DROP/RENAME/ALTER COLUMN}) into the corresponding {@link Dataset} operations. Changes are
+ * applied to the open dataset in the order Spark supplies them.
+ */
+final class LanceSchemaEvolution {
+
+  private LanceSchemaEvolution() {}
+
+  static void apply(Dataset dataset, List<ColumnChange> changes) {
+    for (ColumnChange change : changes) {
```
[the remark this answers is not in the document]
puchengy: This is stale comment and has been addressed already.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
```diff
@@ -0,0 +1,291 @@
...
+    List<String> toDrop = new ArrayList<>(changes.size());
+    for (ColumnChange change : changes) {
+      DeleteColumn delete = (DeleteColumn) change;
+      String name = delete.fieldNames()[0];
+      // Validation already guaranteed a non-ifExists drop targets an existing column; skip a
+      // missing IF EXISTS target so the core is never asked to drop a nonexistent column.
+      if (delete.ifExists() && !currentFields.contains(name)) {
+        continue;
+      }
+      toDrop.add(canonicalPath(name));
```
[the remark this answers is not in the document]
puchengy: This is a stale comment.

lance-spark-base_2.12/src/main/java/org/lance/spark/LanceSchemaEvolution.java
puchengy: Thanks for the suggestion, I think it is the right call, I will implement that in lance-namespace. Thanks for calling this out!

puchengy: Hey @jackye1995, can you chime in here? 

I looked into @fangbo's suggestion and it turns out reusing lance-namespace is a bigger change than I expected. Here's the gist of what we'd need first:

  1. Wire up the Java LanceNamespace alter-column methods to the Rust impls in the lance repo — the Rust side already does the work, but there's no bridge from Java to it yet.
  2. Extend the AddColumnsEntry spec in lance-namespace so a plain ADD COLUMN age INT can be expressed — right now it only supports computed/SQL-expression columns.
  3. Wait for lance and lance-namespace to cut releases, since this repo can only pick up the changes once they're published.

So it spans three repos and needs to land in order. There are two ways:
* Option 1: merge this PR as is and have a follow up to make above changes.
* Option 2: do the above changes to get things right in the first place.

Could you weigh in by sharing your preference? Since it touches multiple repos, I'd feel better having a maintainer back the approach before I start if we go by option 2. Thanks!

PR #764, opened 2026-08-13
feat: support blob v2 columns in add columns

hfutatzhanghb:
## Summary

- apply the target table's persisted column properties and actual Lance file format version to `ADD COLUMNS ... FROM` backfill schemas
- support adding and backfilling a blob v2 `BINARY` column on file format 2.2+ tables
- validate configured blob columns are `BINARY` before append
- document the supported workflow and add integration coverage for success and error paths

## Usage

Persist the future column's blob encoding before running the backfill:

```sql
CREATE TABLE users (
    id INT,
    name STRING
) USING lance
TBLPROPERTIES ('file_format_version' = '2.2');

ALTER TABLE users
SET TBLPROPERTIES ('content.lance.encoding' = 'blob');

CREATE TEMPORARY VIEW content_backfill AS
SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
FROM users;

ALTER TABLE users ADD COLUMNS content FROM content_backfill;
```

After the operation, reads expose `content` as the existing blob v2 descriptor struct, including fields such as `size` and `kind`.

## Why

The add-columns backfill path previously used the source view's raw Spark schema. As a result, a new `BINARY` column did not receive the `lance.blob.v2` Arrow extension metadata from the target table's persisted properties and was written as plain binary.

Using `ALTER TABLE ... SET TBLPROPERTIES` ensures the future column's encoding is persisted for both managed and unmanaged catalogs before `ADD COLUMNS` reloads the target table.

## Validation

- GitHub Actions: all checks passed
- integration coverage verifies blob v2 descriptor values after a successful `BINARY` backfill
- integration coverage verifies non-`BINARY` inputs fail before the invalid column is added
- existing non-blob add-columns scenarios remain covered

Local Maven, pytest, lint, and format checks were intentionally not run per the Lance repository workflow; GitHub Actions was used as the validation environment.

### DIFF
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -27,6 +27,36 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
 
 No table rewrite, no data movement—just a new column that is instantly queryable.
 
+## Adding a Blob v2 Column
+
+To add a blob v2 column, declare the future `BINARY` column's blob encoding when creating the
+target table and use Lance file format version `2.2` or higher. The column property may refer to a
+column that will be added later.
+
+```sql
+CREATE TABLE users (
+    id INT,
+    name STRING
+) USING lance
+TBLPROPERTIES (
+    'content.lance.encoding' = 'blob',
+    'file_format_version' = '2.2'
+);
+
+CREATE TEMPORARY VIEW content_backfill AS
+SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
+FROM users;
+
+ALTER TABLE users ADD COLUMNS content FROM content_backfill;
+```
+
+The source column must have Spark type `BINARY`. After the operation, reads expose `content` as a
+blob v2 descriptor struct, so descriptor fields can be queried without loading the bytes:
+
+```sql
+SELECT id, content.size, content.kind FROM users;
+```
+
 !!! note
     Because we use `_rowaddr` and `_fragid` to address the target dataset's rows for the new column's data, 
     the temporary view should contain `_rowaddr` and `_fragid`.
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -52,10 +52,12 @@ FROM users;
 ALTER TABLE users ADD COLUMNS content FROM content_backfill;
 ```
 
-The source column must have Spark type `BINARY`. Without the encoding property, `ADD COLUMNS`
-still succeeds and writes a plain `BINARY` column. With the encoding property set to `blob`, after
-the operation, reads expose `content` as a blob v2 descriptor struct, so descriptor fields can be
-queried without loading the bytes:
+The source column must have Spark type `BINARY`. The `content.lance.encoding` property must be
+persisted before `ADD COLUMNS` runs. If it is omitted, `ADD COLUMNS` still succeeds but writes a
+plain `BINARY` column. Setting the property after the column has been added does not retroactively
+convert that column to blob v2, so descriptor access such as `content.size` is not available.
+When the property is set to `blob` before the operation, reads expose `content` as a blob v2
+descriptor struct, so descriptor fields can be queried without loading the bytes:
 
 ```sql
 SELECT id, content.size, content.kind FROM users;
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -29,9 +29,9 @@ No table rewrite, no data movement—just a new column that is instantly queryab
 
 ## Adding a Blob v2 Column
 
-To add a blob v2 column, declare the future `BINARY` column's blob encoding when creating the
-target table and use Lance file format version `2.2` or higher. The column property may refer to a
-column that will be added later.
+To add a blob v2 column, create the target table with Lance file format version `2.2` or higher,
+then persist the future `BINARY` column's blob encoding with `ALTER TABLE SET TBLPROPERTIES` before
+adding the column.
 
 ```sql
 CREATE TABLE users (
@@ -39,9 +39,11 @@ CREATE TABLE users (
     name STRING
 ) USING lance
 TBLPROPERTIES (
-    'content.lance.encoding' = 'blob',
     'file_format_version' = '2.2'
 );
+
+ALTER TABLE users
+SET TBLPROPERTIES ('content.lance.encoding' = 'blob');
 
 CREATE TEMPORARY VIEW content_backfill AS
 SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -27,6 +27,10 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
 
 No table rewrite, no data movement—just a new column that is instantly queryable.
 
+!!! note
+    Because we use `_rowaddr` and `_fragid` to address the target dataset's rows for the new column's data,
+    the temporary view should contain `_rowaddr` and `_fragid`.
+
 ## Adding a Blob v2 Column
 
 To add a blob v2 column, create the target table with Lance file format version `2.2` or higher,
@@ -72,7 +76,3 @@ descriptor struct, so descriptor fields can be queried without loading the bytes
 ```sql
 SELECT id, content.size, content.kind FROM users;
 ```
-
-!!! note
-    Because we use `_rowaddr` and `_fragid` to address the target dataset's rows for the new column's data, 
-    the temporary view should contain `_rowaddr` and `_fragid`.
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -52,8 +52,10 @@ FROM users;
 ALTER TABLE users ADD COLUMNS content FROM content_backfill;
 ```
 
-The source column must have Spark type `BINARY`. After the operation, reads expose `content` as a
-blob v2 descriptor struct, so descriptor fields can be queried without loading the bytes:
+The source column must have Spark type `BINARY`. Without the encoding property, `ADD COLUMNS`
+still succeeds and writes a plain `BINARY` column. With the encoding property set to `blob`, after
+the operation, reads expose `content` as a blob v2 descriptor struct, so descriptor fields can be
+queried without loading the bytes:
 
 ```sql
 SELECT id, content.size, content.kind FROM users;
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -66,12 +66,15 @@ FROM users;
 ALTER TABLE users ADD COLUMNS content FROM content_backfill;
 ```
 
-The source column must have Spark type `BINARY`. The `content.lance.encoding` property must be
-persisted before `ADD COLUMNS` runs. If it is omitted, `ADD COLUMNS` still succeeds but writes a
-plain `BINARY` column. Setting the property after the column has been added does not retroactively
-convert that column to blob v2, so descriptor access such as `content.size` is not available.
-When the property is set to `blob` before the operation, reads expose `content` as a blob v2
-descriptor struct, so descriptor fields can be queried without loading the bytes:
+The source column must have Spark type `BINARY`. A blob v2 source column is not accepted because
+Spark reads it as a descriptor struct rather than `BINARY`; copy blob columns between existing
+tables with [INSERT INTO](insert-into.md#copying-blob-v2-columns) instead. The
+`content.lance.encoding` property must be persisted before `ADD COLUMNS` runs. If it is omitted,
+`ADD COLUMNS` still succeeds but writes a plain `BINARY` column. Setting the property after the
+column has been added does not retroactively convert that column to blob v2, so descriptor access
+such as `content.size` is not available. When the property is set to `blob` before the operation,
+reads expose `content` as a blob v2 descriptor struct, so descriptor fields can be queried without
+loading the bytes:
 
 ```sql
 SELECT id, content.size, content.kind FROM users;
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -33,6 +33,12 @@ To add a blob v2 column, create the target table with Lance file format version 
 then persist the future `BINARY` column's blob encoding with `ALTER TABLE SET TBLPROPERTIES` before
 adding the column.
 
+If the target table uses a file format below `2.2`, `ADD COLUMNS` still succeeds, but the `BINARY`
+column is written with legacy blob v1 encoding. Tables created without an explicit
+`file_format_version` may also use an older format, so set it to `2.2` or higher before adding the
+column. For more details, see
+[Blob v2 Writes](../../config.md#blob-v2-writes).
+
 ```sql
 CREATE TABLE users (
     id INT,
--- a/docs/src/operations/dml/add-columns.md
+++ b/docs/src/operations/dml/add-columns.md
@@ -45,6 +45,10 @@ TBLPROPERTIES (
 ALTER TABLE users
 SET TBLPROPERTIES ('content.lance.encoding' = 'blob');
 
+INSERT INTO users VALUES
+    (1, 'alpha'),
+    (2, 'bravo');
+
 CREATE TEMPORARY VIEW content_backfill AS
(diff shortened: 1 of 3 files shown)


hfutatzhanghb:
Hi, @fangbo  @LuciferYang  Could you please help review this PR when have free time? Thanks very much!

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,38 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
...
+);
+
+ALTER TABLE users
+SET TBLPROPERTIES ('content.lance.encoding' = 'blob');
+
+CREATE TEMPORARY VIEW content_backfill AS
+SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
+FROM users;
+
+ALTER TABLE users ADD COLUMNS content FROM content_backfill;
```
fangbo, line 56: I think

`
ALTER TABLE users ADD COLUMNS content FROM content_backfill WITH TBLPROPERTIES ('content.lance.encoding' = 'blob') 
`

is more ergonomic. @hfutatzhanghb @LuciferYang What do you think ?

hfutatzhanghb: @fangbo   Totally agree with you. That is more ergonomic. But i have a small question about the implementation complexity.  This PR's implementation is clear and simpler. We can dicuss about this point.

hfutatzhanghb: Hi @LuciferYang  . Could you please help review this when have free time? Thanks a lot!

geruh: Yeah I agree with @fangbo putting the property on the ADD COLUMNS statement is nicer but since it's additive and both paths are valid, it can be a follow up? Unless I'm missing something

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,38 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
...
+```
+
+The source column must have Spark type `BINARY`. After the operation, reads expose `content` as a
```
geruh: small nit: might be worth saying the behavior without:


The source column must have Spark type `BINARY`. Without the encoding property, `ADD COLUMNS `still succeeds and writes a plain `BINARY` column. After the operation, reads expose content as a blob v2 descriptor struct, so descriptor fields can be queried without loading the bytes:

hfutatzhanghb: very nice advice. Thanks a lot !

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,40 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
...
+);
+
+ALTER TABLE users
+SET TBLPROPERTIES ('content.lance.encoding' = 'blob');
+
+CREATE TEMPORARY VIEW content_backfill AS
+SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
+FROM users;
+
+ALTER TABLE users ADD COLUMNS content FROM content_backfill;
```
LuciferYang, line 56: This example runs the backfill against an empty table. With no source rows, the writer produces no fragments or merged schema, and `AddColumnsBackfillBatchWrite.commit()` returns early when the fragment list is empty. The command therefore finishes successfully without adding `content`, and the following `SELECT content.size` fails.

Could we insert a few sample rows before creating the view, as the integration test does, or support the schema-only empty-table case before presenting this sequence as runnable?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddColumnsBackfillExec.scala
```diff
@@ -55,15 +56,24 @@ case class AddColumnsBackfillExec(
...
+    val backfillSchema = SchemaConverter.processSchemaWithProperties(
+      actualQuery.schema,
+      tableProperties,
+      originalTable.getFileFormatVersion)
+
     val relation = DataSourceV2Relation.create(
       new LanceDataset(
         originalTable.readOptions(),
-        actualQuery.schema,
+        backfillSchema,
```
LuciferYang, line 68: This predates this PR, but the new blob-v2 backfill still goes through the existing `AddColumnsBackfillWrite` path: `managedVersioning` reaches the temporary `LanceDataset` and is then dropped before the batch write, so the driver commits through a plain `CommitBuilder`. A namespace-managed table can therefore bypass the namespace commit protocol, and some backends may reject the commit outright.

Could we track a follow-up to propagate the flag through the backfill writer and configure the commit with the namespace client, table ID, and `namespaceClientManagedVersioning(true)`, as `LanceBatchWrite` does?

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,40 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
 
 No table rewrite, no data movement—just a new column that is instantly queryable.
 
+## Adding a Blob v2 Column
+
+To add a blob v2 column, create the target table with Lance file format version `2.2` or higher,
+then persist the future `BINARY` column's blob encoding with `ALTER TABLE SET TBLPROPERTIES` before
```
LuciferYang, line 33: Could we document what happens when the encoding property is omitted or set too late? `ADD COLUMNS` still succeeds, but it creates a regular `BINARY` column. Setting the property afterward does not retroactively convert that column to blob v2, so descriptor access such as `content.size` will not work.

hfutatzhanghb: Very nice advice! Have fixed. Thanks a lot.

docs/src/operations/dml/add-columns.md
hfutatzhanghb: Got it. Add some sample rows before creating the view.  And I will support the schema-only empty-table case in other PR laterly. Thanks very much！

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddColumnsBackfillExec.scala
hfutatzhanghb: Tracked in https://github.com/lance-format/lance-spark/issues/771

hfutatzhanghb:
Hi, @LuciferYang @fangbo  All comments have been fixed. Could you please take a look again？ Thanks a lot!

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,46 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
 
 No table rewrite, no data movement—just a new column that is instantly queryable.
 
+## Adding a Blob v2 Column
+
+To add a blob v2 column, create the target table with Lance file format version `2.2` or higher,
```
LuciferYang, line 36: This section states the 2.2+ requirement but not what happens below it. `addBlobMetadata` picks v1 or v2 from the manifest's file format version, and anything below 2.2 silently gets blob v1, including tables created without `file_format_version`, which default to 2.1. Blob v1 writes fine at that version, so a user who skips `file_format_version` sees every statement succeed. `content` then reads back as plain BINARY and `content.size` fails during analysis, exactly like the omitted-property case documented below, with no way to tell which step was missed.

config.md already spells this out in the Blob v2 Writes section. Could we repeat it here, or link to it? If we would rather log a warning, it should sit on the ADD COLUMNS path rather than inside `addBlobMetadata`: creating blob v1 columns without a version is supported on purpose, and `BaseBlobCreateTableTest.testBlobColumnsStayBlobV1WithFileFormatVersion21` covers it, so a warning there would fire constantly.

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,46 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
...
+
+CREATE TEMPORARY VIEW content_backfill AS
+SELECT _rowaddr, _fragid, CAST(name AS BINARY) AS content
+FROM users;
+
+ALTER TABLE users ADD COLUMNS content FROM content_backfill;
+```
+
+The source column must have Spark type `BINARY`. The `content.lance.encoding` property must be
+persisted before `ADD COLUMNS` runs. If it is omitted, `ADD COLUMNS` still succeeds but writes a
```
LuciferYang: Thanks for adding these two sentences. They cover what was missing from the earlier revision.

One follow-on: neither behavior has a test behind it. The new test always sets the property, so "omitting it still succeeds with a plain BINARY column" and "setting it afterwards does not convert" are documented but unasserted, and none of the four existing add-column tests adds a BINARY column either.

Could we add a case without the property, asserting `DESCRIBE` shows `content` as binary rather than a struct, then set the property afterwards and check it still shows binary? One more variant would be worth pinning while you are in there: the property key has to match the added column's name exactly, so setting `content.lance.encoding` while the view aliases the column as `content_bytes` also succeeds silently as plain BINARY.

docs/src/operations/dml/add-columns.md
```diff
@@ -27,6 +27,46 @@ ALTER TABLE users ADD COLUMNS name_hash FROM tmp_view;
...
+plain `BINARY` column. Setting the property after the column has been added does not retroactively
+convert that column to blob v2, so descriptor access such as `content.size` is not available.
+When the property is set to `blob` before the operation, reads expose `content` as a blob v2
+descriptor struct, so descriptor fields can be queried without loading the bytes:
+
+```sql
+SELECT id, content.size, content.kind FROM users;
+```
+
 !!! note
```
LuciferYang: The new section landed between the generic example and this note, which pushed the note from line 30 down to here. The `_rowaddr`/`_fragid` requirement applies to every `ADD COLUMNS ... FROM`, not only the blob v2 form, so a reader who needs the plain form now stops after the first example and never reaches it, while a reader of the blob v2 section will take it for a blob v2 rule.

Moving the note back above the new heading would fix it. Giving the generic example its own `##` would work too.

integration-tests/test_lance_spark.py
```diff
@@ -2258,6 +2258,75 @@ def test_merge_into(self, spark):
...
+        )
+        assert "struct" in content_field.data_type.lower()
+
+        rows = spark.sql("""
+            SELECT id, content.size, content.kind
+            FROM default.test_table
+            ORDER BY id
+        """).collect()
+
+        assert [(row.id, row.size, row.kind) for row in rows] == [
```
LuciferYang, line 2320: Both values are 5 bytes with kind 0, so this assertion is blind to a backfill misalignment: if the `_rowaddr` join went wrong and the two rows' bytes were swapped, the expected (5, 0) tuples would still pass. The `assert "struct" in ...` check above accepts any struct type. And both rows come from a single INSERT, so there is no cross-fragment data and the `_fragid` side of the addressing never gets exercised.

Different-length values, say 'alpha' and 'bravo-charlie', would surface a misalignment right away. For the type check, `test_blob_v2_table_reads_content_as_descriptor` in this file asserts `"kind" in content_type` and `"blob_uri" in content_type`, which is stricter and copies straight over.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddColumnsBackfillExec.scala
```diff
@@ -55,15 +56,24 @@ case class AddColumnsBackfillExec(
       query
     }
 
+    val tableProperties = originalTable.properties()
+    val backfillSchema = SchemaConverter.processSchemaWithProperties(
```
LuciferYang, line 60: A question rather than a finding: is it intended that a blob v2 column cannot be the *source* of a backfill?

Reading a blob v2 column gives a descriptor struct on the Spark side, and `applyBlobV2DescriptorSchema` keeps the field metadata when it swaps the type. So `addBlobMetadata` matches the property, sees a non-BINARY type, and throws "must have BINARY type". That happens here in `run()`, before `AppendData` is built, which means `LanceBlobV2CopyThroughRule` never runs, and that rule is what handles descriptor to blob copies (it matches `AppendData` and keys on metadata rather than on type). So `INSERT INTO tgt SELECT id, data FROM src` between two blob v2 tables works and is covered by `BaseBlobV2CopyTest.insertSelect_copiesBlobBytes` plus about twenty sibling cases, while `ALTER TABLE tgt ADD COLUMNS data FROM src_view` is rejected outright.

I could not work out what happened before this change. The plan would have reached the rule back then, but the target schema was still descriptor-typed and `toArrowField`'s blob v2 case only matches `BinaryType`, so it may have failed further downstream anyway. I did not run either version.

If the rejection is deliberate, a line in the new docs section would save the next person the detour. If not, it probably wants its own issue.

LuciferYang (commented):
A second pass over the docs and the new test, plus one question about the source side. None of this blocks. The one I would like to see in this PR is the below-2.2 note, since it is a sentence added to the section you just wrote. The namespace-managed commit point from my earlier review is tracked in #771, so I have left it out here.

docs/src/operations/dml/add-columns.md
hfutatzhanghb: Have fixed in [7aaf05e](https://github.com/lance-format/lance-spark/commit/7aaf05e2d29571992aaf6261c910008fc0e15132).  Thanks for pointing this.

I added a short note to the `ADD COLUMNS` Blob v2 section saying that, with a file format below `2.2`, the operation still succeeds but writes the `BINARY` column with legacy blob v1 encoding. It also mentions that tables created without an explicit `file_format_version` may use an older format, and links to the [Blob v2 Writes](https://github.com/lance-format/lance-spark/blob/main/docs/src/config.md) section for the full behavior.

I didn't add a warning in `addBlobMetadata`. Legacy blob v1 encoding is intentionally supported there, so warning on that path would create noise for valid v1 writes.

docs/src/operations/dml/add-columns.md
hfutatzhanghb: Thanks for the very nice advice! Have Done in [270200d](https://github.com/lance-format/lance-spark/commit/270200ddbb815b4e46cf5dbeb68bfa695a26a97a).

I added two focused integration tests:
- Without the encoding property, `ADD COLUMNS` succeeds and `DESCRIBE` reports `binary`; setting the property afterward still leaves it as `binary`.
- Setting `content.lance.encoding` while adding `content_bytes` also leaves the column as `binary`, pinning the exact-name matching requirement.

docs/src/operations/dml/add-columns.md
hfutatzhanghb: have fixed. Thanks!

integration-tests/test_lance_spark.py
hfutatzhanghb: Learn a lot here! Thanks.  Done in [6dce408](https://github.com/lance-format/lance-spark/commit/6dce408c6ddeb5dd96d647f6881df06207a67e8e).

I used different-length values (`alpha` and `bravo-charlie`) and split them into two inserts so the data can land in different fragments. I also strengthened the `DESCRIBE` assertion to require `kind` and `blob_uri`, matching the descriptor-struct test. The size/kind assertions now detect a row-address misalignment between the two rows.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddColumnsBackfillExec.scala
hfutatzhanghb: Thanks for digging into this. It is currently an intentional limitation, and I have documented it in [8e18cfa](https://github.com/lance-format/lance-spark/commit/8e18cfa3edd8c7d2c5ac9b389c829c2a65c8b703): blob v2 source columns are not accepted because Spark reads them as descriptor structs rather than `BINARY`. The docs now recommend using `INSERT INTO ... SELECT` to copy blob v2 columns between existing tables.

I also opened a follow-up to evaluate whether `ADD COLUMNS FROM` should support blob v2 source columns with a copy-through path: https://github.com/lance-format/lance-spark/issues/775.

geruh (approved):

hfutatzhanghb:
Hi, @LuciferYang   Have fixed all problem you suggested, Could you please take a look at this again when have free time ? Thanks a lot !!!

LuciferYang (approved):
Thank you @hfutatzhanghb

LuciferYang:
Merged into main. Thanks @hfutatzhanghb @fangbo and @geruh

PR #766, opened 2026-08-14
fix: respect an FTS predicate in COUNT(*) aggregate pushdown

### LINKED ISSUE #765: COUNT(*) ignores an FTS predicate and returns the whole table's row count
### Summary

`SELECT count(*) FROM t WHERE lance_match(body, 'hello')` returns the table's **total** row count instead of the number of FTS matches.

### Root cause

The FTS Catalyst rule moves the predicate out of the `Filter` and into the relation options, so nothing is left in `pushedPredicates`. Two places decide "is anything restricting the rows?" by looking only at `pushedPredicates`, and are therefore blind to an active full-text query:

1. `LanceScanBuilder.pushAggregation` — takes the metadata shortcut whenever `pushedPredicates.length == 0`, answering from `ManifestSummary.getTotalRows()`.
2. `LanceCountStarPartitionReader.computeCount` — builds its `ScanOptions` without `fullTextQuery`, so even the scan-based count path counts unfiltered rows.

`LanceScan.pruneByLimit` already guards on `getFullTextQuery() != null` for exactly this reason; the two count paths did not.

### Reproduction

Twenty rows, fifteen of which contain `hello`:

```sql
SELECT count(*) FROM t WHERE lance_match(body, 'hello');
-- expected 15, actual 20

SELECT count(*) FROM t WHERE lance_match(body, 'zzz_no_such_term');
-- expected 0, actual 20

SELECT count(*) FROM t WHERE lance_match(body, 'hello') AND id >= 10;
-- expected 5, actual 10
```

The first two go through the metadata shortcut; the third goes through the scan-based count path, so both paths are affected.

### Affected

Every namespace on which the aggregate pushes down — the shortcut is chosen before any namespace-specific routing.

### Scope note

A `COUNT(*)` whose other filters cannot push down leaves `pushedAggregation` empty and is routed to `queryTable`, which drops a structured full-text query on `dir`. That case, and the row query on the same namespaces, remain unfiltered and are **not** covered by this fix — see the Known Limitations entry in `docs/src/operations/dql/fts.md`.

LuciferYang:
Closes #765.

## What

`LanceScanBuilder.pushAggregation` decided `COUNT(*)` could be answered from `ManifestSummary.getTotalRows()` whenever no predicate was pushed, without checking `readOptions.getFullTextQuery()`. Because the FTS rule moves the predicate out of the `Filter` and into the relation options, a query like

```sql
SELECT count(*) FROM t WHERE lance_match(body, 'hello')
```

has no `Filter` and no pushed predicate, so it returned the table's total row count instead of the number of FTS matches. The same omission in `LanceCountStarPartitionReader.computeCount` made the scan-based count path (FTS combined with a pushable scalar filter) ignore the FTS query too.

`LanceScan.pruneByLimit` already guards on `getFullTextQuery() != null` for exactly this reason; both count paths now do the same. The count scan also opts out of Lance's `_score` autoprojection, which an empty column list would otherwise trigger once per task; the row-scan path deliberately keeps that autoprojection, since it is how `_score` is delivered.

## Scope

This fixes the `COUNT(*)` that pushes down. `build()`'s namespace FTS branch requires no pushed aggregation, so a pushed-down `COUNT(*)` is never routed to a namespace's `queryTable` and is now answered by the local per-fragment scan on every namespace. On catalog-only namespaces (Glue/Hive/Iceberg) the row query already applied the predicate through that same scan, so the count now agrees with it.

Two cases stay wrong and are out of scope, both recorded under Known Limitations in `docs/src/operations/dql/fts.md`:

- on namespaces served by `queryTable` — `dir`, and REST implementations that skip a structured full-text query — the row query still returns every row, so it can disagree with the count;
- a `COUNT(*)` whose other filters cannot push down leaves `pushedAggregation` empty, which routes it through `queryTable` like a row query and leaves it unfiltered there as well.

## Tests

Three cases in `BaseFtsCatalogOnlyNamespaceTest` cover the metadata shortcut (matching and non-matching terms) and the scan-based path; all three fail on the unfixed tree (20/20/10 instead of 15/0/5). The first also asserts the executed plan's scan output list is `[count#N]`, which ties the count to the scan rather than to re-aggregated rows — matching only the word `count` does not, since the un-pushed plan contains `count(1)` as well. A unit test in `LanceScanBuilderTest` covers the other half: with a full-text query set, `build()` must return `LanceScan`, not `LanceLocalScan`.

The fixture's javadoc claimed "Two fragments" while each `INSERT` writes one fragment per write task (four under `local[4]`, so eight in total, 2-3 rows each). That comment misled two rounds of analysis during review, so it is corrected here to state the real layout and warn that per-fragment pushdown reasoning must use eight.

## Follow-up from review

Routing FTS-only `COUNT(*)` through `LanceCountStarPartitionReader` exposed a pre-existing gap: unlike `LanceFragmentScanner.create()`, `computeCount()` opened the dataset with only `initialStorageOptions` and never rebuilt the namespace, so on namespaces that vend short-lived credentials (Iceberg REST, Polaris, Unity) a long FTS count could fail after the driver-fetched credentials expired. `f1086cc` applies the same enabled/disabled refresh gate before the open, keeping the `executor_credential_refresh=false` opt-out that Kerberized HMS catalogs rely on, with `LanceCountStarPartitionReaderTest` covering both flag values.

Verified on `lance-spark-3.4_2.12`: full suite 1196 run, 0 failures, 113 skipped (the skipped set includes three FTS suites disabled via `assumeTrue(false)`).


LuciferYang:
cc @fangbo

fangbo (approved):
Great fix !

LuciferYang:
Thank you @fangbo

Issue #713, opened 2026-07-22
Support DQL/DML for Branch and Tag


fangbo:
Since we have implemented DDL for Branch and Tag in [PR#576](https://github.com/lance-format/lance-spark/pull/576) and [PR#654](https://github.com/lance-format/lance-spark/pull/654). We can continue pushing forward and add support for DQL/DML on branches and tags.

There are a lot discussions in this [PR#198](https://github.com/lance-format/lance-spark/pull/198). As I understand it, one consensus we reached is: it's a bit more ergonomic to centralize branches/tags in the table ID. My idea is that we can treat branch/tag as a normal table and use `__branch__` or `__tag__` as the separator. 

For branch:

```sql
SELECT * FROM <table_name>__branch__<branch_name>
UPDATE <table_name>__branch__<branch_name> SET ...
DELETE FROM <table_name>__branch__<branch_name>
```

For tag:

```sql
SELECT * FROM <table_name>__tag__<tag_name>
```

@hamersaw @LuciferYang I very much look forward to your thoughts on this topic.

LuciferYang:
Thanks for pushing this forward, @fangbo. I re-read the #198 thread against the current write/read paths on `main`, and I'm on board with the table-ID direction. The consensus there still holds up: `VERSION AS OF` is read-only and Spark won't route `UPDATE`/`DELETE`/`MERGE` through it, whereas encoding the ref in the table identifier works for both DQL and DML and is much friendlier for no-code callers that can't rewrite SQL. Modeling a branch as "just another table" also matches how your customer is already using it.

Where `main` stands today, to scope the work:
- DDL is done. `CREATE/DROP BRANCH|TAG` and `SHOW ...` land through `LanceCreateBranchExec` etc. and call `dataset.createBranch(name, Ref)` / `branches().delete(...)`. lance-core already gives us `Ref.ofBranch/ofTag/ofMain`, `Dataset.checkout(Ref)`, and `checkoutTag(...)`.
- The read path has no ref entry yet. `ReadOptions.Builder` only exposes `setVersion(long)`, and `Utils.OpenDatasetBuilder` threads only `version`, so nothing carries a branch/tag/ref into scan planning.
- Ref resolution is numeric-only. `loadTableInternal` / `loadTableFromPath` resolve the version via `Utils.parseVersion` (`Long.parseUnsignedLong`), so a string tag isn't wired for reads yet either.
[...]

PR #770, opened 2026-08-17
refactor: refactor Spark options to use LanceRef reference instead of raw version number

fangbo:
Related https://github.com/lance-format/lance-spark/issues/713

### Description
This PR refactors Spark read and write options to use LanceRef instead of passing around raw version numbers directly. The goal is to make reference handling more explicit and extensible, especially for future branch and tag support.

The change updates the Spark option model and related call sites to carry reference information through the connector, while preserving the ability to resolve snapshot versions where needed. This lays the groundwork for consistent handling of main, versioned, branch, and tag references across read, scan, blob/search, and write paths.

### Summary

1. Replace raw version-based Spark options with LanceRef.
2. Update read/write option APIs and affected Spark call sites.
3. Align test utilities with the new reference model.
4. Prepare the connector for branch/tag-aware reference handling.


geruh (commented):
If I'm following this we are going to 

```java
class LanceSparkWriteOptions {
    LanceRef ref; 
}
```

from what would look like:

```java
class LanceSparkWriteOptions {
    Long version;
    String branch;
    String tag;
}
```


Do we actually need the Spark specific `LanceRef` here? It seems like we could keep these as option fields and only build `org.lance.Ref` at the open boundary.

That also seems simpler for future branch support, since resolving feature to version 3 just updates the version while keeping the branch, instead of having to reconstruct the ref as feature@3. Unless im missing something.

fangbo:
> If I'm following this we are going to
> 
> ```java
> class LanceSparkWriteOptions {
>     LanceRef ref; 
> }
> ```
> 
> from what would look like:
> 
> ```java
> class LanceSparkWriteOptions {
>     Long version;
>     String branch;
>     String tag;
> }
> ```
> 
> Do we actually need the Spark specific `LanceRef` here? It seems like we could keep these as option fields and only build `org.lance.Ref` at the open boundary.
> 
> That also seems simpler for future branch support, since resolving feature to version 3 just updates the version while keeping the branch, instead of having to reconstruct the ref as feature@3. Unless im missing something.

@geruh Thanks very much for your suggestion.

I think using `LanceRef` has several benefits:

1. `LanceRef` can represent multiple scenarios, including `main/version`, `branch/version`, and `tag`.
2. When we support `branch/version` and `tag` in the future, we won’t need to modify `Utils.OpenDatasetBuilder`, `LanceSparkReadOptions`, or `LanceSparkWriteOptions`.
3. It keeps the concept aligned with `Ref` in `lance-core`.

There are also some discussions and strategies here https://github.com/lance-format/lance-spark/issues/713#issuecomment-5054951538

geruh:
Okay makes sense, thanks for clarifying!

geruh (approved):

PR #772, opened 2026-08-18
fix: propagate managedVersioning through backfill commits

### LINKED ISSUE #771: Follow-up: propagate managedVersioning through ADD COLUMNS backfill commit path
## Context
Follow-up from https://github.com/lance-format/lance-spark/pull/764#discussion_r3801092935.

## Problem
`AddColumnsBackfillExec` creates a temporary `LanceDataset` and passes `originalTable.getManagedVersioning`, but the backfill write path does not carry that flag through `AddColumnsBackfillWrite` / `AddColumnsBackfillBatchWrite`.

`AddColumnsBackfillBatchWrite.commit()` commits with a plain `CommitBuilder`:

```java
new CommitBuilder(dataset)
    .writeParams(...)
    .execute(txn)
```

Unlike `LanceBatchWrite`, it does not configure:

```java
.namespaceClient(namespace)
.tableId(tableId)
.namespaceClientManagedVersioning(true)
```

As a result, `ALTER TABLE ... ADD COLUMNS ... FROM ...` on namespace-managed tables can bypass the namespace-managed versioning protocol. Some namespace backends may reject the commit.

This pre-exists PR #764, but the new blob-v2 `ADD COLUMNS` workflow exercises the same path.

## Expected fix
- Propagate `managedVersioning` from the temporary `LanceDataset` into `AddColumnsBackfillWrite` and `AddColumnsBackfillBatchWrite`.
- In `AddColumnsBackfillBatchWrite.commit()`, when `managedVersioning` is true, configure `CommitBuilder` with the namespace client, table ID, and `namespaceClientManagedVersioning(true)`, matching `LanceBatchWrite`.
- Preserve existing non-managed behavior.

## Related files
- `lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddColumnsBackfillExec.scala`
- `lance-spark-base_2.12/src/main/java/org/lance/spark/write/AddColumnsBackfillWrite.java`
- `lance-spark-base_2.12/src/main/java/org/lance/spark/write/AddColumnsBackfillBatchWrite.java`
- `lance-spark-base_2.12/src/main/java/org/lance/spark/write/LanceBatchWrite.java`

hfutatzhanghb:
## Summary
Closes #771

Propagate `managedVersioning` from `LanceDataset` into `AddColumnsBackfillWrite` and `AddColumnsBackfillBatchWrite`, and configure namespace-managed commits in the backfill path the same way `LanceBatchWrite` does.

## Changes
- Pass `managedVersioning` from `LanceDataset.newWriteBuilder` to `AddColumnsBackfillWrite.AddColumnsWriteBuilder`.
- Carry the flag through `AddColumnsBackfillWrite` to `AddColumnsBackfillBatchWrite`.
- Configure `CommitBuilder` with namespace client, table ID, and `namespaceClientManagedVersioning(true)` when `managedVersioning` is true.
- Add a focused unit test for builder-to-batch-write propagation.
- Update the existing `BaseAddColumnsBackfillTest` constructor call.

## Validation
- `./mvnw spotless:check -pl lance-spark-3.5_2.12 -am -DskipTests` passed.
- Main and test sources compile successfully.
- Local unit-test execution is blocked on this macOS x86_64 host because the Lance JNI native library is missing (`cannot find /nativelib/darwin-x86-64/liblance_jni.dylib`).


hfutatzhanghb:
Hi, @LuciferYang   @fangbo . Could you please help review this PR when have free time? Thanks a lot,

geruh (approved):
Thanks @hfutatzhanghb this LGTM, just one small nit! 

Also, looks like UpdateColumns needs this since the `UpdateColumnsWriteBuilder` doesn't take the flag unless I'm missing something.

hfutatzhanghb:
> Thanks @hfutatzhanghb this LGTM, just one small nit!
> 
> Also, looks like UpdateColumns needs this since the `UpdateColumnsWriteBuilder` doesn't take the flag unless I'm missing something.

Great catch! Thanks for pointing this.  managedVersioning was not propagated through UpdateColumnsWriteBuilder, so UPDATE COLUMNS had the same issue. I've now updated the builder/write/batch-write chain and added corresponding test coverage. Thanks very much!

integration-tests/test_lance_spark.py
```diff
@@ -2443,10 +2443,89 @@ def test_add_column_computed_values(self, spark):
         assert result[1].total_compensation == 69000   # 60000 + 9000
         assert result[2].total_compensation == 84000   # 70000 + 14000
 
+    @pytest.mark.requires_rest
+    @pytest.mark.rest_dir_compatible
+    def test_add_column_namespace_backend(self, spark, test_table):
+        """Test ADD COLUMNS FROM against a namespace-managed table."""
```
geruh: meant to post this but didnt push send lol 🤦‍♂️

small nit: I think this should be a bit more explicit then you can drop the doc string test. Maybe something like `test_add_column_from_view_on_rest`

but now we can apply this to the update integ test as well

hfutatzhanghb: 😂 Have fixed !  Thanks very much!

geruh (approved):
Awesome thanks @hfutatzhanghb! Will merge once the tests pass

hfutatzhanghb:
@geruh Thanks very much for reviewing and merging!

LuciferYang:
These two tests never reach the new `if (managedVersioning)` branch: the default workflow only runs local/azurite/minio (where `requires_rest` skips), and the rest-dir dir adapter does not report managed versioning by default, so managedVersioning stays false. The core behavior this PR fixes ships with no test coverage.

Two things are needed to cover it: add `table_version_tracking_enabled=true` to LanceRestDirNamespaceServer's backendConfig so rest-dir reports managed=true (the key exists in the lance-core 11.0.0-beta.10 native lib — verified via strings; the effect of flipping it is untested; it affects every test on that server, so a separate managed instance is safer), and add these tests to a workflow that runs rest-dir — the only one today, spark-search, collects just TestDQLSearchTableFunctions. Otherwise, fix the coverage claims in the description and the commit message: these tests never run in CI, and a manual rest-dir run only covers the managed=false path.

hfutatzhanghb:
> These two tests never reach the new `if (managedVersioning)` branch: the default workflow only runs local/azurite/minio (where `requires_rest` skips), and the rest-dir dir adapter does not report managed versioning by default, so managedVersioning stays false. The core behavior this PR fixes ships with no test coverage.
> 
> Two things are needed to cover it: add `table_version_tracking_enabled=true` to LanceRestDirNamespaceServer's backendConfig so rest-dir reports managed=true (the key exists in the lance-core 11.0.0-beta.10 native lib — verified via strings; the effect of flipping it is untested; it affects every test on that server, so a separate managed instance is safer), and add these tests to a workflow that runs rest-dir — the only one today, spark-search, collects just TestDQLSearchTableFunctions. Otherwise, fix the coverage claims in the description and the commit message: these tests never run in CI, and a manual rest-dir run only covers the managed=false path.

@LuciferYang . Got it. Thanks for catching this.  I'll add a dedicated managed rest-dir configuration, verify that table_version_tracking_enabled=true makes the namespace report managed versioning, and explicitly run these two DML tests in CI.

PR #776, opened 2026-08-21
feat: support branch reads

geruh:
related to #713

#773 added tag reads through `VERSION AS OF`. This is the branch reads follow up.

You can read a named branch the same way other Spark projects do:

```sql
SELECT * FROM catalog.db.users.branch_audit;
```

```scala
spark.read.option("branch", "audit").table("catalog.db.users")
```

```python
spark.read.format("lance") \
    .option("branch", "audit") \
    .load("/path/to/dataset.lance")
````

In this case table.branch_audit is the real branch table, including that branch's schema. .option("branch") is scan time convenience and Spark still analyzes with the current table schema, so use the identifier when the branch schema has diverged.

If a real table named users.branch_audit already exists, that table wins. Don't combine a branch identifier with VERSION AS OF / TIMESTAMP AS OF, and don't set branch and version on the same read. Current;y Branches are read only here. Writes stay a follow-up. 

cc: @fangbo

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java
@@ -32,6 +32,7 @@ import org.apache.spark.sql.connector.catalog.StagedTable;
 import org.apache.spark.sql.connector.catalog.SupportsMetadataColumns;
 import org.apache.spark.sql.connector.catalog.SupportsRead;
 import org.apache.spark.sql.connector.catalog.SupportsWrite;
+import org.apache.spark.sql.connector.catalog.Table;
 import org.apache.spark.sql.connector.catalog.TableCapability;
 import org.apache.spark.sql.connector.read.ScanBuilder;
 import org.apache.spark.sql.connector.write.LogicalWriteInfo;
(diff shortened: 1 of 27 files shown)

### CODE REVIEW

**Comment by fangbo** [comment]
@geruh  Should we support read a branch's specific version ?

File: lance-spark-base_2.12/src/main/java/org/lance/spark/LanceDataset.java [hunk 1]
```diff
@@ -309,33 +313,72 @@ public String getFileFormatVersion() {
...
+    LanceRef scanRef = scanOptions.getRef();
+    if (scanRef != null && scanRef.isBranch() && scanRef.getVersionNumber().isEmpty()) {
+      try (Dataset dataset =
+          Utils.openDatasetBuilder(scanOptions)
+              .initialStorageOptions(initialStorageOptions)
+              .runtimeNamespace(namespaceImpl, namespaceProperties, readOptions.getTableId())
+              .build()) {
+        scanOptions = scanOptions.withRef(Utils.pinOpenedRef(dataset, scanRef));
+        scanSchema = LanceArrowUtils.fromArrowSchema(dataset.getSchema());
+      }
```
fangbo, line 329: Great fix ! 

On the other hand, is it necessary to use `scanSchema` for tag read ?

geruh: tags don't hit this right now. it's only for unpinned `.option("branch")`

geruh:
> @geruh Should we support read a branch's specific version ?

nah not this pr, right now we just need branch reads that points at the current head. Adding version or tags to branches can be a followup.

jackye1995 (approved):
looks good to me!

PR #715, opened 2026-07-22
feat: add task commit factory for Apache Gluten writes

malinjawi:
## Summary
- add `LanceBatchWrite.taskCommit(List<FragmentMetadata>)` so an external columnar writer such as Apache Gluten can produce a commit message accepted by `LanceBatchWrite.commit(...)`
- keep `TaskCommit` and its fragment accessor package-private; the factory returns the opaque `WriterCommitMessage`
- defensively copy the list and reject `null`, as suggested in the design discussion
- add API-visibility, serialization round-trip, and multi-fragment commit tests

Design discussion: https://github.com/lance-format/lance-spark/discussions/709 (agreed by @LuciferYang and @hamersaw)
Consumer: https://github.com/apache/gluten/issues/12263

## Testing
- `make lint`
- `./mvnw test -pl lance-spark-base_2.12 -Dtest=LanceBatchWriteTest,LanceBatchWriteTaskCommitPublicApiTest`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #623, opened 2026-06-12
Expose a stable native scan descriptor for Lance Spark reads and datafusion-comet integration


wirybeaver:
## Problem

DataFusion Comet and other native Spark accelerators need a stable way to understand a Spark-planned Lance table scan without depending on Lance Spark internals. Today the physical plan exposes `org.lance.spark.read.LanceScan`, but there is no public descriptor that captures the scan contract a native reader would need.

The first native-read integration should let Spark keep planning Lance tables through Lance Spark, then let an optional native engine extract a small, explicit read descriptor for ordinary table reads.

## Proposed solution

Add a stable native scan descriptor API for ordinary Lance reads:

- Add serializable `LanceNativeScanPlan` and `LanceNativeSplit` descriptor types.
- Expose `LanceScan.nativeScanPlan()` for supported ordinary table scans.
- Include dataset URI, resolved dataset version, Spark/projected schema JSON, projected columns, pushed filter SQL, limit/offset, batch size, storage options, and per-partition fragment IDs.
- Preserve storage option precedence from the planned Lance scan.
- Return explicit fallback reasons when a native descriptor cannot represent the scan.

Initial v1 scope is intentionally narrow: ordinary table reads only. Search, hybrid search, index-aware execution, aggregation pushdown, namespace-backed credential refresh, and other advanced paths should remain fallback until their semantics can be described and tested explicitly.

## Architecture sketch

The important boundary is that Lance Spark remains the Spark planning authority. A native consumer such as DataFusion Comet should not infer Lance planning semantics from private fields or re-plan the table. It should ask the final Spark `LanceScan` for an explicit descriptor.

```text
Spark SQL / DataFrame read.format("lance")
        |
        v
+--------------------------------+
| Spark DataSource V2 planning   |
| Lance table/catalog resolution |
+----------------+---------------+
                 |
                 v
+--------------------------------+
| lance-spark                     |
| LanceScanBuilder                |
| - resolves dataset URI/version  |
| - resolves schemas/options      |
| - applies projection/filter     |
| - computes stats and splits     |
| - records fallback-only state   |
+----------------+---------------+
                 |
                 | build()
                 v
+--------------------------------+
| org.lance.spark.read.LanceScan |
| Spark Batch + Scan object       |
| stored inside BatchScanExec     |
+----------------+---------------+
                 |
                 | nativeScanPlan()
                 | Optional[LanceNativeScanPlan]
                 v
+--------------------------------+        unsupported v1 feature
[...]

PR #778, opened 2026-08-23
feat(gluten): export a Lance fragment scan as an Arrow C stream

sezruby:
## What

Adds `LanceArrowStreamScanner`, which plans a Lance fragment scan and exports it as an Arrow C Data Interface stream (`ArrowArrayStream`) for native consumers such as Apache Gluten / Velox.

## Why

This is the read-side building block for offloading Lance scans to a native engine. Only the `ArrowArrayStream` C-struct address crosses the JVM/native boundary, so the consumer's Arrow build and classloader do not need to match lance-spark's — Gluten builds Arrow 15 to match Spark 3.5 / Velox, while the Lance Java SDK is on Arrow 18. Passing the raw struct address sidesteps that mismatch entirely.

## How

- `LanceArrowStreamScanner.export(fragmentId, inputPartition)` returns a `LanceArrowStream` handle exposing `stream()` / `streamAddress()`.
- All scan planning (column projection, filter pushdown, limit/offset, row-id / row-address, batch size) is delegated to the existing `LanceFragmentScanner`, so the exported stream yields exactly the same rows in the same order as the Spark columnar reader.
- The Lance native core populates the caller-owned stream directly via `LanceScanner#exportArrowStream(long)` (lance-format/lance#7259), so no Arrow data is materialized on the JVM heap on this path.
- `LanceArrowStream` owns the exported stream and the scan behind it; closing it releases the native scan (through the stream's release callback) and then the scanner and dataset handles.

## Dependency

`exportArrowStream(long)` first ships in lance-core `11.0.0-beta.21`, so this bumps `lance.version` from `11.0.0-beta.10` (isolated in its own commit).

## Testing

- New `LanceArrowStreamScannerTest`: exports each fragment of the bundled test table, re-imports it on the JVM (standing in for a native consumer), and asserts the rows match the columnar reader — run under the leak-checking allocator to verify the export/scanner lifecycle releases cleanly.
- Existing `LanceFragmentColumnarBatchScannerTest` still passes (regression on the touched `LanceFragmentScanner`).

## Related

- Complements #715 (Gluten write path).
- Distinct mechanism from #624 / #623 (serializable native scan descriptor for datafusion-comet): this exposes a live Arrow C stream handle rather than a re-plannable descriptor.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
@@ -18,7 +18,7 @@ import org.lance.spark.read.LanceInputPartition;
 
 import org.apache.arrow.c.ArrowArrayStream;
 import org.apache.arrow.vector.types.pojo.Field;
-import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.util.LanceArrowUtils;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -119,27 +119,64 @@ public final class LanceArrowStreamScanner {
    * synthesizing {@code _fragid}, dropping the {@code _score} a full-text query auto-projects,
    * stripping the {@code _rowaddr} added for blobs, surfacing {@code _rowid} for an empty
    * projection, reordering row-version columns — but this zero-copy export cannot, so any mismatch
-   * in field names or order must fall back to the columnar reader.
+   * must fall back to the columnar reader.
+   *
+   * <p>The comparison covers field names, order, Arrow types, and nullability. The declared Spark
+   * schema is converted to Arrow through {@link LanceArrowUtils} — the same adapter the read path
+   * uses — so it carries the Arrow type distinctions a Spark {@code DataType} alone cannot express
+   * (e.g. {@code LargeUtf8} vs {@code Utf8}, {@code LargeBinary}, {@code Date(MILLISECOND)}, {@code
+   * FixedSizeBinary}, {@code Float16}), which {@code fromArrowSchema} recorded in field metadata
+   * and {@code toArrowSchema} restores here. Checking the {@code ArrowType} (not just the name)
+   * stops a column whose native type differs from the declared one — e.g. a narrower/wider int or a
+   * different time-zone timestamp — from being streamed to the native consumer as if it matched.
    */
   private static void checkNativeSchemaMatchesPartition(
       LanceFragmentScanner fragmentScanner, LanceInputPartition inputPartition) {
-    List<String> declared = new ArrayList<>();
-    for (StructField field : inputPartition.getSchema().fields()) {
-      declared.add(field.name());
-    }
-    List<String> nativeColumns = new ArrayList<>();
-    for (Field field : fragmentScanner.schema().getFields()) {
-      nativeColumns.add(field.getName());
-    }
-    if (!declared.equals(nativeColumns)) {
+    List<Field> declared =
+        LanceArrowUtils.toArrowSchema(inputPartition.getSchema(), "UTC", true).getFields();
+    List<Field> nativeFields = fragmentScanner.schema().getFields();
+    if (!fieldsMatch(declared, nativeFields)) {
       throw new UnsupportedOperationException(
-          "Arrow C stream export requires the native scan schema to match the partition schema, "
-              + "which this zero-copy export cannot re-project or reorder. Declared "
-              + declared
+          "Arrow C stream export requires the native scan schema to match the partition schema "
+              + "(field names, order, Arrow types, and nullability), which this zero-copy export "
+              + "cannot re-project, reorder, or cast. Declared "
+              + describe(declared)
               + " but the native scan produced "
-              + nativeColumns
+              + describe(nativeFields)
               + ". Fall back to the columnar reader for this partition.");
     }
+  }
+
+  /**
+   * Compares two Arrow field lists by name, order, {@code ArrowType} (value equality over width,
+   * precision, time zone, etc.), and nullability, recursing into children so nested list/struct
+   * element types are checked too. Field-level metadata and dictionary encodings are intentionally
+   * ignored: they carry lance-internal markers, not a difference the raw stream cannot reproduce.
+   */
+  private static boolean fieldsMatch(List<Field> declared, List<Field> actual) {
+    if (declared.size() != actual.size()) {
+      return false;
+    }
+    for (int i = 0; i < declared.size(); i++) {
+      Field d = declared.get(i);
+      Field a = actual.get(i);
+      if (!d.getName().equals(a.getName())
+          || d.isNullable() != a.isNullable()
+          || !d.getType().equals(a.getType())
+          || !fieldsMatch(d.getChildren(), a.getChildren())) {
+        return false;
+      }
+    }
+    return true;
+  }
+
+  private static String describe(List<Field> fields) {
+    List<String> parts = new ArrayList<>(fields.size());
+    for (Field field : fields) {
+      String nullability = field.isNullable() ? " null" : " notnull";
+      parts.add(field.getName() + " " + field.getType() + nullability);
+    }
+    return parts.toString();
   }
 
   private static void closeQuietly(AutoCloseable closeable) {
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
@@ -13,12 +13,18 @@
  */
 package org.lance.spark.internal;
 
+import org.lance.spark.LanceConstant;
 import org.lance.spark.LanceRuntime;
 import org.lance.spark.read.LanceInputPartition;
+import org.lance.spark.utils.BlobUtils;
 
 import org.apache.arrow.c.ArrowArrayStream;
+import org.apache.spark.sql.types.StructField;
+import org.apache.spark.sql.types.StructType;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Exports a Lance fragment scan as an Arrow C Data Interface stream ({@link ArrowArrayStream}) for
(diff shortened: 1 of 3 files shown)


sezruby:
Addressed all three findings in e914259.

**Release callback (line 119).** `LanceArrowStream.close()` now runs the C stream release callback via a guarded `releaseStream()` before freeing the struct, so an abandoned or partially-consumed stream drops the provider's `private_data` and record-batch stream. It is idempotent against a consumer that already imported the stream: `Data.importArrayStream` snapshots the callback and closes our struct, so `releaseStream()` skips when `snapshot()` reports the struct is freed or the release address is `NULL` per the C ABI. Added `closeReleasesStreamThatWasNeverImported`, which closes the handle with no JVM import — under the leak-checking allocator it fails if the callback or struct is left dangling.

**Schema contract (line 60).** `export()` now calls `checkExportableSchema()` first and rejects, with `UnsupportedOperationException`, the shapes the columnar reader fixes up on the JVM after import: the synthesized `_fragid`, an empty projection (which surfaces the internal `_rowid`), metadata columns (`_rowid` / `_rowaddr` / row-version / `_score`), and blob columns. Those cannot be reproduced on the zero-copy native path without materializing on the JVM heap, so the caller falls back to the columnar reader instead of receiving a wrong schema. Ordinary data projections still export a stream that matches the partition schema (Lance has no Hive-style partition columns). Added `exportRejectsSchemasNeedingJvmPostProcessing`.

**Construction leak (line 54).** `ArrowArrayStream.allocateNew` now runs inside the cleanup scope, so a throwing allocation (e.g. a bounded allocator) closes both the stream and the scanner/dataset opened by `LanceFragmentScanner.create()`.

`LanceArrowStreamScannerTest` (3/3) and the `LanceFragmentColumnarBatchScannerTest` regression (1/1) pass; `spotless:check` is clean.

sezruby:
Addressed the pushed-aggregation finding in f78423b.

`checkExportableSchema` is now `checkExportablePartition`, which rejects a partition with a present `pushedAggregation` (e.g. `COUNT(*)`) before the schema checks. Such a partition is served by a dedicated reader (`LanceCountStarPartitionReader`) that returns a single `count` column, whereas `LanceFragmentScanner` ignores `pushedAggregation` and scans data rows — so the raw export would emit the wrong output. Filter, limit, offset, and top-N ordering stay exportable because they are pushed faithfully into the native scan.

Added `exportRejectsPushedAggregation` (your reproducer's `COUNT(*)` shape). `LanceArrowStreamScannerTest` 4/4, `LanceFragmentColumnarBatchScannerTest` 1/1, `spotless:check` clean.

sezruby:
Addressed the full-text finding in 33c2788, and generalized the admission check so this class of divergence can't recur.

Rather than enumerate more column-name cases, admission now verifies the actual native scan schema against the declared partition schema. After planning the scan, `checkNativeSchemaMatchesPartition` compares `LanceScanner#schema()` (surfaced via a new `LanceFragmentScanner#schema()`) to `inputPartition.getSchema()` by field name and order, and rejects on any mismatch. A full-text query's auto-projected `_score` therefore falls out as `[id, body]` vs `[id, body, _score]` → reject. The same guard covers the synthesized `_fragid`, the `_rowaddr` added for blobs, the `_rowid` an empty projection surfaces, and reordered row-version columns, since the zero-copy export cannot re-project or reorder the way the columnar reader does after import.

Pushed aggregation stays a separate pre-plan check: an aggregate partition's declared schema can equal the native data schema, so the schema comparison alone cannot catch it.

`exportRejectsSchemasNeedingJvmPostProcessing` now exercises both mismatch directions — native returns fewer columns (`_fragid`) and native returns an extra column (empty projection surfacing `_rowid`, the same shape as the full-text `_score`). A `_rowid` column the native scan produces in matching order is now legitimately exportable, so that assertion was dropped. `LanceArrowStreamScannerTest` 4/4, `LanceFragmentColumnarBatchScannerTest` 1/1, `spotless:check` clean.

sezruby:
@yanghua could you take a look when you have a chance? This is the lance-spark side of the Gluten/Velox read path — it forwards a Lance fragment scan as an Arrow C stream via `LanceScanner#exportArrowStream(long)` (the API from #7259, now shipping in lance-core `11.0.0-beta.21`, which this PR bumps to).

CI is green and it's MERGEABLE; unit tests pass (`LanceArrowStreamScannerTest` 4/4, `LanceFragmentColumnarBatchScannerTest` 1/1). cc @jackye1995

yanghua:
CI is red?

sezruby:
The red CI isn't from this PR's changes; it's the `beta.10 → beta.21` lance-core bump exposing a pre-existing issue in the Kryo codec (`LanceSerializeUtil`), unrelated to the Arrow-stream forwarding.

`LanceSerializeUtil` configures Kryo with an Objenesis instantiator (`StdInstantiatorStrategy`), which constructs objects without calling their constructors. The JDK immutable collections (`List.of`/`copyOf`, `Set.of`, `Map.of`) have no no-arg constructor and keep their elements in a private `java.base`-internal field, so Kryo produces an instance whose backing array is never populated. beta.21's `CompactionOptions.excludedFragmentIds` is a `List.copyOf(...)`, and `OptimizeExec` ships each `CompactionTask` through this codec to the executors — so it fails to deserialize there:

```
java.lang.NullPointerException: Cannot read the array length because "this.elements" is null
```

That's why `OptimizeTest`, `VacuumTest`, and `ShowIndexesTest` fail across every Spark/Scala combination — it's version-independent.

I've split the fix into #783: it registers scoped Kryo `CollectionSerializer`/`MapSerializer` implementations for the immutable-collection base classes (ported from Kryo 5's `ImmutableCollectionsSerializers`) that read each element through Kryo and rebuild the collection via its public factory (`List`/`Set`/`Map.copyOf`), so they roundtrip correctly without routing decode through `ObjectInputStream`. It adds a regression test, and bumps lance-core to beta.21 so the same `OptimizeTest`/`VacuumTest`/`ShowIndexesTest` run against the version that reproduces the failure. #783 is from a fork, so its `Test` / `Integration Test` workflows are waiting on approval — could you trigger them (Approve and run)? Once #783 merges I'll rebase this PR onto it (the bump here becomes a no-op) and CI will be green.

yanghua:
> The red CI isn't from this PR's changes; it's the `beta.10 → beta.21` lance-core bump exposing a pre-existing issue in the Kryo codec (`LanceSerializeUtil`), unrelated to the Arrow-stream forwarding.
> 
> `LanceSerializeUtil` configures Kryo with an Objenesis instantiator (`StdInstantiatorStrategy`), which constructs objects without calling their constructors. The JDK immutable collections (`List.of`/`copyOf`, `Set.of`, `Map.of`) have no no-arg constructor and keep their elements in a private `java.base`-internal field, so Kryo produces an instance whose backing array is never populated. beta.21's `CompactionOptions.excludedFragmentIds` is a `List.copyOf(...)`, and `OptimizeExec` ships each `CompactionTask` through this codec to the executors — so it fails to deserialize there:
> 
> ```
> java.lang.NullPointerException: Cannot read the array length because "this.elements" is null
> ```
> 
> That's why `OptimizeTest`, `VacuumTest`, and `ShowIndexesTest` fail across every Spark/Scala combination — it's version-independent.
> 
> I've split the fix into #783: it registers Kryo's `JavaSerializer` as the default serializer for the immutable-collection base classes so they roundtrip correctly, adds a regression test, and bumps lance-core to beta.21 so the same `OptimizeTest`/`VacuumTest`/`ShowIndexesTest` run against the version that reproduces the failure. #783 is from a fork, so its `Test` / `Integration Test` workflows are waiting on approval — could you trigger them (Approve and run)? Once #783 merges I'll rebase this PR onto it (the bump here becomes a no-op) and CI will be green.

Got it. Will take a look.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * Rejects a partition whose declared Spark schema differs from the schema the native scan
+   * actually produces. The columnar reader reconciles such differences on the JVM after import —
+   * synthesizing {@code _fragid}, dropping the {@code _score} a full-text query auto-projects,
+   * stripping the {@code _rowaddr} added for blobs, surfacing {@code _rowid} for an empty
+   * projection, reordering row-version columns — but this zero-copy export cannot, so any mismatch
+   * in field names or order must fall back to the columnar reader.
+   */
+  private static void checkNativeSchemaMatchesPartition(
```
yanghua, line 133: Can we also check the arrow data type?

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceArrowStreamScannerTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+  /**
+   * Exports each fragment of the bundled test table as an Arrow C Data Interface stream, re-imports
+   * it on the JVM (standing in for a native consumer), and asserts the rows match what the Spark
+   * columnar reader produces. Closing the imported reader and then the {@link
+   * LanceArrowStreamScanner.LanceArrowStream} handle under the leak-checking allocator also
+   * verifies the export/reader/scanner lifecycle releases cleanly.
+   */
+  @Test
+  public void exportsFragmentAsArrowCStream() throws Exception {
```
yanghua, line 52: Can we also check a case about an empty fragment?

yanghua (commented):
Left two comments.

lance-spark-base_2.12/src/main/java/org/lance/spark/internal/LanceArrowStreamScanner.java
sezruby: Done in 235791d. `checkNativeSchemaMatchesPartition` now compares the declared and native schemas by Arrow `ArrowType` and nullability (recursing into children), not only by field name and order. The declared Spark schema is converted through `LanceArrowUtils` — the same adapter the read path uses — so the Arrow-specific distinctions a Spark `DataType` alone cannot express (`LargeUtf8` vs `Utf8`, `LargeBinary`, `Date(MILLISECOND)`, `FixedSizeBinary`, `Float16`) survive the comparison, and a column whose native type differs from the declared one (e.g. a narrower/wider int or a different time-zone timestamp) is no longer streamed as if it matched.

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceArrowStreamScannerTest.java
sezruby: Added `exportsEmptyFragmentScanPreservingSchema` in 235791d: a fragment scan whose filter matches no rows (`x < 0`, every `x` is 0..3) still exports its full `x, y, b, c` schema up front and drains cleanly with zero batches, releasing without a leak under the leak-checking allocator.

pom.xml
```diff
@@ -51,7 +51,7 @@
 
     <properties>
         <lance-spark.version>0.8.0-beta.1</lance-spark.version>
-        <lance.version>11.0.0-beta.10</lance.version>
+        <lance.version>11.0.0-beta.21</lance.version>
```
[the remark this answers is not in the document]
sezruby: Agreed — this is the JDK immutable-collection Kryo bug fixed by #783 (`CompactionOptions.excludedFragmentIds` = `List.copyOf(...)`, which the unchanged `LanceSerializeUtil` codec reconstructs without its backing elements). #783 is gate-approved and only waiting on a maintainer to Approve-and-run its gated fork-PR CI. Once it merges I will rebase this PR, at which point the version bump here becomes a no-op and `OptimizeTest`/`VacuumTest`/`ShowIndexesTest` go green.

yanghua:
@sezruby, do you have WeChat? We can contact each other more effectively.

sezruby:
> @sezruby, do you have WeChat? We can contact each other more effectively.

I don't have WeChat, but feel free to reach me on the ASF Slack (EJ Song) or LinkedIn. And I can join WeChat if there's a group chat for Lance.

yanghua:
And I can join WeChat if there's a group chat for Lance.

> yes, we have multiple WeChat groups for Chinese. But you do not have a WeChat account; how to join?

sezruby:
I created one. @yanghua

sezruby:
@yanghua Rebased onto `main` now that #783 has merged, so the earlier beta.21/Kryo blocker is resolved — the lance-core bump is already on `main` and dropped out of this PR as a no-op, leaving just the three Gluten files.

The `Spark` test workflow passed on the prior head across Spark 3.4–4.2 / Scala 2.12–2.13; I've re-pushed to refresh the checks. The `lance-gatekeeper` check went to `action_required` after the rebase and should re-evaluate on this push. Could you approve the pending workflow run when you get a chance? Thanks!

yanghua (approved):
+1

PR #781, opened 2026-08-24
fix(read): keep fragments the zonemap does not cover

### LINKED ISSUE #780: Zonemap pruning silently drops fragments the index does not cover
## Problem

`ZonemapFragmentPruner` builds its surviving-fragment set **solely** from fragment IDs that appear in the returned `ZoneStats`. A fragment with no zone entry is therefore treated as "provably non-matching" and pruned away.

But a zonemap index can legitimately cover only part of a dataset — fragments appended after the last index build carry no zone entries. Their contents are **unknown**, not provably non-matching. Dropping them makes a filtered scan return a subset of the matching rows, with **no exception and no warning**.

This contradicts the class's own documented contract:

> Zonemap pruning is inexact (conservative): it may include fragments that ultimately contain no matching rows, but **it will never exclude fragments that do contain matching rows**.

Partial coverage is the steady state of a date-partitioned ingestion pipeline: the index is rebuilt periodically while new data is appended continuously. Every filtered read between rebuilds is silently wrong.

## Reproduction

A table with a `STRING` date column, indexed after the first three dates are written, then two more appended:

```sql
CREATE TABLE t (id INT, ingestion_date STRING) USING lance;
-- INSERT 2026-02-14, 2026-02-15, 2026-02-16   (one fragment each)
ALTER TABLE t CREATE INDEX ingestion_date_idx USING zonemap (ingestion_date);
-- INSERT 2026-02-17, 2026-02-18               (not covered by the index)

SELECT count(*) FROM t WHERE ingestion_date = '2026-02-17';
```

| query | expected | actual |
|---|---|---|
| `ingestion_date = '2026-02-17'` (unindexed) | 4 | **0** |
| `ingestion_date = '2026-02-18'` (unindexed) | 4 | **0** |
| sum of per-date counts vs. full scan | 20 | **12** |
| `>= '2026-02-16' AND <= '2026-02-18'` | 12 | **4** |

Indexed dates return correct counts throughout, which is what makes this hard to notice — the table looks healthy until you reconcile per-date counts against a full scan.

Observed in production as a date filter returning 0 rows against a table holding ~470M matching rows for that date.

## Root cause

Three places conspire:

1. **`ZonemapFragmentPruner.pruneFragments`** — the surviving set is the union of `zone.getFragmentId()` over the returned stats. Fragments absent from the stats are never added, so they are excluded.
2. **`LanceScanBuilder.findZonemapIndexedColumns`** — swallows any failure into an **empty set**, which `loadZonemapStats` then reads as *"no information, try every column"* (`if (zonemapColumns.isEmpty() || zonemapColumns.contains(col))`). A failure to read index metadata is indistinguishable from "this dataset has no zonemap index".
3. **`LanceScan.pruneByZonemapStats`** — when the builder's cached result is null, it recomputes pruning itself, without the index metadata that would tell it what is covered.

The underlying defect is that this path has **no safe degradation mode**: any failure to establish coverage, from any cause, silently becomes wrong answers rather than a full scan.

## Two known triggers

**(a) A partially-covering index** — the case above.

**(b) #514** — distributed scalar-index builds racing on a shared `zonemap.lance`, where "whichever task commits last wins, and the read path (`getZonemapStats`) returns only one fragment's worth of zone stats." That is a *different cause* producing partial stats, which then hits this same downstream defect. Fixing #514 removes one trigger; fixing this removes the class of failure.

A third trigger we hit: `describe_indices` throwing on a dataset carrying a MemWAL system index (`No scalar index plugin found for name 'memwal'`) makes `findZonemapIndexedColumns` return its empty set, with the same result.

## Proposed fix

Make pruning conservative when coverage is incomplete or unknown:

- Thread the set of fragments the zonemap actually covers alongside the stats, derived from each index segment's fragment bitmap (already reachable in Java via `IndexDescription.getMetadata()` → `Index.fragments()`, which returns `Optional<List<Integer>>` so a missing bitmap stays distinguishable from an empty one). Fragments outside that set are retained unconditionally.
- Have `findZonemapIndexedColumns` report unreadable index metadata distinctly from "no indices", and disable pruning in that case rather than guessing.
- Keep loading zone stats even when coverage is unknown — sharding detection consumes the same stats and only needs per-fragment min/max. **Only pruning** is disabled. (Disabling stats loading outright regresses storage-partitioned joins; `BucketSpjTest.testBucketJoinPlanShowsSpj` catches this.)
- Remove `LanceScan`'s coverage-blind recomputation.

Net effect: unknown coverage degrades to a correct full scan instead of a silently wrong result. Fully-covered indexes prune exactly as before.

## Not duplicates

- **#514** — a cause of partial stats; this is the downstream handling of them.
- **#557** (*exception throws when zonemap pruner on int column*) — same component, but an exception rather than a silent wrong answer.

## Status

I have a fix ready with tests and would like to claim this issue.

- 5 unit cases in `ZonemapFragmentPrunerTest` covering partial coverage, multi-column intersection, unknown coverage, and a no-degradation guard.
- A `BasePartialZonemapCoverageTest` integration suite driving real Spark SQL against a Lance table whose index covers 3 of 5 dates.
- Verified red/green: reverting the fix reproduces the table above exactly. Full suite (1,944 tests across `lance-spark-base_2.12` + `lance-spark-3.5_2.12`) passes.
- Also verified end-to-end with a built bundle jar against a real Spark job, comparing old vs. new jar on the same dataset.

kangnan-li:
Closes #780

## Problem

`ZonemapFragmentPruner` builds its surviving-fragment set solely from fragment IDs present in the returned `ZoneStats`, so a fragment with **no zone entry** is treated as provably non-matching and pruned away.

A zonemap index can legitimately cover only part of a dataset — fragments appended after the last index build carry no zone entries. Their contents are *unknown*, not provably non-matching. Dropping them makes a filtered scan return a subset of the matching rows with no exception and no warning, contradicting the class's own documented contract:

> it will never exclude fragments that do contain matching rows

Partial coverage is the steady state of a date-partitioned ingestion pipeline: the index is rebuilt periodically while new data is appended continuously. Every filtered read between rebuilds is silently wrong. See #780 for the full analysis and the two other known triggers (#514's racing distributed builds, and `describe_indices` throwing on a MemWAL system index).

## Changes

**`ZonemapFragmentPruner`** — new overload taking `Map<column, uncoveredFragments>`: for each pushed predicate, the fragment set it yields is widened by the fragments *its own columns'* indices do not describe, **before** the per-predicate sets are intersected.

Coverage has to be per column, because indexes are built independently and cover different fragments. A single dataset-wide uncovered set is not sufficient: with an index on `a` covering `{0}` and one on `b` covering `{0,1}`, the union leaves fragment 1 looking covered, so a predicate on `a` still eliminates it even though `a`'s index never described it — the original bug, one column over. Widening per predicate before the intersection is what closes that.

A `null` map means coverage is unknown and disables pruning entirely. The 2-arg overload is retained for source compatibility but deprecated: it asserts full coverage, which it cannot verify.

**`LanceScanBuilder`** — `loadZonemapStats` now returns the stats together with each column's uncovered-fragment set, rather than setting a field as a side effect, so the two cannot fall out of step. Coverage is derived from the loaded `ZoneStats` themselves: a column may carry several zonemap indexes and `getZonemapStats` returns the zones of only one of them, so index metadata would report the union of all of them and claim coverage for fragments the loaded zones never saw. A fragment absent from the zones actually being evaluated is unknown, so it stays in the scan. This also drops the `describeIndices` call from the read path entirely.

Zone stats are loaded for **every** requested column regardless of what index discovery reports. Sharding detection consumes the same stats and needs only per-fragment min/max, so storage-partitioned joins keep working on tables whose index metadata cannot be read — e.g. one carrying a MemWAL system index, where `describe_indices` fails with `No scalar index plugin found for name 'memwal'`. Only pruning is gated on coverage.

**`LanceScan`** — removed the fallback that recomputed pruning without coverage information; it would reintroduce the bug precisely when the builder had deliberately declined to prune. The `zonemapStats` field it was the sole reader of is now gone.

Net effect: unknown or partial coverage degrades to a correct full scan instead of a silently wrong result. Fully-covered indexes prune exactly as before.

## Tests

**Unit** — `ZonemapFragmentPrunerTest` 37 → 43 cases: uncovered fragment survives when another fragment matches; survives when *no* covered fragment matches (the production shape); survives across a multi-column intersection; **survives when covered by one column's index but not another's**; unknown coverage disables pruning; and a guard that fully-covered stats still prune exactly, so the fix cannot degrade into "never prune".

**Integration** — `BasePartialZonemapCoverageTest` drives real Spark SQL against a Lance table with a `STRING` date column whose index covers 3 of 5 dates (the remaining 2 appended afterwards): filters on indexed and unindexed dates, per-date counts reconciling with a full scan, a range spanning both, and a value absent everywhere still returning 0 rows. Two further cases cover multiple indexes: one table with indexes on two different columns covering different fragments, and one with two indexes on the *same* column (the case from review).

Reverting just the fix reproduces the failure end-to-end:

| case | pre-fix | post-fix |
|---|---|---|
| filter on unindexed date | **0** rows (expected 4) | 4 |
| per-date counts vs. full scan | **12** (expected 20) | 20 |
| range spanning indexed + unindexed | **4** (expected 12) | 12 |
| date absent everywhere | 0 | 0 |

Full suite passes: 1,944 tests across `lance-spark-base_2.12` and `lance-spark-3.5_2.12`. Spotless and checkstyle clean.

Additionally verified outside the test harness by building the `4.1_2.13` bundle and running the same dataset through a real Spark job with the old and new jar — old jar reproduces the failure, new jar returns correct counts while still pruning.

## Cost of the fix

Pruning is disabled when coverage cannot be established, so a query that previously pruned (incorrectly) now scans. Measured on dev tables with one distinct date per fragment, probing a date that lives in an uncovered fragment:

| fragments | index coverage | probe | best of 3 | vs. fully-covered |
|---|---|---|---|---|
| 300 | 100% | covered | 0.117s | 1.0x |
| 300 | 50% | uncovered | 0.808s | 6.9x |
| 300 | 10% | uncovered | 1.330s | 11.4x |
| 300 | 0% | uncovered | 1.054s | 9.0x |
| 5000 | 100% | covered | 0.922s | 1.0x |
| 5000 | 50% | uncovered | 30.4s | 33x |
| 5000 | 10% | uncovered | 55.6s | **60x** |

All rows return **correct** counts; before the fix every uncovered probe returned 0 rows instantly.

**The penalty grows with fragment count, not just with the coverage ratio** — at 50% coverage it is 6.9x on 300 fragments but 33x on 5000. (0% coverage is cheaper than 10%: with no stats at all the pruning path is skipped entirely and Spark plans a plain scan, whereas partial stats cost planning work that saves little.) One Spark task is scheduled per surviving fragment, so per-task overhead dominates once thousands of fragments are unpruned. On a table with hundreds of thousands of fragments, a badly stale index will push a filtered read towards full-scan cost. Keeping the index current matters.

That is the intended trade: the fragments the index does not describe genuinely have to be read. A fully-covered index is unaffected — `uncoveredFragmentIds` is empty and the added `addAll` is a no-op (`testFullyCoveredStatsStillPruneExactly`).

`LanceScanBuilder` logs covered-vs-total fragment counts whenever any are uncovered, so a degrading index is visible rather than silent.

## Test matrix

`lance-spark-4.1_2.13` includes `lance-spark-3.5_2.12/src/test/java` in its test sources, so the new suite runs on both Spark lines. Verified on Spark 4.1 / Scala 2.13: `PartialZonemapCoverageTest` (6), `ZonemapIndexTest` (3), `BucketSpjTest` (2) all pass.


lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -534,19 +602,41 @@ private Set<String> findZonemapIndexedColumns(Dataset dataset) {
...
+          LOG.warn(
+              "Zonemap index '{}' reports no fragment bitmap; its columns will not be used"
+                  + " for fragment pruning",
+              idx.getName());
+          continue;
+        }
+        for (int fieldId : idx.getFieldIds()) {
+          String name = fieldIdToName.get(fieldId);
+          if (name != null) {
+            columns.computeIfAbsent(name, unused -> new HashSet<>()).addAll(coverage);
```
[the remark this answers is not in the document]
geruh: I think the problem is that `getZonemapStats(column)` and this bitmap union aren't talking about the same zonemap.

For example:

```text
a_partial covers fragment 0
append fragment 1
z_full covers fragments 0 and 1

getZonemapStats("x") -> stats from a_partial -> {0}
bitmap union -> {0, 1}
```

So we can think fragment 1 is covered even though the stats we're actually pruning with never saw it.

The gatekeeper comment here is valid then. Querying a value that only exists in fragment 1 returns 0 rows. The current coverage tests still pass because they're all one index per column.

I don't think `describe_indices` should be what "decides" coverage or rows matching or not. If a fragment isn't in the `ZoneStats` we're actually evaluating, we don't know anything about it, so it has to stay in the scan.

Could we make it like 

```
dataset fragments - fragment ids present in the loaded ZoneStats = uncovered
```

and add the two index case as an another test?

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -534,19 +602,41 @@ private Set<String> findZonemapIndexedColumns(Dataset dataset) {
...
           }
         }
       }
     } catch (Exception e) {
       LOG.warn("Failed to query zonemap indexes: {}", e.getMessage());
+      return java.util.Optional.empty();
     }
-    return columns;
+    return java.util.Optional.of(columns);
```
geruh: nit: do we need fully qualified import?

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BasePartialZonemapCoverageTest.java
```diff
@@ -0,0 +1,239 @@
...
+import org.junit.jupiter.api.io.TempDir;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.util.UUID;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+/**
+ * End-to-end reads against a zonemap index that does not cover the whole table.
```
geruh: nit: this class doc pretty much retells the issue #780 saw. We can let the code and git history tell the story, also same nit for the comments on the test code as well

lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
```diff
@@ -544,4 +544,141 @@ public void testInListWithMixedWidthLiteralsAgainstLongZoneStats() {
     assertTrue(result.isPresent());
     assertEquals(Set.of(0, 2), result.get());
   }
+
+  // ---------------------------------------------------------------------------
+  // Partial zonemap coverage.
```
geruh: nit: Also here, this banner and the code mention the old behavior and can be dropped let's be current same goes for the test comments

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
```diff
@@ -57,15 +57,50 @@ public final class ZonemapFragmentPruner {
...
+   * @param pushedPredicates the V2 predicates pushed down by Spark
+   * @param zonemapStatsByColumn map from column name to its zonemap zone stats
+   * @param uncoveredFragmentsByColumn per column, the fragments present in the dataset but absent
+   *     from that column's zonemap index; always retained for predicates on that column. A column
+   *     mapping to an empty set (or absent from the map) is treated as fully covered. Pass {@code
+   *     null} when coverage cannot be determined — that disables pruning entirely.
+   * @return present with the set of fragment IDs that might match; empty if no pruning can be
+   *     derived
+   */
+  public static Optional<Set<Integer>> pruneFragments(
```
geruh, line 78: nit: this is also pretty verbose.

geruh (commented):
Hey @kangnan-li thanks for raising!! I did a pass over the logic let me know what you think.

geruh (changes requested):

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScan.java
```diff
@@ -365,20 +356,15 @@ private List<LanceSplit> pruneByLimit(
...
-    if (cachedSurvivingFragmentIds != null) {
-      allowedIds = cachedSurvivingFragmentIds;
-    } else if (!zonemapStats.isEmpty()) {
-      allowedIds =
-          ZonemapFragmentPruner.pruneFragments(pushedPredicates, zonemapStats).orElse(null);
-    } else {
+    // Only the builder can compute this safely: it holds the index metadata that says
+    // which fragments the zonemap actually covers. A null cache means the builder
+    // determined pruning was unsafe (e.g. unknown coverage), so recomputing here —
+    // without that metadata — would silently drop fragments it deliberately kept.
```
geruh: nit: if i'm following correctly coverage comes from the loaded ZoneStats now. The cache is just "builder already pruned." A null cache means do not prune here, not unknown index coverage.

maybe? `// Null means the builder did not prune. Do not recompute here.`

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
kangnan-li: Done in `de40f16` — the fully-qualified `java.util.Optional` went away with `findZonemapIndexedColumns`, which the coverage rework removed entirely.

One remains at `LanceScanBuilder.java:215` (`java.util.Optional<Map<Integer, Object>> keys`), but that one predates this PR: the file imports `org.lance.spark.utils.Optional`, so the JDK type has to be qualified there. Happy to leave it or clean it up separately, whichever you prefer.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BasePartialZonemapCoverageTest.java
kangnan-li: Done in `de40f16` — class doc is now a single line, and the narrative comments inside the test bodies are gone. Kept only short notes where a test's setup isn't self-evident from the SQL.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
kangnan-li: Done in `de40f16` — banner removed, and the comments describing the old behaviour are gone. The remaining ones state what the case pins, not what it used to do.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
kangnan-li: Trimmed in `de40f16` — cut roughly in half, down to the invariant and the per-column scoping, with the param docs shortened.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BasePartialZonemapCoverageTest.java
```diff
@@ -0,0 +1,230 @@
...
+                String.format(
+                    "SELECT * FROM %s WHERE region = 'west' AND ingestion_date = '2026-02-15'",
+                    table))
+            .count(),
+        "fragment 1 is outside region_idx, so the region predicate cannot rule it out");
+  }
+
+  @Test
+  public void testTwoZonemapIndexesOnTheSameColumn() {
+    // getZonemapStats returns the zones of only one of a column's indexes. Coverage must come
```
geruh: nit: drop comment

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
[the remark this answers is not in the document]
kangnan-li: Reproduced it verbatim before changing anything — `SELECT * FROM t WHERE x = 100` returned 0. Your framing is better than what I had: index metadata shouldn't decide coverage at all.

Switched to exactly what you suggested — `dataset fragments − fragment IDs present in the loaded ZoneStats`. That drops the `describeIndices` call from the read path entirely, so there's no second source of truth left to disagree with the stats. It also removes the `memwal` `describe_indices` failure as a side effect.

Added your case as `testTwoZonemapIndexesOnTheSameColumn` (two indexes, same column, built at different points), alongside the existing two-column-different-coverage case. Fails before the change, passes after.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -490,66 +486,87 @@ private static Optional<Long> getCountFromMetadata(Dataset dataset) {
...
+        LOG.debug("Failed to load zonemap stats for column '{}': {}", col, e.getMessage());
       }
     }
+    return result;
+  }
 
-    if (!result.isEmpty()) {
-      LOG.debug("Loaded zonemap stats for {} columns: {}", result.size(), result.keySet());
-    }
+  /** Zone stats for the requested columns, plus which fragments each column's index misses. */
```
geruh: This is still talking about the index, and `uncoveredByColumn` still documents null. I believe this shouldn't return null anymore.


maybe just: 
```suggestion

  private static final class ZonemapLoadResult {
    final Map<String, List<ZoneStats>> stats;
    final Map<String, Set<Integer>> uncoveredByColumn;

    ZonemapLoadResult(Map<String, List<ZoneStats>> stats, Map<String, Set<Integer>> uncovered) {
      this.stats = stats;
      this.uncoveredByColumn = uncovered;
    }
  }
```

lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
```diff
@@ -544,4 +544,124 @@ public void testInListWithMixedWidthLiteralsAgainstLongZoneStats() {
...
+    Map<String, List<ZoneStats>> stats = new HashMap<>();
+    stats.put(
+        column,
+        Arrays.asList(
+            new ZoneStats(0, 0, 100, 0L, 99L, 0), new ZoneStats(1, 0, 100, 100L, 199L, 0)));
+    return stats;
+  }
+
+  @Test
+  public void testUncoveredFragmentSurvivesWhenPredicateMatchesElsewhere() {
```
kangnan-li, line 559: what does this mean? reply comments with an example in the coment, not in the code

kangnan-li: It describes the fixture, which models an index that has fallen behind the table.

`ZoneStats` is `(fragmentId, zoneStart, zoneLength, min, max, nullCount)`, so the helper builds:

```
fragment 0  →  ZoneStats(0, 0, 100,   0,  99, 0)
fragment 1  →  ZoneStats(1, 0, 100, 100, 199, 0)
fragment 2  →  (no entry at all)
```

The dataset has three fragments, but the zones describe only two. That is what a table looks like after:

```sql
INSERT ...                       -- fragment 0
INSERT ...                       -- fragment 1
CREATE INDEX ... USING zonemap (x);   -- indexes fragments 0 and 1
INSERT ...                       -- fragment 2, no zones for it
```

The distinction the tests turn on is between the two reasons a fragment can be missing from the pruner's result:

- fragment 1 with `x = 250` → **ruled out**; its zone says `[100, 199]`, so 250 cannot be there.
- fragment 2 with `x = 250` → **unknown**; nothing describes it, so it may well contain 250.

Only the first justifies dropping it. The old code treated both the same, which is the bug.

Happy to rename the helper to something like `statsMissingFragmentTwo` if the intent is still not obvious from the call sites.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
geruh: Javadocs seems a bit closer! 

null is still being documented but we don't have that behavior. 

Maybe something like:

```
* @param uncoveredFragmentsByColumn per column, the fragments its loaded zones do not describe 
*     and always retained for predicates on that column. An absent or empty entry means fully
*     described.
```

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
```diff
@@ -74,16 +103,35 @@ public static Optional<Set<Integer>> pruneFragments(
...
       Optional<Set<Integer>> fragmentIds = analyzePredicate(predicate, zonemapStatsByColumn);
-      if (fragmentIds.isPresent()) {
-        if (result == null) {
-          result = new HashSet<>(fragmentIds.get());
-        } else {
-          result.retainAll(fragmentIds.get());
+      if (!fragmentIds.isPresent()) {
+        continue;
+      }
+      // Widen this predicate's set with the fragments its own columns' indices do not
```
geruh: nit: can drop this comment or trim, the code is pretty good to follow

lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
```diff
@@ -544,4 +544,124 @@ public void testInListWithMixedWidthLiteralsAgainstLongZoneStats() {
...
+  }
+
+  @Test
+  public void testUnknownFragmentSetDisablesPruningEntirely() {
+    // When the dataset's fragment set cannot be determined, no fragment can be
+    // proven absent, so pruning must be abandoned rather than guessed.
+    Map<String, List<ZoneStats>> stats = partiallyCoveredStats("x");
+    Predicate[] filters = new Predicate[] {TestPredicates.eq("x", 250L)};
+
+    Optional<Set<Integer>> result = ZonemapFragmentPruner.pruneFragments(filters, stats, null);
```
geruh: is this test needed we don't ever pass in null right?

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -490,66 +486,87 @@ private static Optional<Long> getCountFromMetadata(Dataset dataset) {
...
 
-  private Set<String> findZonemapIndexedColumns(Dataset dataset) {
-    Set<String> columns = new HashSet<>();
-    try {
-      Map<Integer, String> fieldIdToName = new HashMap<>();
-      for (LanceField field : dataset.getLanceSchema().fields()) {
-        fieldIdToName.put(field.getId(), field.getName());
-      }
+  /**
+   * Loads zonemap statistics for the requested columns, together with the fragments each column's
```
geruh: Nit: meant to post this one before but this is a description of the fix we can trim this as well can very well say `/* Zone stats plus, per column, dataset fragments those stats do not describe. */` at the very least

geruh (commented):
Nice work @kangnan-li! This shape is a lot better I left a few nits for readability and correctness in the comments let me know what you think

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
kangnan-li: You're right, and it's more than a doc fix — I checked, and `uncoveredByColumn` genuinely cannot be null any more. All three construction sites pass a non-null map; the null path died with `findZonemapIndexedColumns`.

So in `8942e62` I took your suggestion for the class verbatim, and also removed the now-dead `if (uncoveredFragmentsByColumn == null)` branch in `pruneFragments` and the unit test that exercised it. Dead code documenting behaviour that no longer exists is worse than no comment.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
kangnan-li: Fixed in `8942e62` — used your wording. The null branch it described is gone too (see the `ZonemapLoadResult` thread).

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScan.java
kangnan-li: Agreed — your wording is accurate now that coverage comes from the stats. Applied verbatim in `8942e62`.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/BasePartialZonemapCoverageTest.java
kangnan-li: Dropped in `8942e62`.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
kangnan-li: Trimmed in `8942e62` to two lines — kept just the reason the widening happens before the intersection, since that ordering is the part that isn't obvious from the code.

lance-spark-base_2.12/src/test/java/org/lance/spark/read/ZonemapFragmentPrunerTest.java
kangnan-li: No — good catch. `null` is unreachable now, so the test was pinning behaviour that no longer exists. Removed in `8942e62`, along with the branch it covered.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
kangnan-li: Used your line verbatim in `8942e62`.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/ZonemapFragmentPruner.java
kangnan-li: Right — it was still 16 lines against 3–9 for the other javadocs in this package. Trimmed further in `0dfcd66`:

- 3-arg `pruneFragments`: 16 → 7 lines. Dropped the `@param` entries that only restated the signature and kept the one that carries real information (what an absent/empty entry means).
- 2-arg overload: 13 → 6, since `@deprecated` already says everything the prose did.
- `loadZonemapStats`: 10 → 8, cut the sharding sentence (the code says it).

That is the last of your comments on this pass; the rest went out in `8942e62`.

kangnan-li:
Make a few changes based on comments, and resolved conversations for minor changes, and keep major discussions unresolved for reviewer to review again

geruh (approved):
LGTM thanks @kangnan-li!

PR #782, opened 2026-08-25
test: cover managed backfill commits on rest-dir

hfutatzhanghb:
## Summary

- Add an opt-in managed-versioning mode to the local REST directory namespace server.
- Forward the mode through `make docker-test` and the pytest REST fixture.
- Run the ADD COLUMNS and UPDATE COLUMNS backfill tests in a separate managed rest-dir Docker invocation so existing REST search tests keep their current server behavior.

## Context

Follow-up to #772 and https://github.com/lance-format/lance-spark/pull/772#issuecomment-5407815817.

The existing backfill tests were REST-compatible, but no CI job selected them on rest-dir and the default directory adapter reported `managedVersioning=false`. The dedicated invocation enables both `manifest_enabled` and `table_version_tracking_enabled`, matching the managed-versioning setup covered by lance-core 11.0.0-beta.10.

## Validation

- `./mvnw spotless:check -pl lance-spark-3.5_2.13 -am -DskipTests`
- Standalone `javac` compilation of `LanceRestDirNamespaceServer.java`
- Python AST and workflow YAML parsing
- Pytest collection confirms exactly two managed backfill tests on `rest-dir`
- `make -n docker-test` confirms the managed-versioning flag reaches the container

The local host has no Docker CLI and its x86_64 JVM cannot load the available macOS JNI artifact, so the managed REST round-trip is delegated to the new Ubuntu Docker CI step.


integration-tests/test_lance_spark.py
```diff
@@ -2465,10 +2482,17 @@ def test_add_column_from_view_on_rest(self, spark, test_table):
             FROM {test_table}
         """)
 
+        commit_count_before = (
+            create_table_version_count() if create_table_version_count else None
```
geruh: small nit: I'd require `create_table_version_count` here instead of skipping when it's None. Does it let you?

integration-tests/test_lance_spark.py
```diff
@@ -2480,13 +2504,21 @@ def test_add_column_from_view_on_rest(self, spark, test_table):
             (2, "bravo", "bravo"),
         ]
 
+        if commit_count_before is not None:
+            # Negative control: a real direct commit must bypass the namespace counter.
```
geruh: nit: this comment can be dropped the helper name already says it

integration-tests/test_lance_spark.py
```diff
@@ -67,25 +67,40 @@ def _read_options_builder(jvm):
...
+        spark._jvm.org.lance.Dataset.open()
+        .allocator(spark._jvm.org.lance.spark.LanceRuntime.allocator())
         .uri(_table_location(spark, table_name))
         .readOptions(read_options)
         .build()
     )
+
+
+def _commit_dataset_config_directly(spark, table_name):
+    """Make a direct Dataset commit that intentionally bypasses the namespace client."""
```
geruh: nit: same here, docstring can drop

PR #783, opened 2026-08-26
fix: support JDK immutable collections in the Kryo codec

sezruby:
## Problem

`LanceSerializeUtil` (used to ship compaction and index tasks to executors) configures Kryo with an Objenesis-based instantiator (`StdInstantiatorStrategy`), which constructs objects without calling their constructors. The JDK immutable collections — `List.of`/`List.copyOf`, `Set.of`, `Map.of` — have no no-arg constructor and keep their elements in a private, `java.base`-internal field. Kryo therefore produces an instance whose backing storage is never populated, and decoding such a collection throws on first use:

- `KryoException: java.lang.UnsupportedOperationException`
- `NullPointerException: Cannot read the array length because "this.elements" is null`

lance-core value objects passed through this codec carry such collections. Starting in `11.0.0-beta.21`, `CompactionOptions` stores `excludedFragmentIds` as `List.copyOf(...)`, so a compaction task fails to deserialize on the executor:

```
com.esotericsoftware.kryo.KryoException: java.lang.UnsupportedOperationException
Serialization trace:
excludedFragmentIds (org.lance.compaction.CompactionOptions)
```

This is what makes `OptimizeTest` / `VacuumTest` / `ShowIndexesTest` fail once lance-core is bumped past beta.10.

## Changes

1. **Bump lance-core `11.0.0-beta.10` → `11.0.0-beta.21`.** This is the version that introduces the `excludedFragmentIds` field, so it reproduces the failure — the PR's own CI now exercises the affected tests against the version that breaks.

2. **Fix the Kryo codec** with dedicated serializers for the JDK immutable collections (`ImmutableCollectionsSerializers`), ported from Kryo 5's `ImmutableCollectionsSerializers` — auto-wired by default there, but never backported to the Kryo 4.0.2 that Spark bundles. They are registered on the package-private abstract bases reached via `getSuperclass`, which covers every size variant (`List12`/`ListN`, `Set12`/`SetN`, `Map1`/`MapN`) without naming a `java.base`-internal class in source and without `--add-opens`. Each element is read onto a mutable staging buffer through Kryo and the collection is then rebuilt through its public factory (`List`/`Set`/`Map.copyOf`).

   - **No `JavaSerializer`.** `decode()` is reachable from caller-controlled write options, and `JavaSerializer` would route every nested element through `ObjectInputStream.readObject` — a Java deserialization gadget surface right at that entry point (a crafted payload's `readObject` would run before the decoded value is even cast). Keeping every element on the Kryo path avoids that. As a bonus, immutable sub-lists (`List.of(...).subList(...)`) are handled too, since the collection is rebuilt by iteration rather than reconstructed in place.

   - **Reference identity.** Rebuilding means the value returned to callers (the immutable copy) is not the object Kryo registered while reading (the mutable staging buffer). Left alone — as in Kryo 5's own port — a back-reference to an aliased collection resolves to that throwaway buffer, so a decoded alias could be a mutable collection (or the wrong type for a set/map). `ImmutableAwareReferenceResolver` rewrites each collection's Kryo reference id to its immutable copy once rebuilt, so aliased collections round-trip to the same immutable instance. This is stricter than Kryo 5, whose port does not fix aliasing.

   - **Cyclic graphs.** A pure-immutable cycle is impossible (a collection cannot contain itself). A cycle through a mutable object that merely holds an immutable collection is supported. The one unsupportable shape — a back-reference re-entering an immutable collection while it is still being read — is detected and rejected with a clear `KryoException` rather than silently leaking the mutable buffer.

   - **Hardened read buffer.** The list and set serializers stage elements in an `ArrayDeque`, not an `ArrayList`: `CollectionSerializer.read` calls `ArrayList.ensureCapacity` with the wire-supplied length before reading any element, so on this caller-reachable path a malformed huge length would eagerly allocate a giant backing array. `ArrayDeque` avoids that branch and preserves insertion order for the `List.copyOf` rebuild.

## Test

`LanceSerializeUtilTest` (JUnit 5, 7 cases):

- Round-trips `List` / `Set` / `Map` across the sizes that select the distinct backing classes, plus a `Serializable` object holding an immutable-list field — each throws on the pre-fix codec and passes with the fix.
- `doesNotInvokeReadObjectOfNestedElements` locks in the security property: a nested element's `readObject` must not run during a Kryo decode.
- `preservesReferenceIdentityOfAliasedImmutableCollections` asserts aliased list/set/map decode to a single immutable instance (and stay immutable).
- `supportsCycleThroughAMutableObjectHoldingAnImmutableCollection` and `rejectsCycleThatReentersAnImmutableCollectionStillBeingRead` cover the two cyclic-graph outcomes.

The pre-existing `OptimizeTest` / `VacuumTest` / `ShowIndexesTest` provide end-to-end coverage against beta.21.

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
+++ b/lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
@@ -14,14 +14,19 @@
 package org.apache.spark.sql.util;
 
 import com.esotericsoftware.kryo.Kryo;
+import com.esotericsoftware.kryo.KryoException;
 import com.esotericsoftware.kryo.io.Input;
 import com.esotericsoftware.kryo.serializers.CollectionSerializer;
 import com.esotericsoftware.kryo.serializers.MapSerializer;
+import com.esotericsoftware.kryo.util.MapReferenceResolver;
 
+import java.util.ArrayDeque;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Deque;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -41,16 +46,35 @@ import java.util.Set;
  * deserialization gadget surface that is dangerous here because {@code LanceSerializeUtil.decode}
  * is reachable from caller-controlled write options (a crafted payload's {@code readObject} would
  * run before the decoded value is even cast). Instead they keep every element on the Kryo path and
- * rebuild the collection through its public factory ({@code List.copyOf}/{@code Set.copyOf}/ {@code
+ * rebuild the collection through its public factory ({@code List.copyOf}/{@code Set.copyOf}/{@code
  * Map.copyOf}), preserving immutability. No {@code java.base}-internal reflection is used, so no
  * {@code --add-opens} is required, and immutable sub-lists are handled too because the collection
  * is rebuilt by iteration rather than reconstructed in place.
+ *
+ * <p><b>Reference identity.</b> Rebuilding a collection means the value returned to callers (the
+ * immutable copy) is not the object Kryo registered while reading (a mutable staging buffer). Left
+ * alone -- as in Kryo 5's own port -- a back-reference to an aliased collection would resolve to
+ * that throwaway buffer, so a decoded alias could be a mutable {@code ArrayList} (or the wrong type
+ * for a set/map). {@link ImmutableAwareReferenceResolver} closes that gap: each serializer arms the
+ * resolver immediately before delegating to {@code super.read}, which lets the resolver capture the
+ * id Kryo assigns to the collection and, once the immutable copy exists, swap it in so every later
+ * back-reference resolves to the immutable value. Aliased collections therefore round-trip to the
+ * same immutable instance.
+ *
+ * <p><b>Cyclic graphs.</b> A pure-immutable cycle is impossible (an immutable collection cannot
+ * contain itself -- it does not exist yet when its elements are chosen). A cycle can only form
+ * through a mutable object in the loop. When the immutable collection is <em>not</em> the object
+ * the cycle re-enters, it is fully supported (the mutable member is registered before its fields
+ * are read, so the back-reference resolves correctly). The one unsupportable shape is an immutable
+ * collection that a back-reference re-enters <em>while it is still being read</em> -- its immutable
+ * copy does not exist yet, so only the mutable buffer could be handed out. That case is detected
+ * and rejected with a clear {@link KryoException} rather than silently leaking the buffer.
  */
 public final class ImmutableCollectionsSerializers {
 
   private ImmutableCollectionsSerializers() {}
 
-  /** Registers the immutable-collection serializers on {@code kryo}. */
+  /** Registers the immutable-collection serializers and the identity-preserving resolver. */
   public static void register(Kryo kryo) {
     // Reach the package-private abstract bases via getSuperclass so no java.base-internal class is
     // named in source. List/Set/Map are distinct hierarchies, each rebuilt through its own factory;
@@ -59,14 +83,21 @@ public final class ImmutableCollectionsSerializers {
     Class<?> immutableList = List.copyOf(Collections.emptyList()).getClass().getSuperclass();
     Class<?> immutableSet = Set.copyOf(Collections.emptySet()).getClass().getSuperclass();
     Class<?> immutableMap = Map.of().getClass().getSuperclass();
-    kryo.addDefaultSerializer(immutableList, new ImmutableListSerializer());
-    kryo.addDefaultSerializer(immutableSet, new ImmutableSetSerializer());
-    kryo.addDefaultSerializer(immutableMap, new ImmutableMapSerializer());
+    // MapReferenceResolver is Kryo's default resolver, so subclassing it keeps all standard
+    // reference behavior for every other type; only the immutable-collection ids are rewritten.
+    ImmutableAwareReferenceResolver resolver = new ImmutableAwareReferenceResolver();
+    kryo.setReferenceResolver(resolver);
+    kryo.addDefaultSerializer(immutableList, new ImmutableListSerializer(resolver));
+    kryo.addDefaultSerializer(immutableSet, new ImmutableSetSerializer(resolver));
+    kryo.addDefaultSerializer(immutableMap, new ImmutableMapSerializer(resolver));
   }
 
   /** Reads elements onto a mutable buffer through Kryo, then rebuilds an immutable {@code List}. */
   private static final class ImmutableListSerializer extends CollectionSerializer {
-    ImmutableListSerializer() {
+    private final ImmutableAwareReferenceResolver resolver;
+
+    ImmutableListSerializer(ImmutableAwareReferenceResolver resolver) {
+      this.resolver = resolver;
       setElementsCanBeNull(false);
     }
 
@@ -77,14 +108,20 @@ public final class ImmutableCollectionsSerializers {
 
     @Override
     public Collection read(Kryo kryo, Input input, Class<Collection> type) {
+      resolver.arm();
       Collection buffer = super.read(kryo, input, type);
-      return buffer == null ? null : List.copyOf(buffer);
+      Collection immutable = buffer == null ? null : List.copyOf(buffer);
+      resolver.finish(immutable);
+      return immutable;
     }
   }
 
   /** Reads elements onto a mutable buffer through Kryo, then rebuilds an immutable {@code Set}. */
   private static final class ImmutableSetSerializer extends CollectionSerializer {
-    ImmutableSetSerializer() {
+    private final ImmutableAwareReferenceResolver resolver;
+
+    ImmutableSetSerializer(ImmutableAwareReferenceResolver resolver) {
+      this.resolver = resolver;
       setElementsCanBeNull(false);
     }
 
@@ -95,14 +132,20 @@ public final class ImmutableCollectionsSerializers {
 
     @Override
     public Collection read(Kryo kryo, Input input, Class<Collection> type) {
+      resolver.arm();
       Collection buffer = super.read(kryo, input, type);
-      return buffer == null ? null : Set.copyOf(buffer);
+      Collection immutable = buffer == null ? null : Set.copyOf(buffer);
+      resolver.finish(immutable);
+      return immutable;
     }
   }
 
   /** Reads entries onto a mutable buffer through Kryo, then rebuilds an immutable {@code Map}. */
   private static final class ImmutableMapSerializer extends MapSerializer {
-    ImmutableMapSerializer() {
+    private final ImmutableAwareReferenceResolver resolver;
+
+    ImmutableMapSerializer(ImmutableAwareReferenceResolver resolver) {
+      this.resolver = resolver;
       setKeysCanBeNull(false);
       setValuesCanBeNull(false);
     }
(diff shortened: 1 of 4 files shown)


sezruby:
Addressed in the latest push. The fix no longer uses `JavaSerializer`. It now registers scoped Kryo `CollectionSerializer`/`MapSerializer` implementations (ported from Kryo 5's `ImmutableCollectionsSerializers`) that read every element through Kryo and rebuild the collection via its public factory (`List`/`Set`/`Map.copyOf`). No value passes through `ObjectInputStream.readObject`, so the deserialization-gadget path at the caller-controlled `decode()` boundary is closed.

Added a regression test (`doesNotInvokeReadObjectOfNestedElements`): it encodes an immutable list holding a `Serializable` element whose `readObject` sets a flag, decodes it, and asserts the flag stays `false`. Verified against Kryo 4.0.2 — the previous `JavaSerializer` approach set it `true`; the new serializers leave it `false` while still returning a genuine `ImmutableCollections` instance.

lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    @Override
+    protected Collection create(Kryo kryo, Input input, Class<Collection> type) {
+      return new ArrayList<>();
+    }
+
+    @Override
+    public Collection read(Kryo kryo, Input input, Class<Collection> type) {
+      Collection buffer = super.read(kryo, input, type);
```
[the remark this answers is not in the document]
sezruby: Fixed in bb3cbff.

The rebuild-on-read pattern stays (that's what keeps elements off the `ObjectInputStream.readObject` path), but reference identity is now preserved by an `ImmutableAwareReferenceResolver` (a `MapReferenceResolver` subclass, so every other type keeps Kryo's default reference behavior):

- Each serializer arms the resolver immediately before `super.read`. Kryo's `CollectionSerializer`/`MapSerializer` register the collection via `kryo.reference` right after `create()` and before reading elements, so the first `setReadObject` after arming is the collection's own id. Once `List`/`Set`/`Map.copyOf` has run, `finish()` rewrites that id to the immutable copy, so every later back-reference resolves to the immutable value instead of the staging buffer.
- Cyclic graphs: a pure-immutable cycle can't exist (a collection can't contain itself). A cycle through a mutable member is supported when the collection isn't the object the cycle re-enters (the member is registered before its back-reference is read). The one unsupportable shape — a back-reference re-entering an immutable collection while it's still being read, before its immutable value exists — is detected via a pending-id set and rejected with a clear `KryoException` rather than leaking the buffer.

Your reproducer is now a passing test, `preservesReferenceIdentityOfAliasedImmutableCollections`: `assertSame` holds for aliased list/set/map and `add()`/`put()` throws. Added `supportsCycleThroughAMutableObjectHoldingAnImmutableCollection` (mixed cycle round-trips, collection stays immutable) and `rejectsCycleThatReentersAnImmutableCollectionStillBeingRead` (the mid-read re-entry throws).

For reference, Kryo 5's built-in `ImmutableCollectionsSerializers` use the same rebuild-via-factory pattern without this reference fixup, so the aliasing/back-reference behavior here is stricter than upstream Kryo 5.

sezruby:
Addressed the reference-identity finding in bb3cbff (details on the inline thread). Against the review's three requirements:

- **Back-references resolve to the immutable values** — an `ImmutableAwareReferenceResolver` rewrites each collection's Kryo reference id to its immutable copy once rebuilt.
- **Cyclic graphs are explicitly handled** — a mixed cycle where the immutable collection isn't the re-entry point is supported; the one unsupportable shape (a back-reference re-entering a collection still being read) is detected and rejected with a `KryoException`.
- **Regression tests** — aliased list/set/map identity (the posted reproducer), a supported mixed cycle, and the rejected mid-read re-entry. Full `LanceSerializeUtilTest`: 7/7.

On the `breaking-change` label: I don't have permission to set labels on this repo from a fork PR. For what it's worth on the classification, the encode path and the Kryo wire format are unchanged — the change only affects the decode result for the immutable collections, which previously threw `UnsupportedOperationException`/`NullPointerException` on first use, so no previously-working input changes behavior. Happy for a maintainer to apply the label if you'd still like it tracked that way.

sezruby:
@yanghua cc @Xuanwo @wjones127 @jackye1995 @LuciferYang

The gate is green now — `lance-gatekeeper[bot]` approved on bb3cbff (both the Java-deserialization and mutable-alias findings closed), and the Label / PR-title / Gatekeeper checks all pass.

The `Spark`, `Spark Search Docker`, and `Spark AWS Integration` workflows are still in `action_required` since this is a fork PR. Could a maintainer click **Approve and run** so the test suite executes? The bundled `beta.10 -> beta.21` bump makes CI run `OptimizeTest`/`VacuumTest`/`ShowIndexesTest` against the version that exposed the bug, so a green run here also confirms the fix end-to-end.

This unblocks #778 as well — its beta.21 bump currently fails `OptimizeTest` on the exact immutable-collection path this PR fixes.

sezruby:
@yanghua @Xuanwo @wjones127 @jackye1995 @LuciferYang — CI is fully green now: all `Test` / `Integration Test` / `Search Docker` jobs pass across every Spark/Scala combo, gatekeeper approved, and the PR is mergeable/clean.

This is a blocker for updating lance-core in lance-spark. Any bump to `11.0.0-beta.21` or later pulls in `CompactionOptions.excludedFragmentIds` (stored via `List.copyOf(...)`), which the unchanged `LanceSerializeUtil` Kryo codec cannot reconstruct — so `OptimizeExec` shipping a `CompactionTask` to executors throws `NullPointerException` on decode. The version bump bundled here proves that: it makes CI exercise `OptimizeTest`/`VacuumTest`/`ShowIndexesTest` against the failing version, and they now pass with the codec fix.

Could a maintainer merge this? It unblocks the lance-core bump for lance-spark generally, and #778 specifically (which currently fails `OptimizeTest` on this exact path).

lance-spark-base_2.12/src/test/java/org/apache/spark/sql/util/LanceSerializeUtilTest.java
```diff
@@ -0,0 +1,189 @@
...
+public class LanceSerializeUtilTest {
+
+  private static <T> T roundtrip(T obj) {
+    return LanceSerializeUtil.decode(LanceSerializeUtil.encode(obj));
+  }
+
+  // The JDK immutable collections (List.of/copyOf, Set.of, Map.of) have no no-arg constructor and
+  // keep their storage in a java.base-internal field, so the codec's Objenesis instantiator could
+  // not reconstruct them: a decoded collection threw on first use. The size()/contents assertions
+  // below fail on the pre-fix codec and pass once the immutable classes delegate to JavaSerializer.
```
yanghua: // The JDK immutable collections (List.of/copyOf, Set.of, Map.of) have no no-arg constructor and
// keep their storage in a java.base-internal field, so the codec's Objenesis instantiator could
// not reconstruct them. The assertions below pass once dedicated Kryo serializers rebuild the
// collections through their public copyOf factories without invoking Java serialization.

lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
```diff
@@ -0,0 +1,240 @@
...
+    private final ImmutableAwareReferenceResolver resolver;
+
+    ImmutableListSerializer(ImmutableAwareReferenceResolver resolver) {
+      this.resolver = resolver;
+      setElementsCanBeNull(false);
+    }
+
+    @Override
+    protected Collection create(Kryo kryo, Input input, Class<Collection> type) {
+      return new ArrayList<>();
```
yanghua: Use `ArrayDeque` to avoid malformed lengths from triggering large‑memory pre‑allocation?

```
return new ArrayDeque<>();
```

yanghua (commented):
Left some comments.

lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
```diff
@@ -0,0 +1,240 @@
...
+    // registering on the base covers every size variant (List12/ListN, Set12/SetN, Map1/MapN) and
+    // sub-lists. copyOf(empty*) sidesteps the overloaded of() factories.
+    Class<?> immutableList = List.copyOf(Collections.emptyList()).getClass().getSuperclass();
+    Class<?> immutableSet = Set.copyOf(Collections.emptySet()).getClass().getSuperclass();
+    Class<?> immutableMap = Map.of().getClass().getSuperclass();
+    // MapReferenceResolver is Kryo's default resolver, so subclassing it keeps all standard
+    // reference behavior for every other type; only the immutable-collection ids are rewritten.
+    ImmutableAwareReferenceResolver resolver = new ImmutableAwareReferenceResolver();
+    kryo.setReferenceResolver(resolver);
+    kryo.addDefaultSerializer(immutableList, new ImmutableListSerializer(resolver));
```
LuciferYang, line 93: The Changes section of the PR description still describes the earlier, rejected approach: registering Kryo's `JavaSerializer` on the abstract base classes (and says sub-lists are out of scope). The code in this PR does the opposite on both counts — these serializers never touch `JavaSerializer`, and the class javadoc explains why: `decode()` is reachable from caller-controlled write options, and `JavaSerializer` would push every nested element through `ObjectInputStream.readObject`, opening a deserialization gadget right at that entry point.

Could you update the description to match the implementation, and also cover the two designs the body omits: the reference-identity rewrite in `ImmutableAwareReferenceResolver` and the fail-fast on cyclic references?

lance-spark-base_2.12/src/test/java/org/apache/spark/sql/util/LanceSerializeUtilTest.java
LuciferYang: The same stale narrative leaked into this comment: the assertions pass because the immutable collections are read onto a staging buffer via the Kryo path and rebuilt with `List.copyOf`/`Set.copyOf`/`Map.copyOf` — not because anything delegates to `JavaSerializer` (which the javadoc in `ImmutableCollectionsSerializers` explicitly rules out, and `doesNotInvokeReadObjectOfNestedElements` locks down).

Worth fixing together with the description update, so the repo doesn't carry a comment that contradicts its own security test.

lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
sezruby: Good catch — adopted in 99b9fe0. Verified against the bundled Kryo 4.0.2: `CollectionSerializer.read` calls `((ArrayList) collection).ensureCapacity(length)` with the wire-supplied length *before* any element is read, so on this caller-reachable path a malformed huge length would eagerly allocate a giant backing array. Both the list and set serializers now stage in an `ArrayDeque` (which sidesteps that `instanceof ArrayList` branch and preserves insertion order for the `List.copyOf` rebuild). The map serializer keeps `HashMap` — `MapSerializer.read` has no analogous length-driven pre-size.

lance-spark-base_2.12/src/test/java/org/apache/spark/sql/util/LanceSerializeUtilTest.java
sezruby: Done in 99b9fe0 — used your suggested wording: the comment no longer mentions `JavaSerializer` and now states the assertions pass because dedicated Kryo serializers rebuild the collections through their public `copyOf` factories without invoking Java serialization.

lance-spark-base_2.12/src/main/java/org/apache/spark/sql/util/ImmutableCollectionsSerializers.java
sezruby: Updated the PR description to match the implementation. The Changes section no longer describes the `JavaSerializer` approach; it now documents the scoped Kryo serializers (elements stay on the Kryo path, collection rebuilt via `List`/`Set`/`Map.copyOf`, so no `ObjectInputStream.readObject`), notes that immutable sub-lists are handled (rebuilt by iteration), and adds the two designs the body omitted — the reference-identity rewrite in `ImmutableAwareReferenceResolver` and the fail-fast on the unsupportable cyclic shape.

lance-spark-base_2.12/src/test/java/org/apache/spark/sql/util/LanceSerializeUtilTest.java
sezruby: Fixed in 99b9fe0 alongside the description update — the comment now says the assertions pass because the collections are read onto a staging buffer via the Kryo path and rebuilt with `List`/`Set`/`Map.copyOf`, with no delegation to `JavaSerializer` (as `doesNotInvokeReadObjectOfNestedElements` enforces).

yanghua (approved):
+1

PR #758, opened 2026-08-11
feat: balance scalar index fragment batches

ddupg:
## Summary

- Balance scalar index fragment batches using fragment row counts.
- Keep assignments deterministic and document the non-contiguous batching behavior.

## Testing

- Spark 3.5 / Scala 2.12: AddIndexTest (48 tests)
- Scala 2.12 and 2.13: IndexUtilsTest (26 tests each)

Closes #757

PR #784, opened 2026-08-26
feat: add distributed REFRESH INDEX for incremental index maintenance

ivscheianu:
Part of #789. Design: #788.

> **Being split.** Five independent pieces of this change have moved to their own PRs against `main`
> (checklist in #789): option-name normalization (#791), contiguous index segment coverage (#792),
> naming the index on uncommitted segment builds (#793), the `SHOW INDEXES` coverage columns (#794),
> and version-consistent builds with real coverage reporting (#795). The first three are pre-existing
> bugs on `main`.
>
> **The diff below is still the whole change set, and stays that way until those five land.** It does
> not compile without them: `RefreshIndexExec` calls `IndexUtils.committedCoverage`, `liveFragmentIds`
> and `pinVersion` from #795, and `visitRefreshIndex` calls `normalizedOptionName` from #791. Once
> they are merged this is rebased down to the `REFRESH INDEX` slice: 22 files, +2235, of which 574 is
> production code — below the comparable precedents in this repo (#654 tag ops at 555, #576 branch
> ops at 622). The remainder is 1413 lines of tests and 248 of docs.

## What

`CREATE INDEX` rebuilds every fragment, so keeping an index current on an append-heavy table means re-indexing the whole table. This adds an incremental counterpart:

```sql
ALTER TABLE lance.db.users REFRESH INDEX user_id_idx [WITH (num_segments = 8)];
```

It indexes only the fragments the index does not cover, distributed the same way `CREATE INDEX` is: the driver diffs coverage against the table and balances the remainder by row count, executors build one uncommitted segment per batch, and the driver commits them as one logical index. Lance core keeps existing segments whose fragments are disjoint from the incoming ones, so prior coverage survives the commit.

A deferred index (`train = false`) covers nothing, so refreshing one builds the whole table through the same distributed path — replacing the current guidance to call `Dataset.optimizeIndices`, which runs on a single node.

## Also included

**Version pinning for distributed index builds.** Tasks open the dataset themselves, and `pinLoadedBranch` only pins when the ref is a branch, so the driver planned over one version while each task independently resolved the latest. The segmented build path now pins the planning version, and the commit accounts for the coverage it actually establishes: what the segments declare, restricted to the fragments still live, with anything retired mid-build named in a warning and excluded from the reported count. It fails only when *nothing* would be covered.

Reporting the achieved coverage rather than aborting is deliberate. A fragment leaves the manifest either because its rows moved (compaction, an in-place rewrite) or because they were all deleted; only the first leaves data unindexed, and in both cases the segments for what remains are correct. Discarding a finished distributed build would be the wrong response, and it would make a routine concurrent `DELETE` fatal on a long build.

Range-mode BTree stays on unpinned read options: it reads the table back through the catalog, which resolves its own version, so pinning would make the segment record a dataset version older than its own contents. The same coverage rule covers both build paths, so range mode is not judged against a fragment list its coverage does not come from.

**Index coverage in `SHOW INDEXES`**: `indexed_percent` (truncated, so it never overstates; null for an empty table), `num_segments`, `size_bytes` — all derived from metadata the command already fetches, so no extra driver cost.

**Contiguous segment coverage.** Lance only compacts fragments covered by an identical set of index segments, so batches whose fragment ids interleave leave `OPTIMIZE` nothing to group — measured: a 2-segment index blocked compaction of a table that compacted fine unindexed. Batches are contiguous runs again (as they were before #758) while still balanced by row count: boundaries go where an even split would fall, moved to the nearest fragment. Deterministic, and exactly `num_segments` batches even under skew.

**WITH-clause option names are normalized** to lower case at the parser boundary, so `WITH (NUM_SEGMENTS = 8)` is the option it looks like. ANTLR reports identifier text as written and every consumer matched lower-case literals, so an uppercase spelling silently took the default and was forwarded to the index backend as a parameter. This fixes `CREATE INDEX`, `OPTIMIZE` and `VACUUM` the same way.

**Uncommitted segment builds name their index and set `replace`.** Letting Lance derive the segment's name yields `<column>_idx` — the name it assigns an unnamed index — so its collision pre-check fired against the very index being refreshed, and `REFRESH INDEX user_id_idx` on column `user_id` could never succeed. On the uncommitted path `replace` gates only that pre-check; the driver's single `commitExistingIndexSegments` transaction still decides which segments to keep. The explicit name also moves the genuine "already exists with different fields" check ahead of the distributed build instead of after it.

**A `DROP INDEX` during a refresh no longer resurrects the index.** The commit re-resolves the index on a fresh handle, because committing segments under a name Lance no longer knows creates that index rather than extending it.

**Fragment enumeration** uses `Dataset.getFragmentStatistics()` (primitive arrays) instead of `getFragments()`, which materializes a Java object per fragment and per data file on each driver-side pass.

`ScalarSegmentIndexJob` was made exec-agnostic so both commands share one build path instead of duplicating it.

## Not in scope

- **Distributed segment merge.** Each refresh adds segments; consolidating them still requires a full `CREATE INDEX`. Because Lance only compacts fragments under an identical segment set, accumulated refreshes progressively narrow what `OPTIMIZE` can coalesce. Documented, with the guidance to run `OPTIMIZE` before `REFRESH INDEX` rather than after.
- **Vector indexes.** Still not creatable or refreshable from Spark SQL; unchanged here.
- **Inheriting build parameters.** Lance records a built index's parameters inside the index, and its Java API exposes no way to read them back — `Dataset.optimizeIndices` derives them internally but runs on a single node, which is the path this replaces. So `REFRESH INDEX` takes options from the `WITH` clause and falls back to type defaults, and the docs say to repeat the original options.

  For most methods each segment is queried independently, so a mismatch changes performance rather than results (verified: a `btree` refreshed with a different `zone_size` still answers point and range queries correctly). `fts`/`inverted` is the exception: it is read with one configuration for all of its segments, and a set that disagrees fails *every* full-text query while `SHOW INDEXES` still reports full coverage. So the refresh compares the index details of the segments it built against the ones they would join and **fails without committing** when they differ — the segments are uncommitted at that point, so the index is left exactly as it was, and repeating the original options succeeds. A follow-up in lance core exposing `derive_index_params()` through the Java binding would let a refresh inherit the configuration and remove the caveat entirely.
- **A pre-existing `zonemap` read bug.** A partially covered `zonemap` index prunes the fragments it does not cover, so a predicate on the indexed column can return fewer rows than the table holds (`COUNT(*)` over the table stays correct). It is the connector's own driver-side pruning: zone statistics come from the committed segments only, and `ZonemapFragmentPruner` builds its surviving fragment set purely from the fragments those zones name, with no gate on coverage, so a fragment no segment covers contributes no zones and is pruned away. Reproducible with `CREATE INDEX` alone, so not introduced here, and a coverage gate on the pruner would close it. Documented for now, since `indexed_percent` exists precisely to tell operators when a refresh is needed. `btree`, `bitmap` and `bloomfilter` return complete results while partially covered.

  > **Edit:** fixed by #781.
- Multi-field/covering indexes and legacy segments that cannot be partially placed are rejected with a message pointing at `CREATE INDEX`.

### DIFF
--- a/docs/src/operations/ddl/.pages
+++ b/docs/src/operations/ddl/.pages
@@ -11,6 +11,8 @@ nav:
   - show-tblproperties.md
   - drop-table.md
   - create-index.md
+  - refresh-index.md
+  - drop-index.md
   - show-indexes.md
   - create-branch.md
   - drop-branch.md
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -53,7 +53,13 @@ The distributed build used by `zonemap`, `bitmap`, `label_list`, `ngram`, `bloom
 
 | Option         | Type    | Description                                                                                                                                                                                                                        |
 |----------------|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `num_segments` | Integer | Target number of parallel build tasks (upper bound; clamped to fragment count when larger). Fragments are assigned by row count to balance estimated task workloads. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+| `num_segments` | Integer | Number of parallel build tasks, and so of index segments created (clamped to fragment count when larger). Each task takes a contiguous run of fragments, sized to balance estimated workloads by row count. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+
+Option names are case-insensitive: `WITH (NUM_SEGMENTS = 8)` and `WITH (num_segments = 8)` are the same option.
+
+Contiguous coverage matters beyond parallelism: [OPTIMIZE](./optimize.md) can only group fragments
+that the identical set of index segments covers, so segments whose fragment ids interleave leave it
+nothing to coalesce.
 
 ### ZoneMap Options
 
@@ -254,7 +260,8 @@ to scanning the data until it is populated. There are two ways to populate it:
 
 - **[REFRESH INDEX](./refresh-index.md) (recommended):** builds only the fragments the index does not
   cover, distributed across Spark tasks. For a deferred index that is the whole table; afterwards it
-  is whatever has been appended since:
+  is whatever has been appended since. Repeat any method options here, since a deferred index holds
+  no built data to inherit them from:
 
     ```sql
     ALTER TABLE lance.db.users REFRESH INDEX idx_id;
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -252,28 +252,29 @@ fill it in later, or when you intend to build it incrementally:
 A deferred index returns `fragments_indexed = 0` and is treated as fully unindexed: queries fall back
 to scanning the data until it is populated. There are two ways to populate it:
 
-- **Full distributed build (recommended):** re-run `CREATE INDEX` with the same name. This uses the
-  normal distributed build across Spark tasks and atomically replaces the empty index:
+- **[REFRESH INDEX](./refresh-index.md) (recommended):** builds only the fragments the index does not
+  cover, distributed across Spark tasks. For a deferred index that is the whole table; afterwards it
+  is whatever has been appended since:
+
+    ```sql
+    ALTER TABLE lance.db.users REFRESH INDEX idx_id;
+    ```
+
+- **Re-run `CREATE INDEX`** with the same name. This rebuilds every fragment and atomically replaces
+  the existing index, which is what you want when you also need to change the index options or
+  consolidate accumulated segments:
 
     ```sql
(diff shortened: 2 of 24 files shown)


ivscheianu:
First two are fixed. The concurrent-drop one is real, but I don't think the connector can close it: `commitExistingIndexSegments` takes no expected predecessor, and a compensating drop would have the same race. It needs lance-format/lance#6806, so for now I've added a disabled regression pointing at that plus a docs caveat.

ivscheianu:
Superseded by two stacked PRs rebased onto current `main`:

1. **#821** — refactor: decouple index build utilities for shared use (the prerequisite refactoring)
2. **#822** — feat: add distributed REFRESH INDEX for incremental index maintenance (the feature itself)

The five independent bug fixes from the original checklist (#791–#795) landed separately; these two PRs contain only the REFRESH INDEX delta.

PR #785, opened 2026-08-26
docs: add SET UNENFORCED PRIMARY KEY page and fix nav gaps

jackylee-ch:
## Summary
- Document `ALTER TABLE ... SET UNENFORCED PRIMARY KEY`, the only extension statement with no page
- Add `drop-index.md` and `insert-overwrite.md` to the nav; both were unreachable

## Test plan
- [x] `mkdocs build --strict`

### DIFF
--- a/docs/src/operations/ddl/set-unenforced-primary-key.md
+++ b/docs/src/operations/ddl/set-unenforced-primary-key.md
@@ -1,0 +1,56 @@
+# SET UNENFORCED PRIMARY KEY
+
+Declare primary key columns on a Lance table.
+
+!!! warning "Spark Extension Required"
+    This feature requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+## Overview
+
+The `SET UNENFORCED PRIMARY KEY` command records the primary key columns of a table as schema field metadata. Uniqueness is **not** enforced: no write path validates the declared columns, so duplicate values are accepted.
+
+## Syntax
+
+=== "SQL"
+    ```sql
+    ALTER TABLE <table> SET UNENFORCED PRIMARY KEY (<column> [, <column> ...]);
+    ```
+
+Column order is significant and is recorded alongside the declaration.
+
+## Examples
+
+### Single-column primary key
+
+=== "SQL"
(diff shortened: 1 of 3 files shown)


jackylee-ch:
cc @LuciferYang

jackylee-ch:
cc @LuciferYang

docs/src/operations/ddl/set-unenforced-primary-key.md
```diff
@@ -0,0 +1,56 @@
...
+The `SET UNENFORCED PRIMARY KEY` command returns the following information:
+
+| Column                | Type   | Description                                                  |
+|-----------------------|--------|--------------------------------------------------------------|
+| `status`              | String | The result status (`OK`).                                    |
+| `primary_key_columns` | String | The declared columns, comma-separated, in the order given.    |
+
+## How It Works
+
+The command commits a new table version that attaches `lance-schema:unenforced-primary-key` field metadata to each declared column, together with its one-based position in the key. Because the change is a normal commit, earlier versions keep their original schema and remain available for time travel.
```
LuciferYang: The "How It Works" line reads as if the one-based position is stored in the `lance-schema:unenforced-primary-key` key, but the implementation writes two keys: the main key holds the marker value `"true"`, and the position lives in a companion key `lance-schema:unenforced-primary-key:position`. A reader inspecting schema metadata for the position would look at the wrong key.

Worth spelling out both keys here. Non-blocking.

LuciferYang (commented):
Checked each claim in the new page against `SetUnenforcedPrimaryKey`'s exec and plan classes and they all match: metadata key, 1-based position, the two output columns, write-once, and the non-null / primitive-leaf / duplicate-column / branch-tag rejections. Both nav additions resolve and the config link is valid. One optional wording nit inline, non-blocking.

jackylee-ch:
Fixed in f733844: the How It Works line now names both keys — `lance-schema:unenforced-primary-key` holding the `true` marker and the companion `lance-schema:unenforced-primary-key:position` holding the one-based position.

LuciferYang (approved):
Approving. The inline note about the two metadata keys is an optional wording tweak and doesn't block merge.

LuciferYang:

PR #786, opened 2026-08-26
docs: correct the SEARCH table function reference

jackylee-ch:
## Summary
- `search_columns` is required with no positional slot, so `SEARCH` needs named args (Spark 3.5+)
- Drop the positional example, which always throws, and `offset`, which is never read
- Fix Execution (scan path is namespace-dependent) and clarify the current test markers

## Test plan
- [x] `mkdocs build --strict`


docs/src/operations/dql/search.md
```diff
@@ -72,8 +61,8 @@ The result includes the requested table columns and a nullable `_score` float co
 
 ## Execution
 
-Spark plans `SEARCH` as a DataSource V2 batch read with one input partition. The partition reader calls the Lance namespace `queryTable` API. With a directory namespace the search runs in the Spark process executing that reader; with a REST namespace the REST server handles the namespace request.
+Spark plans `SEARCH` as a batch read carrying the full-text query as a scan option, wrapped in an optional filter, a projection, `ORDER BY _score DESC`, and `LIMIT k`. The scan then runs one of two ways: a single-partition server-side read through the Lance namespace `queryTable` API when the namespace supports it, or a distributed per-fragment scan for catalog-only namespaces and for reads that target a branch or tag.
```
geruh: I haven't see how search can target a branch or tag is this true or should we drop?

jackylee-ch: Tag is not reachable, you are right to push on it: `version` goes through `optionalLong`, so a tag name never parses, and there is no `tag_` identifier suffix to carry one. Branch is reachable, since `resolveLanceTable` calls `loadTable` and that matches `BRANCH_SUFFIX`, so `table => '...docs.branch_audit'` lands on the branch and `shouldNamespaceFtsScan` falls back, but nothing tests that path. Dropped the clause in 52baafa and left only the catalog-only condition. Happy to put branch back if you want it documented.

jackylee-ch: Two corrections to my last reply. I over-corrected by dropping branch as well as tag, and `nothing tests that path` was wrong — `LanceScanBuilderTest#testBranchFullTextQueryDoesNotUseNamespaceScan` (LanceScanBuilderTest.java:374) already asserts `shouldNamespaceFtsScan()` is false for `LanceRef.ofBranch`. 4ea4cb3 makes the per-fragment scan the general fallback and lists the three conditions the server-side route needs.

docs/src/operations/dql/search.md
```diff
@@ -72,8 +61,8 @@ The result includes the requested table columns and a nullable `_score` float co
 
 ## Execution
 
-Spark plans `SEARCH` as a DataSource V2 batch read with one input partition. The partition reader calls the Lance namespace `queryTable` API. With a directory namespace the search runs in the Spark process executing that reader; with a REST namespace the REST server handles the namespace request.
+Spark plans `SEARCH` as a batch read carrying the full-text query as a scan option, wrapped in an optional filter, a projection, `ORDER BY _score DESC`, and `LIMIT k`. The scan then runs one of two ways: a single-partition server-side read through the Lance namespace `queryTable` API when the namespace supports it, or a distributed per-fragment scan for catalog-only namespaces.
```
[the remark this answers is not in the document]
jackylee-ch: Fixed in 4ea4cb3, and thanks for the reproducer. I took the `otherwise` phrasing rather than re-listing triggers, because enumerating them went stale twice: the third condition, `pushedAggregation.isPresent()`, was never documented either. The paragraph now states the fallback as the general case and lists what the server-side route requires.

jackylee-ch:
cc @LuciferYang

docs/src/operations/dql/search.md
```diff
@@ -72,8 +61,8 @@ The result includes the requested table columns and a nullable `_score` float co
...
+Spark plans `SEARCH` as a batch read carrying the full-text query as a scan option, wrapped in an optional filter, a projection, `ORDER BY _score DESC`, and `LIMIT k`. The scan then runs either as a single-partition server-side read through the Lance namespace `queryTable` API, or otherwise as a distributed per-fragment scan. The server-side route requires a namespace that implements `queryTable`, a read that does not target a branch or tag, and no pushed-down aggregation.
 
 ## Validation
 
-The Docker integration suite covers `SEARCH` against the directory namespace and a REST namespace backed by a directory namespace. The `Spark Search Docker` GitHub Actions workflow runs both backends for pull requests.
+The `Spark Search Docker` workflow exercises `SEARCH` against directory and REST-directory namespaces. The Docker test still carries an `xfail` marker, but it can pass as `XPASS`; the JVM `SEARCH` cases remain `@Disabled`. [`VECTOR_SEARCH`](vector-search.md) and [`HYBRID_SEARCH`](hybrid-search.md) are also exercised by the workflow.
```
LuciferYang: The Validation section leans on test-framework jargon (`xfail`, `XPASS`, `@Disabled`) that means little to someone reading the feature reference. More importantly, the reason those markers exist is that SEARCH's server-side `queryTable` route currently ignores the structured FTS query and cannot produce the `_score` column that the plan always projects and sorts by, so the query fails on the `_score` projection (the Python case is `xfail` with `raises=Exception`). That is exactly the path the Basic Usage example above takes on a `queryTable`-capable namespace such as a directory namespace, where it selects `_score`. So the section reads like "tested and passing" while the feature's headline example is actually broken.

Consider collapsing the CI mechanics into a one-line coverage statement and adding a short known-limitation note near the top: on a `queryTable`-capable namespace SEARCH currently fails on the `_score` projection. Leave the `xfail`/`@Disabled` details in the test code.

docs/src/operations/dql/search.md
```diff
@@ -6,10 +6,10 @@ Run Lance full-text search from Spark SQL using Lance namespace execution.
     `SEARCH` requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
 
 !!! note "Namespace Tables Required"
-    `SEARCH` resolves the `table` argument through a Spark catalog and executes through the Lance namespace `queryTable` API. Use a Lance namespace catalog table such as `lance.default.documents`, not a raw Lance dataset path.
+    `SEARCH` resolves the `table` argument through a Spark catalog. Use a Lance namespace catalog table such as `lance.default.documents`, not a raw Lance dataset path.
 
-!!! note "Named Arguments"
-    Named arguments require Spark 3.5 or later. On Spark 3.4, use the positional form.
+!!! note "Named Arguments Required"
+    `search_columns` is required and has no positional slot, so `SEARCH` must be called with named arguments. Named arguments require Spark 3.5 or later, so `SEARCH` is not available on Spark 3.4.
```
LuciferYang: `SEARCH is not available on Spark 3.4` is slightly off: the 3.4 module registers SEARCH identically to 3.5/4.0 (no version gating), so the function is visible. What actually blocks it is that `search_columns` is required and cannot be passed positionally, only as a named argument, and named arguments need Spark 3.5+.

The conclusion (unusable on 3.4) is right, but "not available" reads as "not registered". Consider "cannot be used on Spark 3.4", or spell out that it is registered but uncallable because 3.4 lacks named-argument support.

LuciferYang (commented):
Thanks for the cleanup — the corrections all match the code: `search_columns` is required with no positional slot, `offset` is never read on the SEARCH path, and the two-route Execution rewrite (server-side `queryTable` vs per-fragment scan) lines up with `LanceScanBuilder`. Two documentation notes below, both non-blocking.

PR #787, opened 2026-08-26
fix: round half-precision subnormals instead of flushing to zero

jackylee-ch:
## Summary
`floatToHalf` flushed everything below 2^-24 to zero, but under round-to-nearest-even only the
midpoint 2^-25 ties to zero; values in (2^-25, 2^-24) round up to the smallest subnormal. That
band became ±0.0. Exponent -25 now falls through to the existing subnormal path, which already
ties correctly. Affects `ARRAY<FLOAT>` columns marked `'<col>.arrow.float16' = 'true'`.

## Test plan
- [x] `Float16UtilsTest`: the midpoint band fails on `< -24`, and the pinned exponent -33 inputs
      fail if the guard is loosened to `< -33`
- [x] Exhaustive differential sweep of all 2^32 bit patterns against JDK 25 `Float.floatToFloat16`
      and Arrow `Float16.toFloat16`: 4278190080 finite inputs and both infinities agree, 0
      mismatches. NaN payload handling differs from Arrow and is untouched by this PR
- [x] The behaviour delta against the old guard is exactly (2^-25, 2^-24) for both signs
      (16777214 = 2 x (2^23 - 1)), so the write path now matches Arrow's own float-to-half
- [x] float16 suites on Spark 4.1/Scala 2.13 and 3.5/Scala 2.12; `make lint`


jackylee-ch:
@LuciferYang Could you approve the GHA?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
```diff
@@ -209,8 +209,9 @@ public static short floatToHalf(float value) {
       return (short) (halfSign | 0x7C00);
     }
 
-    if (unbiasedExponent < -24) {
-      // Too small for even subnormal representation: flush to zero
+    if (unbiasedExponent < -25) {
```
LuciferYang, line 213: The comment explains the numeric boundary, but the `< -25` early return has a second job it does not mention. In the subnormal branch `int shift = -14 - unbiasedExponent + 13;` is -1 - unbiasedExponent, which reaches 32 at exponent -33. There Java masks the shift count of `mantissa >>> shift` to 5 bits, so it shifts by nothing, and one exponent lower `roundBit = 1 << (shift - 1)` wraps too. With the guard at -25 the largest shift is 24, so nothing is broken today.

I removed the branch and reran it: `1.0e-10f` (exponent -34) comes back as 0xF380, which is -15360, instead of 0.

Worth a sentence in the comment: this branch also keeps shift <= 24, so it cannot be folded away.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
```diff
@@ -209,8 +209,9 @@ public static short floatToHalf(float value) {
...
+      // Below the midpoint between zero and the smallest subnormal (2^-25): flush to zero.
```
LuciferYang: The javadoc of `floatToHalf` still says `underflow (flushes to zero)` (Float16Utils.java:174), which no longer matches the code: only `|v| <= 2^-25` flushes, and (2^-25, 2^-24) rounds up to the smallest subnormal.

The same rule sits in the test comment at `Float16UtilsTest.java:100`, which now contradicts `testRoundingBelowSmallestSubnormal` right below it. And in this comment the parenthetical sits right after `the smallest subnormal`, so it reads as saying the smallest subnormal is 2^-25; it is 2^-24.

Three small edits: make the javadoc say it flushes only at or below the midpoint 2^-25 and rounds to nearest even elsewhere, make the test comment say "at or below 2^-25", and move `(2^-25)` next to `the midpoint` here.

LuciferYang (commented):
The fix is correct and complete. I transliterated `floatToHalf`/`halfToFloat` into a standalone class and swept every float32 bit pattern against JDK 20's `Float.floatToFloat16`: 16777214 mismatches before the change, which is exactly the open band (2^-25, 2^-24) for both signs (2 x (2^23 - 1)), and 0 after. All 65536 half patterns also round-trip through `halfToFloat` then `floatToHalf` unchanged.

Two non-blocking notes inline, both about comments rather than behavior.

The commit message's bit-identical claim needs one caveat: it holds for finite inputs, not for NaN. Arrow canonicalizes every NaN to 0x7E00/0xFE00 while this code keeps the top 10 mantissa bits, so the two agree only on NaNs whose top 10 bits are already 0x200. We also can't simply call that class instead: it first ships in Arrow 16.0.0, while the base module compiles against arrow-memory-core 15.0.2 and the 3.4 module pins 14.0.2.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
jackylee-ch: Confirmed — I reproduced it: with the branch removed, `1.0e-10f` (exponent -34, shift 33 masked to 1) returns 0xF380. Exponent -33 happens to still give 0 because the mantissa truncates away, so -34 is where it first shows. Added the sentence in 43acf7e. Worth noting `testUnderflowToZero` already uses `1.0e-10f`, so the guard is test-covered, not just documented.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
jackylee-ch: All three are right. Fixed the javadoc at :174, the `testUnderflowToZero` comment, and moved `(2^-25)` next to `the midpoint` so it no longer reads as the smallest subnormal. Re-ran the float16 suites on 4.1/2.13 and 3.5/2.12 plus `make lint`.

jackylee-ch:
Pushed 43acf7e with both fixes (appended, not force-pushed). The push reset the workflow approval — @LuciferYang could you re-approve the GHA when you get a chance?

lance-spark-base_2.12/src/test/java/org/lance/spark/utils/Float16UtilsTest.java
```diff
@@ -97,13 +97,29 @@ public void testOverflowToInfinity() {
...
+    // float16 subnormal 2^-24. Round-to-nearest-even sends the midpoint itself to
+    // zero, but everything above it must round up to 2^-24 rather than flush.
+    assertEquals(
+        0x0000, Float16Utils.floatToHalf(0x1p-25f) & 0xFFFF, "2^-25 is a tie and rounds to zero");
+    assertEquals(
+        0x8000, Float16Utils.floatToHalf(-0x1p-25f) & 0xFFFF, "-2^-25 is a tie and rounds to zero");
+    assertEquals(
+        0x0001, Float16Utils.floatToHalf(4.0e-8f) & 0xFFFF, "Above 2^-25 rounds up to 2^-24");
+    assertEquals(
+        0x8001, Float16Utils.floatToHalf(-4.0e-8f) & 0xFFFF, "Below -2^-25 rounds up to -2^-24");
```
LuciferYang: The new test covers the midpoint and the band above it, and `testUnderflowToZero` uses 1.0e-10f at exponent -34, but exponents -33 through -26 have no input at all, and the `>>>` truncation that the guard comment in `Float16Utils.java` warns about starts exactly at exponent -33. I built a variant with the guard at `< -33`: 0x2F7FFFFF comes out as 0xFFFF, and the whole suite still passes because 1.0e-10f keeps flushing to zero.

One assertion would lock that band in: `assertEquals(0x0000, Float16Utils.floatToHalf(Math.nextUp(0x1p-33f)) & 0xFFFF, ...)`; the mirrored negative case expects 0x8000 (negative zero).

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
```diff
@@ -171,8 +171,9 @@ public static float halfToFloat(short halfBits) {
   /**
    * Convert a single-precision (32-bit) float to IEEE 754 half-precision (16-bit) float.
    *
-   * <p>Handles overflow (clamps to +/-Infinity), underflow (flushes to zero), NaN propagation, and
-   * round-to-nearest-even.
+   * <p>Handles overflow (clamps to +/-Infinity), NaN propagation, and round-to-nearest-even.
+   * Magnitudes at or below the midpoint 2^-25 flush to zero; above it round-to-nearest-even
```
LuciferYang: The javadoc sentence "at or below the midpoint 2^-25 flush to zero" is loose at exactly 2^-25: it is not flushed, it takes the subnormal path and lands on zero as an RNE tie, which the guard comment in `Float16Utils.java` states correctly. The comment at the top of `testUnderflowToZero` says the same thing, so both read like FTZ semantics.

Worth aligning both to something like "magnitudes below 2^-25 flush to zero, while 2^-25 itself ties to even and lands on zero".

lance-spark-base_2.12/src/test/java/org/lance/spark/utils/Float16UtilsTest.java
jackylee-ch: Confirmed and pinned. I reproduced your variant: with the guard at `< -33`, `0x2F7FFFFF` returns 0xFFFF, and 841 of 6736 sampled values across exponents -33..-26 diverge from Arrow while the suite stays green. Added `Math.nextUp(0x1p-33f)` and its negative in 4e96290; they fail with `expected: <0> but was: <1>` if the guard is loosened that way.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
jackylee-ch: Right, 2^-25 is not flushed; it takes the subnormal path and ties to even. Reworded both the javadoc and the `testUnderflowToZero` comment to "below 2^-25 flush to zero, while 2^-25 itself ties to even and lands on zero". Also caveated the bit-identical claim in the PR body: it holds for finite inputs, and NaN payload handling differs from Arrow and is unchanged here.

jackylee-ch:
Pushed 4e96290 (appended). Thanks for the exhaustive sweep against `Float.floatToFloat16` — the 16777214 = 2 x (2^23 - 1) count is a much cleaner statement of the band than mine. @LuciferYang could you re-approve the GHA?

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/Float16Utils.java
jackylee-ch: Correction to my earlier reply here: -33 does break without the guard, you were right the first time. I had sampled `0x1p-33f`, which is one of only 128 mantissas at that exponent whose low 16 bits are zero and therefore still truncate to 0; the other 8388480 return garbage (`0x2F000001` -> `0x0001`, `0x2F7FFFFF` -> `0xFFFF`). The comment in 792af71 now names the two failures separately: the `>>>` mask at -33, and `1 << 32` wrapping at -34.

lance-spark-base_2.12/src/test/java/org/lance/spark/utils/Float16UtilsTest.java
jackylee-ch: Also narrowing my own number above: with the guard at `< -33` the divergences are all at exponent -33. Exponents -32..-26 still return 0 because shift 25..31 truncates the whole mantissa, so the band I quoted was wider than the actual damage.

jackylee-ch:
Separately: I ran a mutation sweep over `floatToHalf` with the real test class, using JDK 25 `Float.floatToFloat16` to tell real holes from equivalent mutants. Five mutants survive, all in paths this PR does not touch:

- normal-path `roundBit = 1 << 12` -> `1 << 11`, and the normal-path tie condition forced false: 251627520 wrong outputs each, suite green
- deleting the `halfMantissa > 0x3FF` carry block: 122880 wrong outputs, green
- normal-path tie forced true: 30720, green
- masking the subnormal return with `& 0x3FF`: 16384, green

Three assertions close all five: `1.00048828125f -> 0x3C00`, `0x1.ffep0f -> 0x4000`, `0x1.ffcp-15f -> 0x0400`.

Happy to add them here, or keep this PR to the guard and open a follow-up. Which do you prefer?

jackylee-ch:
Went ahead and added them in 0d559bc rather than leave the holes open — four assertions, all in `Float16UtilsTest`, no production change. Each of the five mutants is now killed: the carry assertion kills `1 << 11`, tie-forced-false and deleting the carry block; the exact-midpoint assertion kills tie-forced-true; the subnormal one kills `& 0x3FF`. Happy to split them into a follow-up if you would rather keep this PR to the guard.

LuciferYang (approved):

LuciferYang:
Merged into main, Thanks @jackylee-ch

jackylee-ch:
Thanks for your review. @LuciferYang

Issue #789, opened 2026-08-26
Distributed REFRESH INDEX for incremental scalar index maintenance


ivscheianu:
Design proposal: https://github.com/lance-format/lance-spark/discussions/788

Incremental maintenance for scalar indexes (`ALTER TABLE ... REFRESH INDEX`), plus three pre-existing bugs the work surfaced.

Items 1-3 are independent of the design outcome and can land first: they are bugs on `main` today that affect `CREATE INDEX`, `OPTIMIZE` and `VACUUM`.

- [ ] 1. #791 — **Normalize `WITH` clause option names at the parser boundary.** Option names are compared case-sensitively, while the parser preserves the spelling as written, so `CREATE INDEX ... WITH (TRAIN = false)` is silently ignored and builds a trained index. `OPTIMIZE` and `VACUUM` have the same gap.
- [ ] 2. #792 — **Keep index segment coverage contiguous.** #758 replaced contiguous fragment batching with row-balanced assignment. Lance only compacts fragments covered by an identical set of index segments, so interleaved coverage leaves `OPTIMIZE` with no two fragments it can group, and an indexed table stops coalescing entirely. Row balance and contiguity are compatible.
- [ ] 3. #793 — **Name the index on uncommitted segment builds and set `replace`.** The builds let Lance derive their own name, which is `{column}_idx`, so its collision pre-check fires against an existing index of that name on the same column. Re-running `CREATE INDEX id_idx` on column `id` fails on every task.
- [ ] 4. #794 — **Index coverage in `SHOW INDEXES`**: `indexed_percent`, `num_segments`, `size_bytes`, derived from metadata the command already fetches.
- [ ] 5. #795 — **Make distributed index builds version-consistent and report real coverage.** Tasks
      each resolved the latest version independently instead of the one the driver planned over, and
      `CREATE INDEX` reported the fragment count it planned rather than the one the commit achieved.
- [ ] 6. **`ALTER TABLE ... REFRESH INDEX`** — #784.

Known gaps recorded in the proposal rather than fixed here: a concurrent `DROP INDEX` can be undone by an in-flight refresh (needs lance-format/lance#6806), and a partially covered `zonemap` index prunes uncovered fragments so a predicate on the indexed column can return fewer rows than the table holds (reproducible with `CREATE INDEX` alone).

> **Edit:** the `zonemap` gap is fixed by #781.

PR #791, opened 2026-08-26
fix: normalize WITH-clause option names at the parser boundary

ivscheianu:
Part of #789.

## What

WITH-clause option names are compared case-sensitively, while the parser hands back the identifier exactly as written. ANTLR reports the raw text and `UpperCaseCharStream` upcases only `LA()`, not `getText()`, so an uppercase option name is not the option it looks like.

Every consumer matches lower-case literals, so today:

```sql
ALTER TABLE t CREATE INDEX i USING zonemap (id) WITH (TRAIN = false);
```

builds a **trained** index rather than a deferred one, and `TRAIN` additionally leaks to the Lance index backend as an index parameter, because the `SparkOnlyOptions` filter is case-sensitive too. `OPTIMIZE` and `VACUUM` have the same gap on their own options.

## Change

Normalize the option name once, where it is constructed: a `normalizedOptionName` helper next to `cleanIdentifier` in each `LanceSqlExtensionsAstBuilder`, lower-casing with `Locale.ROOT`. That covers all three named-argument sites (`CREATE INDEX`, `OPTIMIZE`, `VACUUM`) in one place per module, so no consumer can be missed. Normalizing at the boundary rather than at each read site is deliberate: the read sites are spread across several execs and a new one would silently reintroduce the bug.

Note that this lower-cases backtick-quoted option names too. Option names are not column identifiers, so that is the intended behaviour.

### DIFF
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -38,6 +38,8 @@ The following index methods are supported:
 
 The `CREATE INDEX` command supports options via the `WITH` clause to control index creation. These options are specific to the chosen index method.
 
+Option names are case-insensitive: `WITH (TRAIN = false)` and `WITH (train = false)` are the same option.
+
 ### Common Options
 
 These options apply to all index methods:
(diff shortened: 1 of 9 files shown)


fangbo (approved):
Great fix !

PR #792, opened 2026-08-26
fix: keep index segment fragment coverage contiguous

ivscheianu:
Part of #789.

## What

`IndexUtils.batchFragments` assigns fragments to index segments with a least-loaded-first heap. For equal-sized fragments that deals them out round-robin, so segment coverage interleaves fragment ids: four equal fragments with `num_segments = 2` produce segments covering `[0, 2]` and `[1, 3]`.

Lance's compaction planner only groups fragments covered by an identical set of index metadata entries, and every segment is a separate entry. Interleaved coverage therefore puts every adjacent pair of fragments in a different group, so `OPTIMIZE` coalesces nothing at all on an indexed table.

Measured on Spark 3.5, with `OPTIMIZE ... WITH (target_rows_per_fragment = 1000)` over four single-fragment appends: two interleaved segments leave all four fragments untouched, one contiguous segment covering all four compacts them to one, two contiguous segments compact them to two (one per segment), and with no index at all they compact to one.

Batching used to be contiguous. #758 replaced it with row-balanced assignment, and its docs dropped the line "Each task covers a contiguous batch of fragments." This restores contiguity, at a bounded cost in balance.

## Change

The optimal contiguous partition is found exactly rather than approximated. Binary search the smallest row budget a contiguous packing can respect: for a fixed budget, extending each run as far as it will go uses the fewest runs, so the smallest feasible budget *is* the optimal maximum, with the widest single fragment as the search floor. Packing at that budget can use fewer runs than were asked for, which would cost parallelism, so runs are divided at their balance points until the count is reached. Which run gets divided does not matter: every run already fits the budget, so both halves of any split fit it too, and the budget is minimal, so the heaviest batch cannot fall below it either.

Contiguity does cost some balance, which the earlier revision of this PR wrongly claimed it did not. `[10, 9, 8, 7]` into two batches is 19/15 once runs must stay contiguous, where least-loaded-first reached 17/17 by interleaving. Optimal among contiguous splits is the guarantee, not optimal overall, and interleaving costs the whole table its compaction.

The result is contiguous, minimises the heaviest batch over all contiguous partitions, is deterministic regardless of input order, and always has exactly `num_segments` non-empty batches. `Math.addExact` on the prefix sums keeps the pre-existing overflow rejection.

A cheaper heuristic, placing each boundary where an even split would fall and moving it to the nearest fragment, was tried first and **rejected**: it cuts after indivisible leading fragments have already overshot their shares. On `[95, 93, 89, 8, 1, 4, 74, 88, 38]` across six segments it yields a batch of 162 where fragment 0 alone forces 95, so one task ends up 70% heavier than necessary. The least-loaded-first assignment it replaced reached 95, so that would have been a genuine regression of #758's contract rather than a preservation of it.

Docs no longer read as though contiguity restores compaction. A segment boundary is still a grouping boundary, so `num_segments` also bounds how far `OPTIMIZE` can get.

### DIFF
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -53,7 +53,11 @@ The distributed build used by `zonemap`, `bitmap`, `label_list`, `ngram`, `bloom
 
 | Option         | Type    | Description                                                                                                                                                                                                                        |
 |----------------|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `num_segments` | Integer | Target number of parallel build tasks (upper bound; clamped to fragment count when larger). Fragments are assigned by row count to balance estimated task workloads. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+| `num_segments` | Integer | Number of parallel build tasks, and so of index segments created (clamped to fragment count when larger). Each task takes a contiguous run of fragments, and the runs are chosen so the heaviest task is as light as any contiguous split allows. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+
+Contiguous coverage matters beyond parallelism: [OPTIMIZE](./optimize.md) can only group fragments
+that the identical set of index segments covers, so segments whose fragment ids interleave leave it
+nothing to coalesce.
 
 ### ZoneMap Options
 
--- a/docs/src/operations/ddl/create-index.md
+++ b/docs/src/operations/ddl/create-index.md
@@ -56,8 +56,9 @@ The distributed build used by `zonemap`, `bitmap`, `label_list`, `ngram`, `bloom
 | `num_segments` | Integer | Number of parallel build tasks, and so of index segments created (clamped to fragment count when larger). Each task takes a contiguous run of fragments, and the runs are chosen so the heaviest task is as light as any contiguous split allows. Defaults to `min(fragment_count, spark.default.parallelism)`. |
 
 Contiguous coverage matters beyond parallelism: [OPTIMIZE](./optimize.md) can only group fragments
-that the identical set of index segments covers, so segments whose fragment ids interleave leave it
-nothing to coalesce.
+that the identical set of index segments covers. Interleaved segments leave it nothing to coalesce at
+all, while contiguous ones let it coalesce within each segment's run. A segment boundary is still a
+boundary, so `num_segments` also bounds how far compaction can get.
 
 ### ZoneMap Options
 
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -733,6 +733,20 @@ object IndexUtils extends Logging {
     }
   }
 
+  /**
+   * Splits `fragments` into `numSegments` batches, each a contiguous run of fragment ids, chosen so
+   * that the heaviest batch is as light as possible.
+   *
+   * Contiguity is not cosmetic. Lance's compaction planner only groups fragments that are covered by
+   * the identical set of index segments, so batches whose fragment ids interleave leave every
+   * adjacent pair of fragments in a different group and make OPTIMIZE a no-op for the whole table.
+   * Balance is not sacrificed to get it: the optimal contiguous partition is found exactly, so a
+   * workload that the previous least-loaded-first assignment balanced perfectly still is.
+   *
+   * Assignment is deterministic: the same fragments and segment count always produce the same
+   * batches, whatever order `fragments` arrives in. Every batch holds at least one fragment, so the
+   * result always has exactly `segmentCount` entries.
+   */
   def batchFragments(
       fragments: List[FragmentWorkload],
       numSegments: Option[Int],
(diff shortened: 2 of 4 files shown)


ivscheianu:
Both fixed.

The balance finding is correct and the counter-example reproduces exactly: 162 against an optimal 95, and I confirmed the least-loaded-first assignment it replaced also reached 95, so it was a real regression of #758's contract and not merely a suboptimal split. Replaced the heuristic with the exact contiguous minimax you describe — binary search the smallest feasible budget, then split the remaining runs to reach `num_segments`, which only ever lowers the maximum. Your case is now a regression test, alongside one that compares against every contiguous partition of seven shapes at every segment count; that second test also caught three segments on the same workload giving 200 where 188 is optimal.

On the helper: changed, but the stated reason does not hold. `BaseAddIndexTest` already contains 19 raw `Dataset.open()` calls and none of them passes an allocator, so the new helper was following the file's dominant pattern rather than bypassing a required one. I switched it because `Utils.openDatasetBuilder` is the documented path and the same file already uses it in four places, so new code should not add to the older pattern — not because raw open is disallowed. Worth a separate cleanup if you want those 19 converted.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -733,6 +733,20 @@ object IndexUtils extends Logging {
...
   }
 
+  /**
+   * Splits `fragments` into `numSegments` batches, each a contiguous run of fragment ids, chosen so
+   * that the heaviest batch is as light as possible.
+   *
+   * Contiguity is not cosmetic. Lance's compaction planner only groups fragments that are covered by
+   * the identical set of index segments, so batches whose fragment ids interleave leave every
+   * adjacent pair of fragments in a different group and make OPTIMIZE a no-op for the whole table.
+   * Balance is not sacrificed to get it: the optimal contiguous partition is found exactly, so a
```
geruh: i might be missing something here, but i don't think we can say balance isn't sacrificed.

`[10, 9, 8, 7]` with 2 segments can be `17/17` with the old least loaded assignment, but the best contiguous split is `19/15`.

I think the guarantee is that we get the best balance given that the batches have to stay contiguous. can we say that instead?

docs/src/operations/ddl/create-index.md
```diff
@@ -53,7 +53,11 @@ The distributed build used by `zonemap`, `bitmap`, `label_list`, `ngram`, `bloom
...
 |----------------|---------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `num_segments` | Integer | Target number of parallel build tasks (upper bound; clamped to fragment count when larger). Fragments are assigned by row count to balance estimated task workloads. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+| `num_segments` | Integer | Number of parallel build tasks, and so of index segments created (clamped to fragment count when larger). Each task takes a contiguous run of fragments, and the runs are chosen so the heaviest task is as light as any contiguous split allows. Defaults to `min(fragment_count, spark.default.parallelism)`. |
+
+Contiguous coverage matters beyond parallelism: [OPTIMIZE](./optimize.md) can only group fragments
```
geruh, line 58: I'm a little confused by how strong we're making this claim. contiguity fixes the interleaving problem, but isn't the segment boundary itself still a compaction boundary?

I ran some tests on this . 4 frags w/ `num_segments` set to 4 gives coverage of [0] [1] [2] [3] and the fragment count stayed 4. with `num_segments = 2` it was `[0, 1] & [2, 3]` and it went from `4 -> 2`/ 

if that's right, can we say this more like we're removing the extra boundaries caused by interleaving, rather than saying contiguous coverage solves the compaction problem?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -754,35 +768,139 @@ object IndexUtils extends Logging {
...
 
-    val segmentOrdering: Ordering[SegmentBatch] =
-      Ordering
-        .by[SegmentBatch, (Long, Int, Int)](segment =>
-          (segment.numRows, segment.fragmentIds.size, segment.index))
-        .reverse
+    if (runCount < segmentCount) {
+      // Heaviest splittable run first, so each extra batch is spent where it helps most. Ties break
+      // on length then on the earlier run, to keep the result independent of heap internals.
+      val splittable = PriorityQueue.empty[(Long, Int, Int)](
```
geruh: What is the heap buying us after we've already found the optimal max budget.

if a run is already <= that budget, splitting it can't make either side exceed it. so couldn't we just split runs until we have `segmentCount` and keep the same `minimax` guarantee?

i brute forced a simpler version against every contiguous split for lengths `1-6`, row counts `0-6`, every k, including zeros. it matched the optimal max every time.

if the heap is preserving some second property we actually care about, can we name/test that? otherwise this feels like more machinery than we need.

geruh (commented):
Hey @ivscheianu, thanks for raising! I did a high level pass let me know what you think

fangbo (approved):
Great optimization !

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
ivscheianu: yes, reworded to "optimal among contiguous splits" in the comment and the docs 👍🏻

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
ivscheianu: dropped it. Once the budget is fixed every run already fits, so both halves of any split fit too and the order can't change the heaviest batch.

docs/src/operations/ddl/create-index.md
ivscheianu: Docs now say contiguity removes the extra boundaries interleaving adds, and that num_segments bounds how far OPTIMIZE can get. Thank you!

geruh (approved):
LGTM, already got another approval will go ahead and merge!

PR #793, opened 2026-08-26
fix: name the index on uncommitted segment builds

ivscheianu:
Part of #789.

## What

Both uncommitted segment builders set `.replace(false)` and never call `.withIndexName(...)`:

- `ScalarSegmentIndexTask.execute()`
- `RangeBTreeIndexBuilder.buildForFragmentGroup()`

With no name, Lance derives its own default `{column}_idx` and then runs a name-collision pre-check **before any build work**, even on the uncommitted path (`rust/lance/src/index/create.rs`, `if !existing_named_indices.is_empty() && !self.replace`). So whenever an index named `{column}_idx` already exists on that column, every executor task fails with:

```
LanceError(Index): Index name 'id_idx' already exists, please specify a different name or use replace=True
```

`{column}_idx` is the name Lance itself assigns an unnamed index, so it is a common one to hit. Re-running `CREATE INDEX id_idx USING zonemap (id)` fails today.

## Change

Name the segment after the index it will join, and set `replace`. On the uncommitted path `replace` is consulted **only** by that pre-check — the removal logic lives in `execute()`, which the uncommitted path bypasses — so this is inert apart from suppressing a collision that does not apply. The driver's single `commitExistingIndexSegments` transaction still decides which existing segments to keep.

Naming the segment explicitly has a second benefit: the genuine "already exists with different fields" check now runs against the real index name, so a `CREATE INDEX name_idx USING btree (id)` against an existing `name_idx` on a different column fails before the distributed build rather than after it.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -443,8 +443,8 @@ case class RangeBTreeIndexBuilder(
       // Build an uncommitted BTree segment for this fragment group from the
       // pre-sorted data. No UUID is set: Lance generates the segment UUID, and
       // the fragment ids declare the segment's coverage so the per-partition
-      // segments stay disjoint. See ScalarSegmentIndexTask for why the segment
-      // build names the index and sets replace.
+      // segments stay disjoint. Named and replace-flagged to suppress Lance's
+      // collision pre-check against the existing index.
       val btreeParamsBuilder = BTreeIndexParams.builder()
       if (zoneSize.isDefined) {
         btreeParamsBuilder.zoneSize(zoneSize.get)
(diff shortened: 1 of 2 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -438,9 +441,10 @@ case class RangeBTreeIndexBuilder(
       Data.exportArrayStream(allocator, reader, stream)
 
       // Build an uncommitted BTree segment for this fragment group from the
-      // pre-sorted data. No index name or UUID is set: Lance generates the
-      // segment UUID, and the fragment ids declare the segment's coverage so
-      // the per-partition segments stay disjoint.
+      // pre-sorted data. No UUID is set: Lance generates the segment UUID, and
```
geruh: nit: can we clean this comment up while we are here we don't have to leave the story there. wdyt just something like:

`// replace is for Lance's name check, and the driver commit still publishes.`

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -519,10 +525,13 @@ class ScalarSegmentIndexJob(
 final private[v2] case class FragmentWorkload(fragmentId: Integer, numRows: Long)
 
 /**
- * A task to create a scalar index segment on a batch of fragments.
+ * A task to create a scalar index segment on a batch of fragments. Named after the logical index
```
geruh: nit: we can drop the change here. the comment is enough for me to follow

geruh (commented):
Thanks @ivscheianu, tested locally, and logic looks good! Just left a few nits lmk!

ivscheianu:
> Thanks @ivscheianu, tested locally, and logic looks good! Just left a few nits lmk!

Thank you, @geruh! Ready for a second round when you get the chance 👍🏻

geruh (approved):
LGTM

PR #794, opened 2026-08-26
feat: report index coverage, segment count and size in SHOW INDEXES

ivscheianu:
Part of #789.

## What

`SHOW INDEXES` reports indexed and unindexed fragment and row counts, but leaves the operator to compute coverage by hand and says nothing about how the index is laid out or what it costs. This adds three columns:

| Column | Type | Description |
|---|---|---|
| `indexed_percent` | double | Share of rows covered, truncated to two decimals so it never overstates. Null for an empty table, where coverage is undefined rather than complete. |
| `num_segments` | long | Physical segments backing the logical index. |
| `size_bytes` | long | Total across the segments' files. Null if any segment predates index file size tracking, since a partial sum would understate rather than admit it is unknown. |

All three come from metadata the command already fetches, so there is no extra driver cost. Truncating rather than rounding matters: a table one row short of full coverage reads as 99.99, not 100.

## Change

`indexed_percent` is derived from the row counts already read. `size_bytes` needs the whole segment list rather than its first element, so the grouping keeps it instead of calling `.head` immediately; output is still one row per logical index. `num_segments` reads the statistics blob, falling back to `num_indices` for older cores that report only that key.

The docs note two things an operator needs and would otherwise guess wrong:

- Partial coverage is not merely slower. A partially covered `zonemap` index prunes the fragments it does not cover, so a predicate on the indexed column can return fewer rows than the table holds, while `COUNT(*)` over the table stays correct. It comes from the connector's own driver-side pruning, which builds its surviving fragment set purely from the fragments the committed zones name, with no gate on coverage. Reproducible with `CREATE INDEX` plus an append, so it predates this change, but this column is what makes it visible. `btree`, `bitmap` and `bloomfilter` return complete results while partially covered.

  > **Edit:** fixed by #781.
- A vector index is listed here too, but Spark SQL can neither create nor refresh one; it is maintained through the Lance SDK.

### DIFF
--- a/docs/src/operations/ddl/show-indexes.md
+++ b/docs/src/operations/ddl/show-indexes.md
@@ -7,7 +7,7 @@ List all indexes defined on a Lance table.
 
 ## Overview
 
-The `SHOW INDEXES` command returns one row for each index on a Lance table. The information is retrieved using the `Dataset.describeIndices` method, and the output columns align with the attributes of `org.lance.index.IndexDescription`, excluding the per-segment metadata list.
+The `SHOW INDEXES` command returns one row for each index on a Lance table. The information is retrieved using the `Dataset.describeIndices` method, and the output columns align with the attributes of `org.lance.index.IndexDescription`. Per-segment metadata is not listed individually; `num_segments` and `size_bytes` summarise it.
 
 This command is useful for inspecting existing indexes, verifying index creation, and understanding the high-level properties of each index.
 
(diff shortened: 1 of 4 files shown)


ivscheianu:
Hello, @geruh! I’d appreciate it if you could also handle this one when you find some time. It'll allow me to work on the final PR that wires everything together. Thanks a lot!

docs/src/operations/ddl/show-indexes.md
```diff
@@ -54,11 +54,26 @@ The `SHOW INDEXES` command returns the following columns:
...
 | `num_unindexed_fragments` | long        | Number of fragments that are not yet indexed.                      |
 | `num_unindexed_rows`    | long          | Approximate number of rows that are not yet covered by the index.  |
+| `indexed_percent`       | double        | Share of rows the index covers, as a percentage truncated to two decimals, so it never overstates coverage. Null for an empty table. |
+| `num_segments`          | long          | Number of physical index segments backing this logical index.       |
+| `size_bytes`            | long          | Total size of all index files across the segments. Null if any segment predates index file size tracking. |
+
+## Interpreting the Output
+
+An `indexed_percent` below 100 means part of the table is not covered — either rows were appended
+since the index was last built, or [OPTIMIZE](./optimize.md) rewrote fragments an index had
```
geruh, line 64: question: this is optimize with out the remapping flag set right

geruh (approved):
LGTM thanks @ivscheianu

docs/src/operations/ddl/show-indexes.md
ivscheianu: yes, correct.  Will tighten the wording in the next PR

PR #795, opened 2026-08-26
fix: make distributed index builds version-consistent and report real coverage

ivscheianu:
Part of #789.

## What

Three problems in the distributed `CREATE INDEX` path, plus the driver-side cost of enumerating fragments.

**Tasks did not agree on a version.** Each task opens the dataset itself, and nothing pinned the version it opened at, so every task resolved the latest independently. Which fragments a segment covers is fixed by the driver's batch either way, but the version each segment records is not, and that version is what core checks against.

**Range mode read one snapshot and stamped another.** Its coverage comes from the fragment ids in the scanned rows, and the scan resolves its own version through the catalog, so a fragment appended after planning could be declared covered by a segment stamped with a version that predates it. Core skips staleness validation for a fragment absent at the stamped version, because the segment commit path passes `prune_historically_missing = false`. Those keys were then trusted however they changed afterwards, and a predicate on the indexed column could miss rows a full scan returned.

**The command reported the fragment count it planned, not the one it achieved.** Commit intersects each segment's declared coverage with the live fragments, so a fragment retired while the build ran contributes nothing, yet `fragments_indexed` still counted it.

## Change

Both distributed build paths now run against the version the driver planned over. The segmented path hands the pinned options to every task. Range mode pins its producer scan as well, so the rows a segment holds and the version it records describe one snapshot; a fragment appended after planning is simply not part of the build, which was already true of the segmented path.

Read options that name no version on main are refused rather than quietly falling back to an unpinned scan. That fallback is invisible, and it is the exact failure the pin exists to prevent.

`commitIndexSegments` computes the coverage the commit will actually establish, reports that, and names anything retired in a warning.

Retirement alone does not discard the build. A fragment leaves the manifest either because its rows moved (compaction, an in-place rewrite) or because they were all deleted; only the first leaves data unindexed, and either way the segments covering what remains are correct. Throwing away a finished distributed build over that would make a routine concurrent `DELETE` fatal on a long one. Only a total loss is refused, and not because anything would be corrupted: an existing segment is trivially disjoint from an empty bitmap and survives. The transaction would just publish an index over no data and report a build that achieved nothing, and this is the last point where it can be declined.

Fragment enumeration goes through `Dataset.getFragmentStatistics()`, which returns primitive arrays, instead of `getFragments()`, which materializes a Java object per fragment and per data file. The driver enumerates twice per command, so on a large table that is the bulk of planning cost.

## Not closed

Staleness is narrowed rather than eliminated. The commit's rebase window still waves `Operation::Merge` through, which core documents can rewrite a column in place. Nothing in this connector's DML reaches it, since `UPDATE COLUMNS` commits `Operation::Update` and is pruned.

Worth raising upstream separately: the segment commit path validates coverage with `prune_historically_missing = false` while the merge path passes `true`, so a caller that hands core a bitmap naming a fragment absent at the segment's own version gets it committed unvalidated. The connector can avoid producing that state, but it cannot detect it.

### DIFF
--- a/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
+++ b/lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
@@ -88,14 +88,7 @@ case class AddIndexExec(
     val btreeBuildMode = IndexUtils.btreeBuildMode(indexType, args)
     val scalarSegmentIndexType = IndexUtils.scalarSegmentIndexType(method)
 
-    // Plan and build against a single pinned version. Tasks open the dataset themselves, so without
-    // pinning each one resolves the latest version independently. The driver's batches fix which
-    // fragments a segment covers either way, but not the version each segment records, and that
-    // version is what core checks against. It revalidates a segment's coverage of a fragment only
-    // when the recorded version predates the commit and the fragment already existed at it, so a
-    // task opening after a concurrent rewrite of the indexed column stamps the current version over
-    // keys it read earlier and those keys are trusted unchecked. Coverage the commit cannot
-    // establish is accounted for at commit time; see IndexUtils.establishedCoverage.
+    // Plan and build at one pinned version; the commit opens the live dataset.
     val (fragmentWorkloads, canonicalColumns, buildReadOptions) = {
       val ds = Utils.openDatasetBuilder(readOptions).build()
       try {
@@ -170,11 +163,6 @@ case class AddIndexExec(
     val (nsImpl, nsProps, tableId, initialStorageOpts) =
       extractNamespaceInfo(lanceDataset, readOptions)
 
-    // Range-mode BTree uses preprocessed data from Spark and keeps its dedicated path: its coverage
-    // follows the fragment ids in the scanned rows rather than the fragment list planned above. The
-    // pinned options drive both halves of that path, the producer scan and the builder, so the rows a
-    // segment holds and the version it is stamped with describe the same snapshot; the job explains
-    // why both are needed.
     if (btreeBuildMode.contains("range")) {
       val segments = new RangeBasedBTreeIndexJob(
         this.copy(columns = canonicalColumns),
@@ -302,13 +290,6 @@ case class AddIndexExec(
  * covering exactly those fragments, so the resulting segments have disjoint fragment coverage and
  * can be committed directly as a single logical index.
  *
- * Unlike the segmented path, coverage here is derived from the fragment ids present in the scanned
- * rows rather than from a fragment list fixed at planning time, so the producer scan is pinned to the
- * same version as the build. Both halves have to describe one snapshot: the segment records the
- * version the builder opened, and core validates a segment's coverage against that version only for
- * a fragment that existed at it. A scan left to resolve its own version could hand the builder rows
- * from a fragment appended afterwards, whose keys would then be committed unvalidated.
- *
  * @param addIndexExec       The AddIndexExec instance that initiated this job
  * @param readOptions        Configuration options for reading the Lance dataset
  * @param numFragments       Number of fragments in the dataset, used to bound shuffle partitions
@@ -328,16 +309,7 @@ class RangeBasedBTreeIndexJob(
 
   private val VALUE_COLUMN_NAME = "value"
 
-  /**
-   * The version [[readOptions]] is pinned to.
-   *
-   * The caller pins these options with [[IndexUtils.pinVersion]], and CREATE INDEX only runs against
-   * a writable table, so the ref names a version on main; a branch or tag target is rejected before
-   * this point by `LanceDataset.ensureWritable`. Anything else cannot be expressed as a scan option
-   * at all, since branch and version are mutually exclusive there, so it would leave the scan
-   * unpinned and reading a different snapshot than the build stamps. That is the defect this pin
-   * exists to prevent, and it is silent, so refuse rather than fall through to it.
-   */
+  /** Version `readOptions` is pinned to. Throws if the ref is not a version on main. */
   private def pinnedVersion: Long = {
     val ref = readOptions.getRef
     if (ref == null || !ref.isMain || !ref.getVersionNumber.isPresent) {
@@ -369,15 +341,6 @@ class RangeBasedBTreeIndexJob(
     }
     val fullTableName = parts.mkString(".")
 
-    // Read the indexed column with the row id and fragment id metadata columns, at the version the
-    // build is pinned to. The scan otherwise resolves its own version through the catalog, which is
-    // at or after the pinned one, and coverage here comes from the fragment ids in the scanned rows:
-    // a fragment appended after planning would be declared covered by a segment stamped with a
-    // version that predates it, and core skips staleness validation for a fragment absent at the
-    // stamped version, so its keys would be trusted however they changed afterwards. Pinning the
-    // scan keeps the rows and the stamp describing one snapshot. Fragments appended after planning
-    // are simply not part of this build, exactly as in the segmented path, whose fragment list is
-    // fixed at planning time too; Dataset.optimizeIndices covers them incrementally.
     val fragmentColumn = LanceDataset.FRAGMENT_ID_COLUMN.name
     val df = session.read
       .option(LanceSparkReadOptions.CONFIG_VERSION, pinnedVersion)
@@ -682,13 +645,7 @@ object IndexUtils extends Logging {
       .get(method.toLowerCase(Locale.ROOT))
       .filter(scalarSegmentIndexTypes.contains)
 
-  /**
-   * Pins `readOptions` to the version `dataset` is open at.
-   *
-   * Distributed index builds hand read options to tasks that open the dataset themselves. Pinning
-   * makes every task observe the fragment set the driver planned over instead of resolving the
-   * latest version independently.
-   */
+  /** Pins `readOptions` to the version `dataset` is open at. */
   def pinVersion(
       readOptions: LanceSparkReadOptions,
       dataset: Dataset): LanceSparkReadOptions =
@@ -722,21 +679,10 @@ object IndexUtils extends Logging {
       .toSet
 
   /**
-   * Refuses to publish a segment set that would establish no coverage at all.
-   *
-   * Commit intersects each segment's declared coverage with the dataset's live fragments, so a
-   * fragment retired while the build ran contributes nothing. Lance accepts such a set: an empty
-   * fragment bitmap is valid metadata, and an existing segment is trivially disjoint from it and so
-   * survives. Nothing is corrupted, but the transaction would publish segments that index no data and
-   * report a build that achieved nothing, and this is the last point at which it can still be
-   * declined.
-   *
-   * Partial loss is not refused. A fragment leaves the manifest either because its rows moved
-   * (compaction, an in-place rewrite) or because they were all deleted; only the first leaves data
-   * unindexed, and either way the segments covering what remains are correct. Discarding a finished
-   * distributed build over that would be the wrong response, and it would make a routine concurrent
-   * DELETE fatal on a long one. What the commit actually established is reported afterwards, by
-   * [[establishedCoverage]].
+   * Refuses a segment set that declares coverage but intersects no live fragment. Partial loss is
+   * accepted: fragments can be retired by compaction or deletion during the build, and the segments
+   * covering what remains are still correct. [[establishedCoverage]] reports what was actually
+   * committed.
    */
   def requireCommittableCoverage(
       liveFragmentIds: Set[Int],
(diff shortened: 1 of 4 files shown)


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -163,27 +169,31 @@ case class AddIndexExec(
...
+    // Range-mode BTree uses preprocessed data from Spark and keeps its dedicated path: its coverage
+    // follows the fragment ids in the scanned rows rather than the fragment list planned above. The
+    // build is pinned like the segmented one all the same. The scan resolves its own version through
+    // the catalog, so an unpinned executor can open a version newer than the rows it was handed and
+    // stamp the segment with it, and core only validates segments stamped older than the commit.
     if (btreeBuildMode.contains("range")) {
       val segments = new RangeBasedBTreeIndexJob(
         this.copy(columns = canonicalColumns),
-        readOptions,
+        buildReadOptions,
```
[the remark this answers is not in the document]
geruh: I agree with this and still can repro:

`session.table` is live. the builder is pinned. that's two snapshots.

pin v1, insert fragment F, scan F, rewrite F, build at v1. WHERE id = 1001 returns 0. same split as this job.

can we pin the scan with `VERSION AS OF` the planning version too? commit can stay live.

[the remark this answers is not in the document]
ivscheianu: Yes, you're right, pinned the scan too. Thanks!

ivscheianu:
hello @geruh, thanks a lot for the review! I'd appreciate if you can take another look when you find some time, thanks a lot!

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -316,9 +369,19 @@ class RangeBasedBTreeIndexJob(
...
+    // at or after the pinned one, and coverage here comes from the fragment ids in the scanned rows:
+    // a fragment appended after planning would be declared covered by a segment stamped with a
+    // version that predates it, and core skips staleness validation for a fragment absent at the
+    // stamped version, so its keys would be trusted however they changed afterwards. Pinning the
+    // scan keeps the rows and the stamp describing one snapshot. Fragments appended after planning
+    // are simply not part of this build, exactly as in the segmented path, whose fragment list is
+    // fixed at planning time too; Dataset.optimizeIndices covers them incrementally.
     val fragmentColumn = LanceDataset.FRAGMENT_ID_COLUMN.name
-    val df = session.table(fullTableName)
+    val df = session.read
```
geruh, line 345: pin now sense here now. Can we add a test for this similar to the gatekeeper repro?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -619,6 +682,115 @@ object IndexUtils extends Logging {
       .get(method.toLowerCase(Locale.ROOT))
       .filter(scalarSegmentIndexTypes.contains)
 
+  /**
+   * Pins `readOptions` to the version `dataset` is open at.
```
geruh: this is backwards now. We can probably drop the latter half of this comment

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -88,18 +88,25 @@ case class AddIndexExec(
     val btreeBuildMode = IndexUtils.btreeBuildMode(indexType, args)
     val scalarSegmentIndexType = IndexUtils.scalarSegmentIndexType(method)
 
-    val (fragmentWorkloads, canonicalColumns) = {
+    // Plan and build against a single pinned version. Tasks open the dataset themselves, so without
```
geruh: nit: this comment talks about the old behavior. Let's drop or at the very leas say something like:

`// plan and build at one version, and commit stays on the live dataset.`

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -163,27 +170,31 @@ case class AddIndexExec(
     val (nsImpl, nsProps, tableId, initialStorageOpts) =
       extractNamespaceInfo(lanceDataset, readOptions)
 
-    // Range-mode BTree uses preprocessed data from Spark and keeps its dedicated path.
+    // Range-mode BTree uses preprocessed data from Spark and keeps its dedicated path: its coverage
+    // follows the fragment ids in the scanned rows rather than the fragment list planned above. The
+    // pinned options drive both halves of that path, the producer scan and the builder, so the rows a
+    // segment holds and the version it is stamped with describe the same snapshot; the job explains
```
geruh: nit: similar to the run comment we can drop

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -276,6 +302,13 @@ case class AddIndexExec(
  * covering exactly those fragments, so the resulting segments have disjoint fragment coverage and
  * can be committed directly as a single logical index.
  *
+ * Unlike the segmented path, coverage here is derived from the fragment ids present in the scanned
```
geruh: nit: we can drop this section since this is the old failure mode

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -295,6 +328,26 @@ class RangeBasedBTreeIndexJob(
 
   private val VALUE_COLUMN_NAME = "value"
 
+  /**
+   * The version [[readOptions]] is pinned to.
+   *
```
geruh: nit: i think at the very least we could say:

```/** Version `readOptions` is pinned to. Throws if the ref is not a version on main. */```

File: lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala [hunk 8]
```diff
@@ -316,9 +369,19 @@ class RangeBasedBTreeIndexJob(
     }
     val fullTableName = parts.mkString(".")
 
-    // Read the indexed column with the row id and fragment id metadata columns.
+    // Read the indexed column with the row id and fragment id metadata columns, at the version the
```
geruh: nit: we can delete this block

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -619,6 +682,115 @@ object IndexUtils extends Logging {
...
+
+  /** Fragment ids the given segments declare coverage of. */
+  def declaredCoverage(segments: Seq[Index]): Set[Int] =
+    segments.iterator
+      .flatMap(_.fragments().orElse(Collections.emptyList[Integer]()).asScala)
+      .map(_.intValue)
+      .toSet
+
+  /**
+   * Refuses to publish a segment set that would establish no coverage at all.
```
geruh: nit: also telling the pr story, I think here we could keep refuse if nothing, and the partial loss portion

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
```diff
@@ -619,6 +682,115 @@ object IndexUtils extends Logging {
...
+    val declared = declaredCoverage(segments)
+    if (declared.nonEmpty && declared.intersect(liveFragmentIds).isEmpty) {
+      throw new IllegalStateException(
+        s"Index '$indexName' build raced a concurrent operation: every fragment it covers " +
+          s"(${describeFragmentIds(declared)}) was retired while the build ran, so the segments " +
+          "would cover nothing. No index change was committed; re-run the command.")
+    }
+  }
+
+  /**
```
geruh, line 700: nit: name kind of implies this we can say this builds committed frag ids, and instersected with what is still live or something along those lines.

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
```diff
@@ -105,6 +108,75 @@ public void tearDown() throws IOException {
     }
   }
 
+  /**
+   * Pins the two Lance behaviours the coverage report rests on: a segment commit returns the
```
geruh: nit: java doc on the test. test name is essentially the comment.

lance-spark-base_2.12/src/test/scala/org/apache/spark/sql/execution/datasources/v2/IndexUtilsTest.scala
```diff
@@ -257,4 +276,131 @@ class IndexUtilsTest {
       classOf[ArithmeticException],
       () => IndexUtils.batchFragments(fragmentWorkloads(Long.MaxValue, 1), Some(1), 1))
   }
+
+  // ── declaredCoverage / committedCoverage ───────────────────────
```
geruh: nit: do we need the section headers in these files?

lance-spark-base_2.12/src/test/scala/org/apache/spark/sql/execution/datasources/v2/IndexUtilsTest.scala
```diff
@@ -257,4 +276,131 @@ class IndexUtilsTest {
...
+  def establishedCoverage_reportsWhatTheCommitReturned(): Unit = {
+    val built = Seq(coveringSegment(0, 1), coveringSegment(2))
+
+    assertEquals(
+      Set(0, 1, 2),
+      IndexUtils.establishedCoverage(built, built, Set(0, 1, 2), "idx_id"))
+  }
+
+  /**
+   * Lance prunes a fragment whose indexed field was rewritten under the same id, so the committed
```
geruh: nit: test should imply this comment

lance-spark-base_2.12/src/test/java/org/lance/spark/branch/BaseBranchDDLTest.java
```diff
@@ -590,6 +590,28 @@ public void testBranchAndVersionOptionsFail() {
     Assertions.assertTrue(conflictMessages.contains("version"));
   }
 
+  /**
+   * The version option has to pin the scan itself, not merely be accepted. Distributed index builds
```
geruh: nit: we can drop the java doc here since the test name & logic implies behavior

geruh (commented):
Thanks @ivscheianu! Did a closer pass here, and the logic looks great. Just left some nits.

lance-spark-base_2.12/src/test/scala/org/apache/spark/sql/execution/datasources/v2/IndexUtilsTest.scala
ivscheianu: not sure, I see that's the pattern on the test files, e.g.:
```
// ── extractTrain ──────────────────
```

```
// ── toJson — SparkOnlyOptions filtering ───────
```
but happy to remove 👍🏻

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/AddIndexExec.scala
ivscheianu: The mechanism is tested by `testVersionOptionPinsTheScanToThatVersion` in BaseBranchDDLTest, same API call this path uses.

I tried to think of a targeted test here but the false-negative scenario needs a concurrent rewrite landing between the scan and the build, and I don't think I can control that timing from a single CREATE INDEX statement without something flaky. Once https://github.com/lance-format/lance-spark/pull/794 lands a coverage assertion might work, but right now I'm not sure what to assert that the mechanism test doesn't already cover.

Open to ideas if you see something I missed!

ivscheianu:
Thanks a lot @geruh for the review, really appreciate it! Addressed the comments, please have a look when you get some time. Since I got your attention, I'd appreciate if you can also have a look on the follow-ups, they're small 😄 
https://github.com/lance-format/lance-spark/pull/793
https://github.com/lance-format/lance-spark/pull/794

geruh (approved):
LGTM thanks @ivscheianu!

Issue #541, opened 2026-05-18
Indexed nearest-neighbor join optimization for lance-spark (tracking issue, 7-PR split)


sezruby:
## Summary

Add an indexed nearest-neighbor join optimization to `lance-spark`: route SQL of the shape

```sql
SELECT *
FROM   queries q
LEFT OUTER JOIN documents d
  APPROX NEAREST 10 BY DISTANCE vector_l2_distance(q.vec, d.vec)
```

through a per-fragment Lance index probe + Spark-side merge instead of the default
`O(|L| × |R|)` cross product. Targets workloads where the right side is a Lance
dataset with a vector index (IVF-PQ / HNSW); falls back to the existing Spark plan when
those conditions don't hold.

PoC PR (against my fork's `main`, ~1 week behind upstream — same diff a draft PR
against `lance-format/lance-spark:main` would show):
**https://github.com/sezruby/lance-spark/pull/1**

Headline numbers from the PoC (full results in [`BENCHMARK_RESULTS.md`](https://github.com/sezruby/lance-spark/blob/knn-phase0/lance-spark-knn_2.12/BENCHMARK_RESULTS.md)):

- DataFrame API vs. naive `crossJoin + UDF + window`: **~600×** (small scale, oracle-validated)
- SQL with Spark 4.2 `NearestByJoin` rule ON vs. OFF (apples-to-apples): **~17×**
- Cohere wiki, 1.8M rows × dim 1024, on a real cluster: **~160×** (median across 7 iterations, range 100–200×)

## Design (TL;DR)

Lance vector indexes are fragment-local, so per-fragment probes are independent
and trivially parallel. The runtime path is:

```
LanceProbeExec        ← per-task: open Lance, run nearest search, emit (leftId, rowId, dist)
   │
   ▼ ShuffleExchangeExec (Catalyst-inserted, hash-partitioned on leftId)
   │
LanceMergeExec        ← per-leftId top-K heap merge across probe outputs
   │
   ▼
LanceMaterializeExec  ← point-fetch the chosen rowIds back from Lance
```

All three are real Catalyst `SparkPlan` nodes — visible in `df.explain()`,
AQE engages on the merge shuffle, ColumnPruning / projection fusion all work.
No public API change for end users once the rule lands: registering one
Spark session extension is the entire opt-in.

Full design (problem, alternatives, why each piece is shaped the way it is, the
ColumnPruning gotcha, the Catalyst-vs-Kryo payload analysis, fragment-grouping
[...]

PR #796, opened 2026-08-26
feat: indexed nearest-by join over Lance (Spark 4.2 SQL)

sezruby:
## What

Accelerates Spark 4.2's native `APPROX NEAREST k BY DISTANCE ...` join (SPARK-56395) over Lance tables. The default lowering turns `NearestByJoin` into an O(|L|×|R|) cross-product + `MaxMinByK`; this rule intercepts the operator and runs a per-left-row **native ANN/exact probe against the Lance vector index** instead — no shuffle, no broadcast.

Part of the effort tracked in #541. This is the narrower **Spark 4.2 SQL-interception slice** only; the manual `df.kNearestJoin` DataFrame API is intentionally deferred and can be added later as a version-agnostic base module.

## How

New **opt-in** module `lance-spark-knn-4.2_2.13`:

- **Catalyst `IndexedNearestByJoinRule`** (a `postHocResolutionRule`) recognizes `NearestByJoin` and rewrites it to a single no-shuffle logical node. It must run in postHoc, **not** the optimizer — Spark's `RewriteNearestByJoin` fires first in the `FinishAnalysis` batch, so an injected optimizer rule would only ever see the already-lowered cross-product.
- **`LanceKnnJoinStrategy` / `LanceKnnJoinExec`** lower to one `mapPartitions` → per-left-row native probe → bounded top-K heap → late-materialize by row address.
- **Probe core** (`LanceProbe` / `LanceKnnJoinStage` / `TopKHeap` / `Metric`) supports L2, cosine, and dot metrics with `nprobes` / `refineFactor` tuning.
- **Prefilter pushdown**: right-side `WHERE` is translated to a Lance filter string, including nested struct-field predicates rendered as dotted paths (`col.field`). Conservative by design — if any sub-expression is unsupported the whole rewrite is **refused** and Spark's brute-force lowering runs unchanged (never a silently dropped residual).

Off by default (`spark.lance.knn.indexedNearestByJoin.enabled=false`); wired via `LanceKnnSparkSessionExtensions`.

## Config

| key | default | meaning |
|---|---|---|
| `spark.lance.knn.indexedNearestByJoin.enabled` | `false` | master switch for the rule |
| `spark.lance.knn.nprobes` | (index default) | IVF probe count |
| `spark.lance.knn.refineFactor` | (unset) | over-fetch factor for re-ranking |

## Tests

- `IndexedNearestByJoinRuleTest` — pure Catalyst unit tests (rule matching, filter translation accept/refuse, nested struct-field paths); no Lance execution.
- `IndexedNearestByJoinSqlTest` — end-to-end SQL over real Lance tables: exact-path oracle equivalence, `WHERE` prefilter pushdown, rule-off fallthrough, and IVF-PQ recall (default / clustered / refineFactor) checks.
- `TopKHeapTest`, `LanceProbeValidationTest` — probe-core units.

> Note: the default CI matrix builds connector modules with `-am`, which pulls upstream deps only; this module is downstream of the connectors, so a reactor build (e.g. `./mvnw test -pl lance-spark-knn-4.2_2.13 -am`) is needed to exercise it.

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -85,6 +85,11 @@ filename = "lance-spark-4.2_2.13/pom.xml"
 search = "<version>{current_version}</version>"
 replace = "<version>{new_version}</version>"
 
+[[tool.bumpversion.files]]
+filename = "lance-spark-knn-4.2_2.13/pom.xml"
+search = "<version>{current_version}</version>"
+replace = "<version>{new_version}</version>"
+
 # Bundle module pom.xml files - parent version
 [[tool.bumpversion.files]]
 filename = "lance-spark-bundle-3.4_2.12/pom.xml"
--- a/docs/src/config.md
+++ b/docs/src/config.md
@@ -58,6 +58,27 @@ The following features require the Lance Spark SQL extension to be enabled:
 - [OPTIMIZE](operations/ddl/optimize.md) - Compact table fragments for improved query performance
 - [VACUUM](operations/ddl/vacuum.md) - Remove old versions and reclaim storage space
 
+### Indexed Nearest-Neighbor Join Extension
+
+The `APPROX NEAREST` join (Spark 4.2 `APPROX NEAREST ... BY DISTANCE` / `BY SIMILARITY` syntax) can
+be rewritten to probe the Lance vector index instead of running a brute-force cross product. This is
+a **separate** extension, packaged in the `lance-spark-knn-4.2` module, and requires Spark 4.2 or
+later. Enable it alongside (or instead of) the connector extension:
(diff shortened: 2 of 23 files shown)


sezruby:
For reviewability, this feature is being split into a stack of smaller PRs. This PR stays open as the full-feature reference; the slices land independently:

- **[1/3] Probe core primitives** — #797 (open)
- **[2/3]** no-shuffle per-partition join execution — to follow
- **[3/3]** Catalyst `APPROX NEAREST` interception rule + SQL/recall tests + docs — to follow

PR #799, opened 2026-08-26
feat(knn): indexed Lance vector-join execution stage (Spark 4.2) [2/3]

sezruby:
Part [2/3] of the indexed nearest-by vector join over Lance, split by architectural layer from the umbrella #796.

This slice adds the per-partition **execution stage** (`internal/LanceKnnJoinStage`) that drives the join, layered on the probe-core primitives from **#797** ([1/3]).

## Stacked on #797
This PR is stacked on #797. Until #797 merges to `main`, this diff also shows the [1/3] probe-core files (`Metric`, `TopKHeap`, `ScoredRowRef`, `LanceProbe`) because `main` does not yet contain them; once #797 merges, this diff reduces to the join-stage layer alone. **Review target here:** `internal/LanceKnnJoinStage.scala` and `internal/LanceKnnJoinStageTest.scala`.

## What this stage does
- `runPartition`: open one `LanceProbe` per partition, stream the join output lazily (no per-partition buffering), close the native handle on task completion (success or failure).
- Fused vs split routing (`foldsInOneScan`): with no over-fetch and a payload projection that does not collide with an injected column (`_rowid` / `_distance` / `_score`), search and project in ONE native scan; otherwise probe → trim → late-materialize survivors by `_rowid`.
- `resolveReadContext` / `mergeReadOptions`: pin the Lance snapshot once on the driver so every task probes one consistent version; mirror the connector's incompatible-pinned-ref guard.
- `coerceToSpark`: schema-aware payload coercion (structs, arrays, and the Arrow entry-list shape a `MapType` cell arrives in) to the shape the join's encoder accepts.

## Tests
`LanceKnnJoinStageTest` (backend-free) covers lazy output, fold-vs-split routing for reserved-name projections, read-option merge conflicts, and payload coercion. Full module `test` phase: 30 pass.

The SQL Catalyst rule, physical/logical plan, strategy, and session extension — plus their tests and docs — follow in [3/3].

### DIFF
--- a/.bumpversion.toml
+++ b/.bumpversion.toml
@@ -85,6 +85,11 @@ filename = "lance-spark-4.2_2.13/pom.xml"
 search = "<version>{current_version}</version>"
 replace = "<version>{new_version}</version>"
 
+[[tool.bumpversion.files]]
+filename = "lance-spark-knn-4.2_2.13/pom.xml"
+search = "<version>{current_version}</version>"
+replace = "<version>{new_version}</version>"
+
 # Bundle module pom.xml files - parent version
 [[tool.bumpversion.files]]
 filename = "lance-spark-bundle-3.4_2.12/pom.xml"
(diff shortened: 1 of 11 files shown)


sezruby:
Collapsed into #797. The join stage (`LanceKnnJoinStage` + `LanceKnnJoinStageTest`) now lands together with its only caller in the probe-core PR, since a probe primitive with no in-tree caller is hard to review in isolation. The over-fetch / trim / late-materialize split path this PR carried is removed: the native `probeRows` search returns the final top-k best-first with the payload already fused in, so there is no JVM-side merge, and the code that served the split (`TopKHeap`, `LanceProbe.probe`/`materialize`, the `ef` knob, `foldsInOneScan`) had no live caller. The Catalyst / SQL integration remains the [3/3] follow-up.

PR #802, opened 2026-09-03
refactor: drop the unreachable blob column vector duplicates

jackylee-ch:
## Summary
- Delete `vectorized/BlobPositionColumnVector` and `vectorized/BlobSizeColumnVector`, which nothing
  references. The live implementations are the private nested classes of the same name in
  `LanceFragmentColumnarBatchScanner` (:206, :306), which that file resolves without importing the
  `vectorized` copies

## Test plan
- [x] `make test` on Spark 4.1/Scala 2.13 and 3.5/Scala 2.12; `make lint`


geruh (approved):
LGMT thanks @jackylee-ch!

PR #803, opened 2026-09-03
fix: read uint64 columns without boxing through BigInteger

jackylee-ch:
## Summary
`UInt8Accessor.getLong` boxed through `UInt8Vector.getObjectNoOverflow`, which returns the unsigned
value as a `BigInteger`. `longValueExact()` then throws `ArithmeticException: BigInteger out of long
range` for every uint64 at or above 2^63, and NPEs on a null slot, instead of wrapping. The class
javadoc, `LanceArrowUtils.scala:102`, and the `_rowaddr` read at
`LanceFragmentColumnarBatchScanner.java:197` all treat uint64 as wrapping into `LongType`.

Reachable through any uint64 column, including `_rowid` and `_rowaddr`, which
`LanceArrowUtils.scala:538` maps to `ArrowType.Int(64, false)`.

## Test plan
- [x] New `LanceArrowColumnVectorTest` case: fails on the old accessor with `ArithmeticException`
- [x] `*Vector*`/`*Blob*`/`*Update*` suites on Spark 4.1/Scala 2.13 and 3.5/Scala 2.12; `make lint`


lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/UInt8Accessor.java
```diff
@@ -27,7 +27,11 @@ public class UInt8Accessor {
   }
 
   final long getLong(int rowId) {
-    return accessor.getObjectNoOverflow(rowId).longValueExact();
+    // Read the raw two's-complement long rather than boxing through getObjectNoOverflow(), which
```
geruh: nit: this is a bit verbose we can drop or trim to something like:

```
// Arrow get() throws on a null slot. Spark getLongs calls getLong without isNullAt.
```

wdyt?

lance-spark-base_2.12/src/test/java/org/lance/spark/vectorized/LanceArrowColumnVectorTest.java
```diff
@@ -25,13 +26,39 @@
...
 
 public class LanceArrowColumnVectorTest {
 
+  @Test
+  public void uint64ValuesAboveLongMaxValueWrapInsteadOfThrowing() {
+    Field field = Field.nullable("id", new ArrowType.Int(64, false));
+    try (BufferAllocator allocator = new RootAllocator();
+        UInt8Vector vector = (UInt8Vector) field.createVector(allocator)) {
+      vector.allocateNew(3);
+      // 2^63 and 2^64-1 are ordinary uint64 values. LongType carries them as their
```
geruh: nit: im fine with these comments but can drop wdyt?

lance-spark-base_2.12/src/main/java/org/lance/spark/vectorized/UInt8Accessor.java
jackylee-ch: Took your wording verbatim. The BigInteger rationale stays in the PR description, which becomes the squash body, so the reasoning is still in history.

lance-spark-base_2.12/src/test/java/org/lance/spark/vectorized/LanceArrowColumnVectorTest.java
jackylee-ch: Dropped it.

jackylee-ch:
cc @LuciferYang

geruh (approved):
LGTM, thanks for addressing feedback!

PR #804, opened 2026-09-03
fix: reuse and close executor namespace clients

hfutatzhanghb:
## Summary

- recreate the executor namespace lifecycle fix on top of the latest `main`
- create the executor-side namespace client once per `LanceColumnarPartitionReader` instead of once per fragment
- reuse the task-owned namespace across every fragment in the input partition
- close `AutoCloseable` namespace implementations after fragment resources on normal completion and scan/open failures
- give Arrow C stream exports and pushed `COUNT(*)` scans explicit namespace lifecycle owners
- add regression coverage for columnar reuse, Arrow/count cleanup, failure cleanup, and the disabled-refresh path

This PR replaces #728 and uses a new branch based on current `main`.

## Root cause

With `executor_credential_refresh=true`, `LanceFragmentScanner.create()` called
`LanceRuntime.getOrCreateNamespace()` before every fragment. Despite its name, that method
always creates a new `LanceNamespace` connection. The fragment scanner owned and closed its
native scanner and dataset, but nobody owned or closed the namespace client.

Hive2 and Hive3 namespace implementations are `Closeable` and own HMS client pools. As a
result, a Spark task scanning multiple fragments repeatedly created HMS-backed namespace
clients and leaked their connections.

## Fix

`LanceColumnarPartitionReader` now owns executor namespace initialization and cleanup because
its lifecycle matches one Spark task. Initialization remains lazy and preserves the existing
`executor_credential_refresh` and `useNamespaceOnWorkers` gates. The namespace stays attached
to the task-local read options while fragments are scanned, then is cleared and closed after
the active fragment scanner.

The failure path preserves the original exception and attaches cleanup failures as suppressed
exceptions. Repeated reader close calls remain idempotent.

## Validation

- `./mvnw spotless:check`
- `./mvnw test -pl lance-spark-3.5_2.13 -DskipTests` (main and test sources compile)
- `./mvnw test -pl lance-spark-3.5_2.13 -Dtest=org.lance.spark.internal.LanceFragmentScannerTest#testExecutorNamespaceOwnerClosesClient`


hfutatzhanghb:
@LuciferYang  @fangbo  Hi, Could you please help review this PR when have free time? Thanks very much !!!

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
```diff
@@ -49,6 +62,7 @@ public boolean next() throws IOException {
         fragmentReader = null;
         toClose.close();
       }
+      initializeExecutorNamespace();
```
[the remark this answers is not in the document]
hfutatzhanghb: Fixed in 703bafb. I added a shared task/handle-scoped ExecutorNamespace owner that preserves the existing refresh and worker gates. LanceArrowStreamScanner now acquires the namespace before opening the fragment and retains it in LanceArrowStream, which closes it after the stream, scanner, and dataset. LanceCountStarPartitionReader now uses the same owner in an outer try-with-resources scope, so its scanner/reader and dataset close before the namespace on success and failure. The ordinary columnar reader was switched to the same owner. Added regression coverage for Arrow initialization/close, count success/failure cleanup, and the shared owner lifecycle.

hfutatzhanghb:
Hi, Sir. @LuciferYang  @geruh  Could you please help review this PR when have free time? Thanks very much !!!

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceFragmentScannerTest.java
```diff
@@ -224,56 +231,141 @@ public void testGetColumnNamesWithFragmentId() throws Exception {
...
+      assertNotNull(partition.getReadOptions().getNamespace());
+      assertEquals(1, RecordingNamespace.INITIALIZE_CALLS.get());
+    }
+
+    assertNull(partition.getReadOptions().getNamespace());
+    assertEquals(1, RecordingNamespace.CLOSE_CALLS.get());
+  }
+
+  @Test
+  public void testPartitionReaderReusesAndClosesExecutorNamespace() throws Exception {
```
geruh: These are constructing the `LanceColumnarPartitionReader`.  Should they live in the `LanceColumnarPartitionReaderTest` file looks like they have a close() test

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceFragmentScannerTest.java
```diff
@@ -224,56 +231,141 @@ public void testGetColumnNamesWithFragmentId() throws Exception {
...
-   * gate, executors of Kerberized HMS catalogs hit {@code GSS initiate failed} because they lack a
-   * TGT for the eager {@code describeTable()} RPC.
-   *
-   * <p>Strategy: hand a real, loadable {@link LanceNamespace} impl to the partition. If the gate
-   * regresses (rebuild not skipped), {@code LanceNamespace.connect} would succeed via {@code
-   * Class.forName}, {@link RecordingNamespace#initialize} would run, and {@code
-   * readOptions.setNamespace(...)} would fire — all observable here. The bogus dataset URI lets the
-   * outer {@code Utils.openDatasetBuilder().build()} call fail predictably, since the gate runs
-   * <i>before</i> the dataset is opened. No real Lance dataset is required.
+   * executor opens a namespace-backed table with the flag disabled, the partition reader must
```
geruh: nit: nice drop on the test java doc. I think we can drop this whole comment in favor of the test since i can understan what we are doing from the method name and the two asserts which mention refresh=false must not connect.

wdyt?

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
```diff
@@ -98,22 +115,57 @@ public CustomTaskMetric[] currentMetricsValues() {
...
-    if (!currentScanStatsAdded) {
-      metricsTracker.addScanStats(fragmentReader.getScanStats());
-      currentScanStatsAdded = true;
+    failure = closeResources(failure);
+    if (failure != null) {
+      throw asIOException(failure);
     }
-    // Null-first so close() is idempotent (PartitionReader extends Closeable, whose contract
-    // requires it): a repeat call short-circuits rather than raising
-    // `ArrowArrayStream is already closed` from a second ArrowArrayStream.release().
```
geruh, line 110: I think we might want to keep the comment here since the null-first is what makes close() idempotent, and spark closable isn't the only reason. wdyt?

```
// Null-first so close() is idempotent. A repeat call must not raise ArrowArrayStream is already closed from a second release().
```

geruh (commented):
Thanks for this @hfutatzhanghb! Just a few nits

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceFragmentScannerTest.java
hfutatzhanghb: Thanks a lot. Very nice advice. I moved the three tests that construct LanceColumnarPartitionReader into LanceColumnarPartitionReaderTest alongside the existing close() coverage. I kept the direct ExecutorNamespace ownership test in place since it does not construct the partition reader and its RecordingNamespace fixture is also reused by LanceArrowStreamScannerTest.

lance-spark-base_2.12/src/test/java/org/lance/spark/internal/LanceFragmentScannerTest.java
hfutatzhanghb: Agreed. The method name and assertion messages already explain the expected behavior, so I removed the Javadoc.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceColumnarPartitionReader.java
hfutatzhanghb: Agreed. I restored the concise null-first comment above the resource reset so the idempotence requirement remains explicit.  Thanks very much!

hfutatzhanghb:
Hi, @geruh . Could you please review this PR again when have free time? Thanks very much!

geruh (approved):

hfutatzhanghb:
@geruh Thanks very much for reviewing and merging!

PR #808, opened 2026-09-04
fix: log dataset close failures in the blob reference resolver

jackylee-ch:
## Summary
- `BlobReferenceResolver.close` dropped close failures silently. Route them through the package's
  `CloseableUtil.closeQuietly`, which logs at WARN with the cause

## Test plan
- [x] blob/join suites on Spark 4.1/Scala 2.13 and 3.5/Scala 2.12; `make lint`


geruh (approved):
thanks @jackylee-ch! 🚀

PR #810, opened 2026-09-07
fix: release blob file handles on resolve error paths

jackylee-ch:
## Summary
- `resolveBatch` abandoned every `BlobFile` `takeBlobs` had returned when the count check or the
  null check threw, and the tail of the list when `read()` threw. `BlobFile` wraps a native handle
  with no cleaner, and Spark retries the task in the same JVM. Release all of them in a `finally`,
  and do the same in `resolve`

## Test plan
- [x] `*Blob*`/`*Join*` suites on Spark 4.1/Scala 2.13 and 3.5/Scala 2.12 (144 each); `make lint`

### DIFF
--- a/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
+++ b/lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
@@ -80,11 +80,15 @@ public class BlobReferenceResolver implements AutoCloseable {
     List<Long> rowAddresses = new ArrayList<>(1);
     rowAddresses.add(ref.getRowAddress());
     List<BlobFile> blobs = dataset.takeBlobs(rowAddresses, ref.getColumnName());
-    if (blobs.isEmpty()) {
-      return new byte[0];
-    }
-    try (BlobFile blob = blobs.get(0)) {
-      return blob.read();
+    try {
+      if (blobs.isEmpty()) {
+        return new byte[0];
+      }
+      return blobs.get(0).read();
+    } finally {
+      for (BlobFile blob : blobs) {
+        CloseableUtil.closeQuietly(blob);
+      }
     }
   }
 
@@ -137,31 +141,37 @@ public class BlobReferenceResolver implements AutoCloseable {
       List<Long> addresses = group.distinctAddresses; // requested order
       List<BlobFile> blobs = dataset.takeBlobs(addresses, group.columnName);
 
-      // takeBlobs must return exactly one BlobFile per requested address, in order. A mismatch
-      // means the selection hit deleted/null-descriptor rows, in which case positional mapping
-      // would skew and silently write the wrong bytes into the target table — fail loudly instead.
-      if (blobs.size() != addresses.size()) {
-        throw new IOException(
-            String.format(
-                "takeBlobs returned %d blobs for %d requested addresses (column=%s, dataset=%s); "
-                    + "cannot map results to rows",
-                blobs.size(), addresses.size(), group.columnName, group.datasetUri));
-      }
-
-      for (int i = 0; i < addresses.size(); i++) {
-        BlobFile blob = blobs.get(i);
-        if (blob == null) {
+      // Every handle takeBlobs returned is released in the finally below, including on the two
+      // throws: BlobFile wraps a native handle with no cleaner, so an abandoned one is only
+      // reclaimed when the JVM exits, and Spark retries the task in the same JVM.
+      try {
+        // takeBlobs must return exactly one BlobFile per requested address, in order. A mismatch
+        // means the selection hit deleted/null-descriptor rows, in which case positional mapping
+        // would skew and silently write the wrong bytes into the target table — fail loudly.
+        if (blobs.size() != addresses.size()) {
           throw new IOException(
               String.format(
-                  "takeBlobs returned a null blob for address %d (column=%s, dataset=%s)",
-                  addresses.get(i), group.columnName, group.datasetUri));
+                  "takeBlobs returned %d blobs for %d requested addresses (column=%s, dataset=%s); "
+                      + "cannot map results to rows",
+                  blobs.size(), addresses.size(), group.columnName, group.datasetUri));
         }
-        byte[] data;
-        try (BlobFile b = blob) {
-          data = b.read();
+
+        for (int i = 0; i < addresses.size(); i++) {
+          BlobFile blob = blobs.get(i);
+          if (blob == null) {
+            throw new IOException(
+                String.format(
+                    "takeBlobs returned a null blob for address %d (column=%s, dataset=%s)",
+                    addresses.get(i), group.columnName, group.datasetUri));
+          }
+          byte[] data = blob.read();
+          for (int vectorIndex : group.indicesByAddress.get(addresses.get(i))) {
+            resolved.put(vectorIndex, data);
+          }
         }
-        for (int vectorIndex : group.indicesByAddress.get(addresses.get(i))) {
-          resolved.put(vectorIndex, data);
+      } finally {
+        for (BlobFile blob : blobs) {
+          CloseableUtil.closeQuietly(blob);
         }
       }
     }


jackylee-ch:
cc @LuciferYang

LuciferYang:
Missed this ping, will check it out.

lance-spark-base_2.12/src/test/scala/org/lance/spark/utils/BlobReferenceResolverTest.scala
```diff
@@ -0,0 +1,191 @@
...
+      val error = assertThrows(
+        classOf[IOException],
+        () => resolver.resolveBatch(indices(refs), refs))
+      assertTrue(error.getMessage.contains("Not found"), error.getMessage)
+      resolver.assertReleased()
+    } finally resolver.close()
+  }
+
+  @Test
+  def singleReadFailureReleasesHandle(): Unit = withSource() { (uri, refs) =>
```
LuciferYang, line 183: This test doesn't actually pin the `resolve` change. `resolve` requests a single address, so `takeBlobs` returns one blob, and the old `try (BlobFile blob = blobs.get(0))` closed it just as well when `read()` threw: revert the try/finally to the old form and this test still passes. For a single-address request the old and new code are equivalent, so `resolve`'s change is close to behavior-neutral.

The case that would show the new code's value (takeBlobs returning more handles than requested, where the old code closed only `blobs.get(0)` and leaked the rest) can't be constructed from a single-address request. So this is really a smoke test for `resolve`; the release semantics are pinned by the `resolveBatch` tests. If `resolve` were broken, this test wouldn't catch it.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -76,15 +76,18 @@ public BlobReferenceResolver(Map<String, BlobSourceContext> sourceContexts) {
...
-      return new byte[0];
-    }
-    try (BlobFile blob = blobs.get(0)) {
-      return blob.read();
+    List<BlobFile> blobs = takeBlobs(ref.getDatasetUri(), rowAddresses, ref.getColumnName());
+    try {
+      if (blobs.isEmpty()) {
+        return new byte[0];
+      }
+      return blobs.get(0).read();
```
LuciferYang, line 86: `resolve` calls `blobs.get(0).read()` with no null check, unlike `resolveBatch`, which throws a descriptive IOException for a null blob. `takeBlobs` returns null for a null-descriptor row, so `resolve` throws a bare NPE where `resolveBatch` reports column/dataset context. This isn't a leak (the finally's `closeQuietly(null)` is a no-op) and the behavior is pre-existing; also `resolve`/`resolveIfNeeded` have no production caller today, only tests.

Since this PR already rewrote `resolve`'s body, adding a null guard matching `resolveBatch` would make the two paths report errors consistently. Optional, can be deferred.

lance-spark-base_2.12/src/test/scala/org/lance/spark/utils/BlobReferenceResolverTest.scala
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  @Test
+  def batchReadFailureReleasesAllHandles(): Unit = withSource() { (uri, refs) =>
+    val resolver = new RecordingResolver(_ => deleteDataFiles(uri))
+    try {
+      val error = assertThrows(
+        classOf[IOException],
+        () => resolver.resolveBatch(indices(refs), refs))
+      assertTrue(error.getMessage.contains("Not found"), error.getMessage)
```
LuciferYang, line 177: The two read-failure tests (this one and line 187 in `singleReadFailureReleasesHandle`) force `read()` to fail by deleting the data files, then assert `getMessage.contains("Not found")`. That leans on two lance-core behaviors: the read must fail after `takeBlobs` returns (which assumes Lance reads lazily and holds no fd / no mmap at take time), and the error text stays exactly "Not found". A lance-core message change, or a platform where the file is opened at take time, turns these red.

The direction is fail-fast (the test breaks loudly, it won't miss a real leak), so it's not a merge blocker. Asserting only the `IOException` type rather than the literal text would be more robust.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -133,41 +136,50 @@ public Map<Integer, byte[]> resolveBatch(List<Integer> indices, List<BlobReferen
...
+          }
+          byte[] data = blob.read();
+          for (int vectorIndex : group.indicesByAddress.get(addresses.get(i))) {
+            resolved.put(vectorIndex, data);
+          }
         }
-        for (int vectorIndex : group.indicesByAddress.get(addresses.get(i))) {
-          resolved.put(vectorIndex, data);
+      } finally {
+        for (BlobFile blob : blobs) {
```
LuciferYang, line 171: The blob-release loop is identical in `resolve` (line 88) and here in `resolveBatch`. A future change to release semantics has to touch both, which is easy to miss. Since the PR already extracted a `takeBlobs` method, extracting a `closeAll(List<BlobFile>)` helper alongside it would read more cleanly. Pure cleanup, can be deferred.

LuciferYang (commented):
The `resolveBatch` fix looks correct: every handle `takeBlobs` returned is released once on all error paths (empty, count-mismatch, null-blob, read failure), with no double-close, and `LargeBinaryWriter` reads the copied bytes so closing before return doesn't affect callers. Nothing blocking.

A few non-blocking notes inline. The main one: `singleReadFailureReleasesHandle` is effectively a smoke test — the single-address `resolve` change is behavior-neutral (the old try-with-resources also closed that one blob on a read failure), so the test wouldn't catch a `resolve` regression; the real release semantics are pinned by the `resolveBatch` tests. The rest are optional: `resolve` lacks the null-blob guard `resolveBatch` has, the read-failure tests couple to lance-core's "Not found" text, and the release loop is duplicated across the two methods.

lance-spark-base_2.12/src/test/scala/org/lance/spark/utils/BlobReferenceResolverTest.scala
```diff
@@ -0,0 +1,191 @@
...
+  private class RecordingResolver(afterTake: JList[BlobFile] => Unit)
+    extends BlobReferenceResolver {
+    var acquired: Seq[BlobFile] = Seq.empty
+
+    override def takeBlobs(
+        uri: String,
+        addresses: JList[java.lang.Long],
+        column: String): JList[BlobFile] = {
+      val blobs = super.takeBlobs(uri, addresses, column)
+      acquired = blobs.asScala.filter(_ != null).toVector
```
LuciferYang: In `BlobReferenceResolverTest`, `acquired` is assigned rather than appended (line 53), while `resolveBatch` calls `takeBlobs` once per (datasetUri, columnName) group. There's only ever one group today, so it doesn't show. Add a test that spans two blob columns or two source tables and `assertReleased()` only checks the last group, so a leak in the first one goes unnoticed. Multiple groups are what the finally inside the group loop is there for.

Swapping `=` for `++=` isn't enough on its own. The non-zero handle assertion on line 55 would then cover the previous group too, whose handles the finally has already zeroed, and it fails right there. Take a per-call snapshot instead: `val fresh = blobs.asScala.filter(_ != null).toVector`, assert on `fresh` at lines 54 and 55, then `acquired ++= fresh` after that and before `afterTake`. `assertReleased()` could also use a non-empty check; on an empty `acquired` it passes without asserting anything.

lance-spark-base_2.12/src/test/scala/org/lance/spark/utils/BlobReferenceResolverTest.scala
LuciferYang, line 240: `singleReadFailureReleasesHandle` doesn't guard the new finally. Keep the `takeBlobs` seam and swap only `resolve`'s try/finally back to try-with-resources: it still passes, because it requests a single row and the old form closed that handle just as well when `read()` threw. The `assertReleased()` inside `batchSuccessReleasesAllHandles` held under the old per-item close too. The three that do turn red are the null, count-mismatch and read-failure cases, all on `resolveBatch`.

No production code reaches `resolve` either, only `resolveIfNeeded` calls it and only tests call that. Renaming this one to say it's a smoke test would be enough; deleting it is better left to the PR that deletes `resolve`, or its finally ends up with no test at all. The `assertReleased()` in `batchSuccessReleasesAllHandles` is harmless to keep, it just doesn't guard the new finally.

lance-spark-base_2.12/src/test/scala/org/lance/spark/utils/BlobReferenceResolverTest.scala
LuciferYang, line 234: The two read-failure tests (lines 177, 187) hard-code lance's error text "Not found". Deleting the data files is a sound way to force the failure: `takeBlobs` holds no fd, so the first read is guaranteed to fail. But the text comes from lance-io, so a wording change over there turns both of these red.

Asserting the `IOException` type, plus that the message does not contain "takeBlobs returned", keeps the literal text out and still shows the failure came from lance's read. Unrelated but nearby: the `blobs.remove(...)` at line 160 relies on the list from `takeBlobs` being mutable. Returning `new ArrayList(super.takeBlobs(...))` from the override would be safer.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
LuciferYang: `resolve` calls `blobs.get(0).read()` with no null check, and `takeBlobs` does return null for a null-descriptor row: this PR's own `nullBlobReleasesVisitedAndUnvisitedHandles` builds that case from real data. So the same fault gives you a context-free NPE here and a descriptive IOException with column/dataset in `resolveBatch`. That's the existing behavior, the old code NPE'd the same way, and it isn't a leak.

No production code reaches `resolve` or `resolveIfNeeded` (only tests call the latter), so deleting them is the cleanest end state, though both are public and that should be a separate PR. If they stay, a null guard matching `resolveBatch` would make the two paths report the same way, and it wouldn't need a test of its own.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -133,41 +136,50 @@ public Map<Integer, byte[]> resolveBatch(List<Integer> indices, List<BlobReferen
...
+      List<BlobFile> blobs = takeBlobs(group.datasetUri, addresses, group.columnName);
+
+      // Every handle takeBlobs returned is released in the finally below, including on the two
+      // throws: BlobFile wraps a native handle with no cleaner, so an abandoned one is only
+      // reclaimed when the JVM exits, and Spark retries the task in the same JVM.
+      try {
+        // takeBlobs must return exactly one BlobFile per requested address, in order. A mismatch
+        // means the selection hit deleted/null-descriptor rows, in which case positional mapping
+        // would skew and silently write the wrong bytes into the target table — fail loudly.
+        if (blobs.size() != addresses.size()) {
```
LuciferYang, line 168: The explanation above this check no longer matches the pinned lance version. In 11.0.0-beta.21 a null-descriptor row comes back as a Java `null` in place and the list length doesn't change; this repo's own `BaseBlobV2CopyTest` asserts exactly that. A deleted row makes lance error out instead. Neither one can produce a count mismatch, and the null check below is what catches a null row.

Since these lines are being touched anyway, "means the selection hit deleted/null-descriptor rows" would read better as something closer to "a guard on the takeBlobs contract". The class javadoc at line 43 makes the same claim in different words, so it's worth fixing both, otherwise the file states it twice.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
LuciferYang: The release loop has the same shape in `resolve` (line 88), `resolveBatch` (line 171) and `close` (line 204, over `Dataset`). All three behave the same today, so this is only about readability.

A `closeAllQuietly(Iterable<? extends AutoCloseable>)` on `CloseableUtil` would fold all three in, which fits better than a `BlobFile`-specific helper. Keep `closeQuietly` per element though: `blobs` really does contain nulls on the null-descriptor path. Not urgent, and if `resolve`/`resolveIfNeeded` go away per the other comment it drops to two.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -133,41 +136,50 @@ public Map<Integer, byte[]> resolveBatch(List<Integer> indices, List<BlobReferen
...
+      } finally {
+        for (BlobFile blob : blobs) {
+          CloseableUtil.closeQuietly(blob);
         }
       }
     }
     return resolved;
   }
 
+  List<BlobFile> takeBlobs(String datasetUri, List<Long> addresses, String columnName) {
```
LuciferYang, line 203: `takeBlobs` exists only so `BlobReferenceResolverTest`'s subclass can override it, but nothing on the method says so. The repo already has a form for this: `SparkWrite.getWriteOptions`, `StagedCommit.getStorageOptions` and four others carry a one-line javadoc "Visible for testing.", while `@VisibleForTesting` appears nowhere.

One matching javadoc line covers it. There's precedent for leaving it off (`LanceScanBuilder.shouldNamespaceFtsScan` has none); the cost is that whoever refactors next inlines it back and only finds out at compile time.

LuciferYang (commented):
The fix looks right, and I checked the premise rather than taking it on faith: `org.lance.BlobFile` has no Cleaner and no finalize, the native handle is a raw pointer stashed by jni-rs `set_rust_field`, and only `close()` going through `take_rust_field` reclaims it, so an abandoned handle really does survive until the executor exits. `read()` copies into a fresh JVM array, so closing in the finally can't affect the bytes `LargeBinaryWriter` writes after `resolveBatch` returns, and each handle gets closed exactly once, which matters because `BlobFile.close()` is not idempotent (a second close throws out of `take_rust_field`). Reading the old per-item close against each test, the null, count-mismatch and read-failure cases all land on a non-zero handle in `assertReleased()`, so those three are genuinely pinned.

Nothing here blocks the merge. Two of the comments are worth acting on in this PR. `acquired` in the new test is assigned rather than appended, so `assertReleased()` will quietly only check the last group as soon as anything exercises more than one, and just swapping in `++=` trips a different assertion, so the comment has the shape that works instead. And the comment above the count check names two causes that lance 11.0.0-beta.21 cannot actually produce, which sends the next reader to the wrong branch. The rest are small enough to defer.

jackylee-ch:
Both acted on in 67a8284. `acquired` now appends a per-call snapshot and `assertReleased()` rejects an empty recording; the new `multipleGroupsReleaseEveryGroupsHandles` spans two sources and asserts 6 observed handles, which is 3 under the old assignment.

The count comment and the duplicate claim in the class javadoc now say the check guards the `takeBlobs` contract instead of naming deleted/null-descriptor rows.

Deferred as you suggested: the smoke-test rename, the "Not found" coupling, `resolve`'s null guard, `closeAllQuietly`. 4.1/2.13 1498 tests, 3.5/2.12 1441, 0 failures; lint clean.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -133,41 +138,53 @@ public Map<Integer, byte[]> resolveBatch(List<Integer> indices, List<BlobReferen
...
+
+        for (int i = 0; i < addresses.size(); i++) {
+          BlobFile blob = blobs.get(i);
+          if (blob == null) {
+            throw new IOException(
+                String.format(
+                    "takeBlobs returned a null blob for address %d (column=%s, dataset=%s)",
+                    addresses.get(i), group.columnName, group.datasetUri));
+          }
+          byte[] data = blob.read();
```
LuciferYang: `resolveBatch` drops the per-iteration close and defers releasing the whole group's blobs to the `finally`. A `BlobFile`'s `FileScheduler` is opened lazily on `read()` and freed only on `close()`, so the read loop opens one reader per blob and holds every one until the group ends. When a group's blobs belong to distinct `BlobSource`s (which is how large blobs are stored, one file each), peak concurrently-open readers / FDs / object-store connections goes from ~1 to the group's distinct-file count, which can exhaust file descriptors on large batches.

You can have both: close each blob right after `read()` and null it out in the list, so the `finally`'s `closeQuietly(null)` skips it and there's no double-close (`BlobFile.close` isn't idempotent); on an early throw the `finally` still closes the untouched tail. Success-path peak returns to ~1, and error-path release is unchanged.

lance-spark-base_2.12/src/main/java/org/lance/spark/utils/BlobReferenceResolver.java
```diff
@@ -76,15 +78,18 @@ public BlobReferenceResolver(Map<String, BlobSourceContext> sourceContexts) {
...
-      return new byte[0];
-    }
-    try (BlobFile blob = blobs.get(0)) {
-      return blob.read();
+    List<BlobFile> blobs = takeBlobs(ref.getDatasetUri(), rowAddresses, ref.getColumnName());
+    try {
+      if (blobs.isEmpty()) {
+        return new byte[0];
+      }
+      return blobs.get(0).read();
```
LuciferYang: `resolve()` reads `blobs.get(0).read()` with no null check, while `resolveBatch` throws a descriptive `IOException` for the same null blob. The new javadoc here says a null-descriptor row "comes back as a null in place and leaves the count unchanged", so such a row makes `resolve()` throw a raw NPE instead of the `IOException` its signature declares. The `finally`'s `closeQuietly(null)` is a no-op so nothing leaks; the problem is just the wrong exception type and the behavioral split between the two sibling methods.

Today `resolve()` and `resolveIfNeeded()` are only called from tests (production goes through `resolveBatch`), so this isn't a live crash. But now that the javadoc documents the null-in-place case, giving `resolve()` the same null guard `resolveBatch` has, plus a targeted test, is cheap and closes the gap.

LuciferYang (commented):
The leak fix looks correct and the tests cover the paths well. Two notes from reading the error handling, both non-blocking: the batch path now holds every reader in a group open until the group's `finally`, and `resolve()` is missing the null guard that `resolveBatch` has. Details inline.

jackylee-ch:
Both fixed in cf2cd1c.

The FD point is right and it was a regression I introduced: each blob is now closed as soon as it has been read, with `ownedThrough` recording how far that got so the `finally` releases exactly the untouched tail without a second close. `batchClosesEachBlobBeforeTheNextRead` asserts every earlier blob is already released — it fails with "blob 0 still open at read 1" against the deferred version.

`resolve()` now throws the same descriptive `IOException`, pinned by `singleNullBlobReportsColumnAndDataset` (NPE without it). 4.1/2.13 1500 tests, 3.5/2.12 1443, 0 failures.

PR #819, opened 2026-09-13
build: upgrade to Lance 12.0.0-beta.18 and lance-namespace 0.11.1

FANNG1:
## Summary

Upgrades `lance-core` from `11.0.0-beta.21` to `12.0.0-beta.18`. `lance-core` 12.x is built against `lance-namespace` 0.11.1, so `lance-namespace` moves from `0.8.6` to `0.11.1` in the same change — keeping 0.8.6 fails to compile the tests, because `DirectoryNamespace` (shipped inside `lance-core`) references response types that do not exist in 0.8.6.

This is a dependency upgrade only; no new connector features.

## Behavior changes absorbed

| Lance/namespace change | What this PR does |
|---|---|
| `LanceNamespace.queryTable` returns `QueryTableResponse` instead of `byte[]`; `namespaceExists`/`tableExists` return responses | Unwrap with `.getData()`; update the two test fakes |
| `Update.updateMode` is now mandatory — the transaction protobuf cannot persist "absent" distinctly from `RewriteRows`, so the JNI rejects an empty mode | Position-delta writes (`SparkPositionDeltaWrite`, 3.4 and 3.5) state `RewriteRows` explicitly, which is what an absent mode always meant |
| Lance's default file format version is now **2.2**, which rejects legacy (v1) blob columns outright | A table that requests `<col>.lance.encoding = 'blob'` **without** pinning `file_format_version` is now created at `2.1`, so it keeps the documented v1 behavior instead of failing the write |
| `stable` / `next` resolve to 2.2 or newer | `BlobUtils.fileFormatSupportsBlobV2` recognizes them, so those tables get blob v2 instead of falling through to v1 and being rejected at write time |
| Lance rejects null struct values at file format version 2.0 instead of silently reading them back as non-null structs with null children | `testNullStructV2_0ReadsAsEmptyStruct` becomes `testNullStructV2_0IsRejected` |
| `STABLE` resolves to 2.2 | The two `STABLE` assertions in `BaseSparkConnectorWriteTest` no longer hard-code `2.0`/`2.1` |

The blob-encoding rule is the only user-visible change, and it was chosen to keep existing pipelines byte-identical: without it, every `CREATE TABLE ... TBLPROPERTIES ('x.lance.encoding' = 'blob')` that does not pin a version would start failing. Users who want blob v2 already have to set `file_format_version` explicitly, per the docs. `docs/src/config.md` and `docs/src/operations/ddl/create-table.md` are updated.

## Testing

Full module test suites, all green:

- `lance-spark-3.5_2.12` — 1400 tests, 0 failures, 0 errors
- `lance-spark-3.4_2.12` — 1253 tests, 0 failures, 0 errors
- `lance-spark-4.2_2.13` — 1456 tests, 0 failures, 0 errors
- whole reactor `test-compile` passes; `spotless:check` clean

https://claude.ai/code/session_011gFXXNur56ZyJhu3qdRMcr


FANNG1:
Superseded by #850, which upgrades to the 12.0.0 GA release on top of current `main` (main already absorbed the namespace-response, `updateMode`, and `fileFormatSupportsBlobV2` parts of this PR).

PR #820, opened 2026-09-13
fix: require full zonemap coverage for SPJ

ddupg:
## Summary

- derive live fragment IDs from the same Lance scan plan used to build input splits
- require zonemap stats to cover every live fragment, with every zone in each fragment resolving to one sharding key, before advertising key-grouped partitioning
- ignore stats for retired fragments and retain the legacy overload as a deprecated compatibility entry point

## Motivation

The previous detector inferred the fragment set from the zonemap stats themselves. When a live fragment had no zonemap stats, it was invisible to that check, so the scan could still advertise key-grouped partitioning with only partial coverage. Spark could then choose a storage-partitioned join without the shuffle required for correctness.

## Testing

- `./mvnw test -pl lance-spark-3.5_2.13 -am -Dtest=SparkLanceShardingUtilsTest,LanceScanTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw test -pl lance-spark-3.5_2.12 -am -Dtest=BucketSpjTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw spotless:check -pl lance-spark-base_2.12`


fangbo (approved):
LGTM

PR #821, opened 2026-09-13
refactor: decouple index build utilities for shared use

ivscheianu:
Part of #789 (1/2). Prepares the index build infrastructure for `REFRESH INDEX` (#784).

## Summary

- **`ScalarSegmentIndexJob`** takes a `SparkContext` and primitive parameters instead of the `AddIndexExec` instance, so `REFRESH INDEX` can share the same distributed build path without coupling to `CREATE INDEX`.
- **`extractNamespaceInfo`** and **`isSystemIndex`** move to `IndexUtils`; **`parseNumSegments`** is extracted there as well.
- New `IndexUtils` methods for the refresh path: `methodForIndexType` (reverse `IndexType` → SQL method lookup), `resolveRetainedSegments` (commit-time index re-resolution to catch concurrent `DROP INDEX`), `requireUniformSegmentDetails` (FTS segment compatibility guard).
- **`balancedRunLengths`** improved to split the heaviest splittable run first via a priority queue, so the extra batches land where they help most.

No new SQL commands or grammar changes; existing behavior is preserved. 19 new unit tests for the extracted and added methods (61 total in `IndexUtilsTest`, all pure JVM, no native deps).


ivscheianu:
hey @geruh, can we continue with this one? thanks a lot for all the support in reviewing all these PRs!

geruh (approved):
LGTM!

PR #822, opened 2026-09-13
feat: add distributed REFRESH INDEX for incremental index maintenance

ivscheianu:
Part of #789 (2/2). **Depends on #821** — review that first; the diff here includes it until it merges.

Design: #788. Supersedes #784 (rebased and split).

## Summary

`CREATE INDEX` rebuilds every fragment, so keeping an index current on an append-heavy table means re-indexing the whole table. This adds an incremental counterpart:

```sql
ALTER TABLE lance.db.users REFRESH INDEX user_id_idx [WITH (num_segments = 8)];
```

It indexes only the fragments the index does not cover, distributed the same way `CREATE INDEX` is: the driver diffs coverage against the table and balances the remainder by row count, executors build one uncommitted segment per batch, and the driver commits them as one logical index. Lance core keeps existing segments whose fragments are disjoint from the incoming ones, so prior coverage survives the commit.

A deferred index (`train = false`) covers nothing, so refreshing one builds the whole table through the same distributed path — replacing the guidance to call `Dataset.optimizeIndices`, which runs on a single node.

### Guards

- The index is re-resolved on a fresh handle before the commit, so a `DROP INDEX` during the build is caught rather than silently undone.
- FTS segments are compared against the ones they would join and rejected on configuration mismatch before anything is committed, because an inverted index loads one configuration for all of its segments.

### Known gap

A concurrent `DROP INDEX` between re-resolution and the commit can be silently undone. This needs `lance-format/lance#6806` on the core side. Pinned by a disabled regression test.

### Not in scope

- **Distributed segment merge** — each refresh adds segments; consolidation still requires `CREATE INDEX`.
- **Vector indexes** — not creatable from Spark SQL.
- **Inheriting build parameters** — no binding reads a built index's parameters back; options come from the `WITH` clause with type defaults as fallback.

## Changes (this PR only, excluding #821)

| Component | Files |
|---|---|
| Grammar | `LanceSqlExtensions.g4` (+2 lines: `refreshIndex` rule, `REFRESH` keyword) |
| Parser visitors | 5 files, `visitRefreshIndex` added |
| Logical plan | `RefreshIndex.scala` (new) |
| Strategy | `LanceDataSourceV2Strategy.scala` (+3 lines) |
| Physical exec | `RefreshIndexExec.scala` (227 lines, new) |
| Tests | `BaseRefreshIndexTest.java` (766 lines, 26 tests), parser tests, per-version stubs |
| Docs | `refresh-index.md` (214 lines) |


ivscheianu:
hello, @geruh! Here's the final one. Looks big, but a significant part of it are tests and also a bit of documentation, hopefully it's manageable. Thanks a lot!

ivscheianu:
Hello @geruh, @hamersaw , gentle reminder on this, maybe we can get some traction on this PR so we can close the feature. Thank you for your time!

docs/src/operations/ddl/refresh-index.md
```diff
@@ -0,0 +1,215 @@
...
+!!! warning "Spark Extension Required"
+    This feature requires the Lance Spark SQL extension to be enabled. See [Spark SQL Extensions](../../config.md#spark-sql-extensions) for configuration details.
+
+## Overview
+
+Appending to a table creates new fragments, and an existing index does not cover them until it is
+maintained. `REFRESH INDEX` builds index data only for those uncovered fragments and adds it to the
+index, so cost scales with the newly appended data rather than with the table.
+
+This is the incremental counterpart to re-running [CREATE INDEX](./create-index.md), which rebuilds
```
geruh: can we explain how this differs from `OPTIMIZE INDEX`? that also fills missing coverage, but runs on the driver. `REFRESH INDEX` distributes the build across Spark executors.

docs/src/operations/ddl/refresh-index.md
```diff
@@ -0,0 +1,215 @@
...
+- **Concurrent `DROP INDEX`**: do not drop an index while a refresh of it is in flight. The refresh
+  re-resolves its target immediately before committing and fails if it is gone, but the check and the
+  commit are separate transactions and Lance does not currently treat a concurrent same-name drop as
+  a conflict. A `DROP INDEX` landing in that window is undone: the index reappears covering only the
+  fragments the refresh built. Tracked upstream in
+  [lance#6806](https://github.com/lance-format/lance/pull/6806).
+- **Segment Growth**: each refresh adds segments to the index, and Lance can only compact fragments
+  that are covered by the identical set of index segments. So accumulated refreshes progressively
+  narrow what [OPTIMIZE](./optimize.md) can group: run `OPTIMIZE` before `REFRESH INDEX` rather than
+  after, and rebuild with [CREATE INDEX](./create-index.md) to consolidate the segments back into one
```
geruh: can we point to `OPTIMIZE INDEX` for consolidation? `CREATE INDEX` isn't the only option, and "back into one" isn't guaranteed since each build task produces a segment.

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/LanceDataSourceV2Strategy.scala
```diff
@@ -60,6 +60,9 @@ case class LanceDataSourceV2Strategy(session: SparkSession) extends SparkStrateg
         indexName.toLowerCase(Locale.ROOT),
         args) :: Nil
 
+    case RefreshIndex(ResolvedIdentifier(catalog, ident), indexName, args) =>
+      RefreshIndexExec(asTableCatalog(catalog), ident, indexName.toLowerCase, args) :: Nil
```
geruh: nit: can we use `toLowerCase(Locale.ROOT)`, same as the optimize case above?

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/RefreshIndexExec.scala
```diff
@@ -0,0 +1,219 @@
...
+ *
+ * <p>Indexes only the fragments the named index does not cover, then adds the resulting segments
+ * to it. Cost scales with new data, not the table.
+ *
+ * <p>Build options come from the {@code WITH} clause, not the existing index: Lance does not
+ * expose a built index's parameters as readable metadata. For most types a mismatch only changes
+ * performance; for {@code fts} it is rejected before commit — see
+ * {@code IndexUtils.requiresUniformSegmentDetails}.
+ */
+case class RefreshIndexExec(
```
geruh: nit: can we drop or shorten this comment to what the command does? the option rules are already in the guide

docs/src/operations/ddl/refresh-index.md
```diff
@@ -0,0 +1,215 @@
...
+its segments, and a set that disagrees cannot be queried. `REFRESH INDEX` therefore compares what it
+built against the segments it would join and **fails without committing** when they differ, leaving
+the index exactly as it was:
+
+```
+Index 'idx_text' uses the fts method, whose segments must all share one configuration, and the
+segments this build produced are configured differently from the ones they would join. Nothing was
+committed and the index is unchanged. Re-run with the options the index was created with, or rebuild
+it in full with ALTER TABLE ... CREATE INDEX.
+```
```
geruh: i think we can drop the full error message here? the explanation and recovery steps below should be enough. wdyt?

docs/src/operations/ddl/refresh-index.md
```diff
@@ -0,0 +1,215 @@
...
+
+- **Index Methods**: supported for the same methods as
+  [CREATE INDEX](./create-index.md#index-methods): `zonemap`, `btree`, `bitmap`, `label_list`,
+  `ngram`, `bloomfilter`, `rtree`, and `fts` (or `inverted`). A vector index cannot be created from
+  Spark SQL either, so refreshing one is rejected and pointed at the Lance SDK rather than at
+  `CREATE INDEX`, which could not build it.
+- **Full-Build Options**: `train`, `build_mode`, and `rows_per_range` are rejected. Range-mode
+  `btree` redistributes and sorts the whole table, which an incremental refresh does not do — use
+  `CREATE INDEX` to rebuild that way.
+- **Method Options Are Not Inherited**: new segments are built from the `WITH` clause and the index
```
geruh: this is redundant w/ the Index Method Options section. can we link there instead?

docs/src/operations/ddl/refresh-index.md
```diff
@@ -0,0 +1,215 @@
...
+refresh restores it:
+
+=== "SQL"
+    ```sql
+    OPTIMIZE lance.db.users;
+
+    ALTER TABLE lance.db.users REFRESH INDEX zone_idx;
+    ```
+
+The other methods record row ids, which follow their rows through a rewrite, so compaction carries
```
geruh: wdyt about dropping this paragraph seems like the compaction note lists which methods lose coverage already.

geruh (commented):
Hey @ivscheianu, thanks for adding this!! I took a pass, and left a few comments

Issue #333, opened 2026-03-24
Report shuffle read size / input size


hamersaw:
Currently, the Spark connector does not report any metrics for analysis. This makes it difficult to reason about performance.

PR #824, opened 2026-09-13
feat: report write metrics to stage output metrics and the SQL tab

venkata91:
Part of #333

## Problem

Stage `outputBytes` and `outputRecords` are always 0 for Lance writes, in the Stages UI and in the
history server REST API, because the connector reports no write metrics. The repo's own benchmark
harness hits this: `QueryMetricsListener.java:101` sums `outputMetrics().bytesWritten()` and always
gets 0.

## Fix

Adds a `write/metric` package mirroring `read/metric`. `LanceWriteMetricsTracker` counts rows in
`write()` and sums bytes from `DataFile.getFileSizeBytes()` as fragments complete;
`LanceDataWriter` reports both from `currentMetricsValues()`.

Notes:

- `bytesWritten` and `recordsWritten` are reserved names. `execution.metric.CustomMetrics.updateMetrics`
  matches exactly those two and routes them into the task's output metrics, which is what fills the
  stage counters. A test pins the literals so a rename cannot break that silently.
- Only `recordsWritten` is advertised as a SQL metric, per @lance-gatekeeper's review. SQL metrics
  come from `currentMetricsValues()`, whose last poll is just before `commit()`, and Lance fragments
  only complete inside `commit()` unless the write is sharded. An advertised `bytesWritten` would
  read 0 on an ordinary write, so the writer publishes it to output metrics after commit instead.
- `abort()` clears the task's output metrics. The reserved names are routed from inside the write
  loop, so a failed attempt has already reported a partial row count, and
  `AppStatusListener.onTaskEnd` folds a task's metrics into the stage totals whatever its end
  reason. Without the clear, a 500-row write that retries once reports `outputRecords` 900 while
  the SQL tab correctly reports 500.

Not closed: `LanceBatchWrite.useCommitCoordinator()` is `false`, so a speculative attempt killed
after it commits can still double count. Rarer than the retry case, and inherent to that setting.

## Verification

- `./mvnw test -pl lance-spark-3.5_2.13 -am`: 1408 tests, 0 failures
- `testFailedAttemptDoesNotInflateOutputMetrics` injects a failure at row 450 of 500 on attempt 0.
  Reverting the `abort()` clear fails it with `expected: <500> but was: <900>`
- `testInsertPopulatesOutputMetrics` asserts `outputMetrics` after an `INSERT`; commenting out the
  post-commit publish fails it with `outputMetrics.bytesWritten should be > 0, was 0`
- `./mvnw spotless:apply`


lance-spark-base_2.12/src/main/java/org/lance/spark/write/SparkWrite.java
```diff
@@ -147,6 +149,11 @@ public SortOrder[] requiredOrdering() {
         .toArray(SortOrder[]::new);
   }
 
+  @Override
+  public CustomMetric[] supportedCustomMetrics() {
+    return LanceWriteMetrics.allMetrics();
```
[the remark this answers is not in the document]
venkata91: You're right, and thanks for the reproducer -- I ran it before changing anything and it fails exactly as you described on 4ea797a:

```
[ERROR] LanceWriteMetricsIntegrationTest.testInsertPopulatesSqlBytesWrittenMetric
org.opentest4j.AssertionFailedError: SQL bytesWritten should be > 0, was 0
```
with the existing `outputMetrics.bytesWritten > 0` assertion still passing in the same run. Advertising a metric that renders 0 on every single-fragment write is worse than not advertising it.

**I looked for a post-commit path to the SQLMetric before taking your first option.** Against the actual bytecode:

- `WritingSparkTask.$anonfun$run$5` (3.5.5): the loop-body `updateMetrics` is at offset 43 and the final post-loop one at offset 95; `DataWriter.commit()` is at 156/215. Nothing reads `currentMetricsValues()` after offset 95. Same shape in 4.1.1 (last `updateMetrics` at 27, `commit` at 95/165).
- `DataWritingSparkTaskResult` carries only `numRows` and `writerCommitMessage`, and `WriterCommitMessage` is a bare marker interface with zero methods, so nothing can be read back out of it.
- In `V2TableWriteExec.writeWithV2`, after `BatchWrite.commit(messages)` there is only `logInfo` and `commitProgress_$eq(StreamWriterCommitProgress)` before `areturn`. No `postDriverMetricUpdates`, no touch of `customMetrics()`.

So on 3.4/3.5 there is no path, full stop. There is one on 4.1+: `Write.reportDriverMetrics()`, which `V2ExistingTableWriteExec.run()` calls after `writeWithV2` and posts via `SQLMetrics.postDriverMetricUpdates` into the very SQLMetrics built from `supportedCustomMetrics()`. I didn't take it: it doesn't exist on the 3.4/3.5 modules this connector still supports, it only covers plans routed through `V2ExistingTableWriteExec` (append/overwrite/replace, not CTAS/RTAS), and it'd mean version-forked metric plumbing for a value users already get correctly from `outputBytes`.

**So: your first option.** `bytesWritten` is no longer in `supportedCustomMetrics()`. It's still reported as a `CustomTaskMetric`, which is enough -- `CustomMetrics.updateMetrics` routes the reserved name into `TaskContext.taskMetrics().outputMetrics()` regardless of whether the metric is advertised -- and the post-commit `publishOutputMetrics()` stays, so stage-level `outputBytes` in the Stages UI and history server (the actual problem) is correct. `docs/src/performance.md` now states per metric where it shows up and why `bytesWritten` is off the SQL tab.

**One refinement to your description, verified rather than assumed.** `recordsWritten` does *not* have this problem, so I kept it advertised. Rows are counted in `write()`, so the value is already final at Spark's last poll; there is no commit-time contribution. I checked this on your own harness rather than reasoning about it: with the SQLMetric listener extended to capture `recordsWritten` too, the SQL value is exactly 50 for a 50-row insert while `bytesWritten` is 0 in the same query. The two metrics fail differently and only one needed dropping.

**Regression test.** Your reproducer is kept, inverted: `testSqlMetricsAreOnlyTheOnesThatCanBeCorrect` asserts SQL `recordsWritten == rows`, asserts no SQL metric named `bytesWritten` is present at all, and asserts `outputMetrics.bytesWritten > 0`. A unit test also pins that `allMetrics()` never contains `bytesWritten`, so re-adding it trips CI rather than shipping a 0.

`./mvnw test -pl lance-spark-3.5_2.13 -am` on openjdk@17: **1407 tests, 0 failures, 110 skipped** (write metric tests: 2 + 9, all passing).

lance-spark-base_2.12/src/main/java/org/lance/spark/write/metric/LanceWriteMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+ * <p>Only {@code recordsWritten} is advertised as a SQL metric. SQL metrics come from {@code
+ * currentMetricsValues()}, whose last poll is just before {@code commit()}. Fragments only complete
+ * inside {@code commit()} unless the write is sharded, so an advertised {@code bytesWritten} would
+ * read 0 on an ordinary write. {@link LanceWriteMetricsTracker#publishOutputMetrics()} reports it
+ * after commit instead.
+ */
+public final class LanceWriteMetrics {
+  public static final String BYTES_WRITTEN = "bytesWritten";
+
+  public static final String RECORDS_WRITTEN = "recordsWritten";
```
geruh: hi

lance-spark-base_2.12/src/main/java/org/lance/spark/write/metric/LanceWriteMetrics.java
```diff
@@ -0,0 +1,63 @@
...
+ * outputBytes}/{@code outputRecords}. That happens for any reported task metric with a reserved
+ * name, advertised here or not.
+ *
+ * <p>Only {@code recordsWritten} is advertised as a SQL metric. SQL metrics come from {@code
+ * currentMetricsValues()}, whose last poll is just before {@code commit()}. Fragments only complete
+ * inside {@code commit()} unless the write is sharded, so an advertised {@code bytesWritten} would
+ * read 0 on an ordinary write. {@link LanceWriteMetricsTracker#publishOutputMetrics()} reports it
+ * after commit instead.
+ */
+public final class LanceWriteMetrics {
```
geruh, line 20: nit: feel free to skip or as a follow up but i think we can keep the story here in the `performance.md` docs and shorten some of the comments. 

The same reserved name, last poll, and abort explanation is in this class, the tracker, and the retry test is in the docs already.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/metric/LanceWriteMetricsTracker.java
```diff
@@ -0,0 +1,121 @@
...
+          }
+        },
+      };
+
+  /** Counts a row as it is handed to the writer, rather than deriving it from fragments. */
+  public void incrementRecordsWritten() {
+    recordsWritten++;
+  }
+
+  /** Data files with an unknown size are skipped rather than counted as zero. */
```
geruh: skipped rather than counted as zero has the same effect on the total. this can just say sums known fragment data-file sizes.

geruh (commented):
Nice work @venkata91! Tested this out locally, and functionality looks good to me. I left 2 nits let me know what you think.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/metric/LanceWriteMetrics.java
venkata91: Agreed! Addressed in 7fa450a: the lifecycle details now stay in performance.md, and this class comment only describes the advertised SQL metrics.

lance-spark-base_2.12/src/main/java/org/lance/spark/write/metric/LanceWriteMetricsTracker.java
venkata91: Addressed in 7fa450a with the suggested wording: "Sums known fragment data-file sizes."

geruh (commented):
LGTM thanks @venkata91!

PR #825, opened 2026-09-13
ci: take the MinIO test binary from quay.io instead of dl.min.io

FANNG1:
## Problem

Every Docker-based CI job is currently failing on `main` and on all open PRs — the seven `Integration Test` matrix entries plus `Search Docker Test`. They die while building the test base image:

```
#9 0.376 curl: (22) The requested URL returned error: 410
#9 ERROR: process "/bin/sh -c curl -fL -o /usr/local/bin/minio https://dl.min.io/server/minio/release/linux-amd64/minio && chmod +x /usr/local/bin/minio" did not complete successfully: exit code: 22
```

MinIO retired the standalone binary downloads at `dl.min.io` — both the floating URL and the `archive/minio.RELEASE.*` ones now return **HTTP 410** — and also locked down the Docker Hub repo, so `docker pull minio/minio` returns `pull access denied`. Their quay.io images are still public.

This is unrelated to any particular PR; the last green run on `main` was 2026-09-09.

## Fix

Lift the server binary out of a pinned quay.io image with a multi-stage `COPY`:

```dockerfile
COPY --from=quay.io/minio/minio:RELEASE.2025-09-07T16-13-09Z /usr/bin/minio /usr/local/bin/minio
```

- Pinned to `RELEASE.2025-09-07T16-13-09Z` — the last community release, and the last GitHub release that ships binaries at all — rather than `:latest`, so image builds stay reproducible.
- `COPY --from` resolves the right architecture; the old download hard-coded `linux-amd64`.
- `docker/docker-compose.yml` moves to the same pinned quay.io image, for the same "pull access denied" reason.

Nothing about how the tests use MinIO changes — `integration-tests/conftest.py` still runs `minio server /tmp/minio-data` as a local process on port 9000.

## Verification

The download source and the new layer were checked directly:

```
$ curl -sIL -o /dev/null -w '%{http_code}\n' https://dl.min.io/server/minio/release/linux-amd64/minio
410
$ docker build .        # FROM ubuntu:24.04 + the COPY line above
#8 0.437 minio version RELEASE.2025-09-07T16-13-09Z (commit-id=07c3a429bfed433e49018cb0f78a52145d4bedeb)
```

CI on this PR exercises the real thing end to end.

https://claude.ai/code/session_011gFXXNur56ZyJhu3qdRMcr


hamersaw (approved):

hamersaw:
Thanks for the fix!

geruh:
Thanks for this @FANNG1, I've seen some other projects move to this for the short term. I wonder if we should move to something like rust fs or seaweed in the long term?

FANNG1:
@geruh Good question. My take: no strong reason to switch in the short term, and the long-term answer probably depends on how these projects evolve and on what the maintainers want.

Short term, nothing is actually at risk. What we pin here is an immutable tag on a public Quay repo, so it will keep resolving; the cost of staying is no upstream security updates for the test-only MinIO, not a build that breaks again. Worth noting `RELEASE.2025-09-07T16-13-09Z` is the last community release either way — it is also the newest GitHub release that still ships binaries at all.

Long term, the two candidates look quite different today:

- **RustFS** — Apache-2.0, positions itself as a drop-in MinIO replacement, so the smallest migration. Still pre-GA (`1.0.0-rc.6`).
- **SeaweedFS** — Apache-2.0, clearly more mature, but S3 is one gateway in a larger system, so it is a bit more than the single `minio server /tmp/minio-data` process the fixture runs today.

If we do evaluate, I would weight one thing above maturity or activity: whether the replacement implements the conditional-write semantics Lance's S3 commit path relies on. A backend that is "S3-compatible" for get/put can still fail the commit path in confusing ways, and that is cheap to check up front by running the existing `minio` backend suite against it.

The migration surface is small either way — the binary line in `docker/Dockerfile.test-base`, the `minio` fixture in `integration-tests/conftest.py` (start command, `/minio/health/live` probe, port), and the two Compose services — so there is no cost to waiting for RustFS to reach GA before comparing them properly.

PR #831, opened 2026-09-15
docs: fix two broken cross-page anchors

jackylee-ch:
`fts.md` links to `create-index.md#fts-options` and `insert-into.md` links to `create-table.md#copying-blob-v2-columns-with-ctas`. Neither id is emitted — the real slugs are `fts-inverted-options` (`create-index.md:84`) and `ctas-from-a-blob-v2-source` (`create-table.md:367`), so both links land on the page top instead of the section.

Swept all 95 intra-docs links across the 37 pages; these are the only two broken ones. Verified against a `mkdocs build --strict` site: the two new ids are present in the rendered HTML and the two old ones are absent.


geruh (approved):
LGTM, thanks!

PR #832, opened 2026-09-15
docs: state the full write restriction on tags and branches

jackylee-ch:
The tag paragraph listed six DML and schema commands, but `LanceDataset.ensureWritable` rejects every mutating command once the read carries a branch-or-tag ref. `CREATE INDEX`, `DROP INDEX`, `OPTIMIZE`, `VACUUM`, `SET UNENFORCED PRIMARY KEY`, the two backfill commands and branch/tag DDL all reach it through `requireWritable` (11 call sites), and DML reaches it through `newWriteBuilder`.

Leads with the general rule the way the branch section 45 lines below already does, so the sentence cannot go stale as commands are added, and keeps the command names as a hint rather than a contract. Also notes that branch identifiers behave identically — the guard is `isBranchOrTag()`, not tag-only.


geruh (approved):
Thanks!

PR #833, opened 2026-09-15
fix: accept any numeric literal for OPTIMIZE options

jackylee-ch:
The extension grammar boxes a numeric literal as `Long`, `Float` or `Double` depending on how it was written (`1`, `0.5`, `0.5d`), but `buildOptions` cast each option to one fixed box. So `OPTIMIZE t WITH (materialize_deletions_threshold = 1)` — the natural way to ask for "always materialize" — threw `java.lang.ClassCastException: class java.lang.Long cannot be cast to class java.lang.Float`, and `= 0.5d` threw the same for `Double`. Only `0.5` and `0.5f` parse as `Float` and worked; the one existing test happens to use `0.2f`.

Routes every numeric option through a `Number` accessor that takes all three boxes, and names the option when the value is not numeric — the wording `AddIndexExec.extractTrain` already uses for `train`. Booleans get the same treatment, so a wrong literal no longer surfaces a raw cast error either.

Reverting just the threshold accessor reproduces the `Long cannot be cast to Float` failure and turns the non-numeric message back into `class java.lang.String cannot be cast to class java.lang.Float`. With the fix: 7 tests, 0 failures on both 4.1/2.13 and 3.5/2.12.


lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseOptimizeTest.java
```diff
@@ -137,4 +137,57 @@ public void testWithoutArgs() {
...
     Assertions.assertEquals("[10,1,10,1]", result.collectAsList().get(0).toString());
   }
+
+  @Test
+  public void testMaterializeDeletionsThresholdAcceptsAnyNumericLiteral() {
+    prepareDataset();
+
+    // The grammar boxes 1 as Long, 0.5 as Float and 0.5d as Double, so every spelling has to
+    // reach withMaterializeDeletionsThreshold instead of failing the cast. Only the first call
+    // has fragments left to compact, so assert acceptance rather than the compaction counts.
```
geruh, line 147: nit: test is pretty selfdocumenting we can drop the comment

lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/OptimizeExec.scala
```diff
@@ -36,25 +36,48 @@ case class OptimizeExec(
...
 
     builder.build()
   }
 
+  /**
+   * The SQL extension grammar boxes a numeric literal as Long, Float or Double depending on how it
+   * was written (`1`, `0.5`, `0.5d`), so an option documented as a float cannot assume one of them.
+   * Options documented as Long stay Long-only on purpose: widening them would let `2.5` through as
+   * a silently truncated `2` instead of being rejected.
+   */
```
geruh, line 59: nit:  we can drop the javadoc. floatArg is any Number. the Long-only rule belongs on longArg w/Number.longValue would turn 2.5 into 2.

geruh:
Thanks for working on this @jackylee-ch!!

jackylee-ch:
Done in ed75395. Dropped both — the test comment and the helper javadoc. Kept a one-liner on `longArg` so the Long-only rule (reject rather than `longValue`-truncate 2.5 to 2) stays visible where you said it belongs.

geruh (approved):
lgtm!

Issue #43, opened 2025-07-30
misleading error exception


xx789633:
https://github.com/lancedb/lance-spark/blob/621876f9fd49174ed6d681ee0a33596c8497df0f/lance-spark-base_2.12/src/main/java/com/lancedb/lance/spark/LanceCatalog.java#L61

This `TableAlreadyExistException` is misleading if the dataset creation error is caused by some other problems like invalid uri.

PR #834, opened 2026-09-15
fix: name the table when drop index misses

jackylee-ch:
`ALTER TABLE t DROP INDEX no_such_idx` surfaces lance-core's raw error:

```
Lance index not found: name=no_such_idx, /Users/runner/work/lance/lance/rust/lance/src/index.rs:1640:24
```

It names the index and a build-machine path, but not the table — a session working over several tables cannot tell which one was meant.

Wraps the `dropIndex` call so the message leads with the table identifier and index name, keeping the original as the cause. Same shape as `IndexUtils.runSegmentTasks`, which already wraps segment build failures in a `RuntimeException(context, cause)`.

Reverting just the wrapper makes the new test fail with `error should name the table, got: Lance index not found: name=no_such_idx, .../index.rs:1640:24`. With the fix: 3 tests, 0 failures on both 4.1/2.13 and 3.5/2.12.

Part of #43.


lance-spark-base_2.12/src/main/scala/org/apache/spark/sql/execution/datasources/v2/DropIndexExec.scala
```diff
@@ -42,6 +42,14 @@ case class LanceDropIndexExec(
     val dataset = Utils.openDatasetBuilder(readOptions).build()
     try {
       dataset.dropIndex(indexName)
+    } catch {
+      // lance-core reports a missing index as a bare RuntimeException whose message names the
+      // index and its own Rust source path but not the table, so a session working over several
+      // tables cannot tell which one was meant. Keep the cause and lead with both identifiers.
+      case e: Exception =>
```
geruh, line 47: nit: this pretty verbose we can just say something like 

`// Add table context while preserving the underlying Lance error.`

Also, I have a question should this have the scope of Exception or just throwables?

lance-spark-base_2.12/src/test/java/org/lance/spark/update/BaseAddIndexTest.java
```diff
@@ -1727,6 +1727,29 @@ public void testDropIndex() {
     }
   }
 
+  @Test
+  public void testDropMissingIndexNamesTableAndIndex() {
+    prepareDataset();
+
+    // Without the wrapper, lance-core's message is
```
geruh: nit: we can drop the comment since test name explains intent. Also, should we assert that the message includes both identifiers?

geruh (commented):
Thanks for raising @jackylee-ch! Left some comments

geruh (approved):
LGTM, thanks for addressing comments! pending CI

PR #835, opened 2026-09-15
docs: drop the dead rows_per_range create index option

jackylee-ch:
`rows_per_range` is documented as a `Long` with "Default is 1000000", but nothing reads it. In `main` it appears only in `AddIndexExec`'s `SparkOnlyOptions` strip set (`:801`) and that class's javadoc (`:70`), so it never reaches the Lance index backend either — `IndexUtilsTest` asserts it is stripped from the JSON params. The range build takes its partition count from the fragment count (`Math.max(1, numFragments)`), so setting the option has no effect and no warning.

Contrast within the same command: `zone_size` is read at `AddIndexExec.scala:333`, and `rows_per_zone` is absent from the strip set so it passes through to Lance. Only `rows_per_range` is inert.

Replaces the row with a sentence on what actually determines the range partitioning. The strip set is left alone so statements that already pass the option keep working.


docs/src/operations/ddl/create-index.md
```diff
@@ -78,7 +78,9 @@ For the `btree` method, the following options are supported:
 |------------------|--------|----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | `zone_size`      | Long   | The number of rows per zone in the B-tree index.                                                                                                                                                         |
 | `build_mode`     | String | Index building mode: 'fragment' builds indexes in parallel by fragment; 'range' sorts data by indexed columns first, then partitions and builds indexes in parallel by partition. Default is 'fragment'. |
-| `rows_per_range` | Long   | The number of rows per range when built using range mode. Default is 1000000.                                                                                                                            |
+
+`build_mode = 'range'` partitions on fragment id so each partition holds whole fragments, and takes
+its partition count from the fragment count. Range width is not configurable.
```
geruh: the build_mode cell still says range sorts by the indexed columns and then partitions amd this says fragment id. can consolidate or atleast make them match the fragment count is the width.

PR #836, opened 2026-09-16
perf: weight pruned scan stats by row count

ddupg:
## Summary

- estimate post-zonemap scan row counts from deletion-aware row counts of surviving fragments
- scale byte estimates by the surviving-row ratio before applying projection estimates
- preserve conservative handling for empty datasets, retired fragments, and non-selective pruning

## Testing

- `./mvnw test -pl lance-spark-3.5_2.13 -am -Dtest=LanceStatisticsTest,LanceSplitTest,LanceScanTest,LanceScanBuilderTest -Dsurefire.failIfNoSpecifiedTests=false`
- `./mvnw spotless:check -pl lance-spark-base_2.12`
- `./mvnw test -pl lance-spark-3.5_2.12 -am -Dtest=LanceStatisticsTest -Dsurefire.failIfNoSpecifiedTests=false`


lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatistics.java
```diff
@@ -67,6 +67,30 @@ public static LanceStatistics estimatePostPruning(
     return new LanceStatistics((long) (totalRows * ratio), (long) (totalFilesSize * ratio));
```
geruh, line 67: oh interesting, we arent using the `estimatePostPruning` and is dead code by the looks of it. Maybe we should remove this in a follow up or somehting.

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatistics.java
```diff
@@ -67,6 +67,30 @@ public static LanceStatistics estimatePostPruning(
...
   }
 
+  /**
```
geruh, line 70: nit: this javadoc is pretty verbose we can probably shorten if not drop the java doc wdyt?

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceScanBuilder.java
```diff
@@ -247,15 +247,24 @@ public Scan build() {
...
       // (when the projected schema is narrower than the full schema).
       long projectedRows = summary.getTotalRows();
       long projectedFullSize = summary.getTotalFilesSize();
-      if (survivingFragmentIds != null && summary.getTotalFragments() > 0) {
-        double ratio = (double) survivingFragmentIds.size() / summary.getTotalFragments();
-        projectedRows = (long) (projectedRows * ratio);
-        projectedFullSize = (long) (projectedFullSize * ratio);
+      if (survivingFragmentIds != null && !scanPlan.getFragmentRowCounts().isEmpty()) {
+        // Zonemap stats can still contain retired fragment IDs. Only row counts from the live
+        // fragments in this scan plan may contribute to the estimate.
```
geruh, line 257: nit: `getOrDefault(fragmentId, 0L)` is already droppping retired ids. the comment on top seems redundant maybe we can drop or simplify?

geruh (commented):
Thanks for raising @ddupg, did a pass!

lance-spark-base_2.12/src/main/java/org/lance/spark/read/LanceStatistics.java
ddupg: Good catch. This was already dead code before this PR, so I’d prefer to clean it up separately.

ddupg:
> Thanks for raising @ddupg, did a pass!

@geruh Thanks for the review! I’ve updated the PR based on comments.

geruh (approved):
lgtm!

PR #843, opened 2026-09-20
fix: keep parameterized SQL and SQL UDFs working with the extension

jackylee-ch:
## Summary

Enabling `spark.sql.extensions=org.lance.spark.extensions.LanceSparkSessionExtensions` breaks two
plain Spark SQL features that have nothing to do with Lance, because `LanceSparkSqlExtensionsParser`
does not delegate two `ParserInterface` methods:

- `parsePlanWithParameters` (Spark 4.1+) is not overridden. The interface default forwards to
  `parsePlan(sqlText)` and drops the `ParameterContext`, so `spark.sql("... id = ?", args)` fails
  with `[UNBOUND_SQL_PARAMETER] Found the unbound parameter: _35`. `SparkSqlParser` implements it
  for real, so the parameters are lost only because the wrapper shadows it.
- `parseRoutineParam` (Spark 4.0+) throws a message-less `UnsupportedOperationException`.
  `SessionCatalog` calls it through `UserDefinedFunction.fromCatalogFunction` with its own parser,
  so resolving a persistent SQL UDF fails as soon as the extension is installed.

Both methods now delegate, keeping the `CREATE INDEX` interception and the extension-grammar
fallback that `parsePlan` already had; the guard moved into a private helper so the two entry points
share it. Spark 4.0 has no `parsePlanWithParameters`, so only `parseRoutineParam` changes there.

## Testing

New `SqlExtensionsParserDelegationTest` in each 4.x module (self-contained session, no Lance table
needed). Without the fix both cases fail with the errors above; with it:

- `mvn test -pl lance-spark-4.1_2.13` — 1494 tests, 0 failures (incl. `CreateIndexStandardSyntaxTest`)
- `mvn test -pl lance-spark-4.0_2.13 -Dtest=CreateIndexStandardSyntaxTest,SqlExtensionsParserDelegationTest` — 56 tests, 0 failures
- `mvn test -pl lance-spark-4.2_2.13 -Dtest=CreateIndexStandardSyntaxTest,SqlExtensionsParserDelegationTest` — 57 tests, 0 failures
- `make lint` clean


jackylee-ch:
cc @LuciferYang

LuciferYang:
Checked the delegation completeness, which is the main risk for this kind of wrapper fix: on the 4.0/4.1/4.2 wrappers, `parsePlanWithParameters` and `parseRoutineParam` were the only `ParserInterface` members not reaching the delegate, and both are fixed here. Every other override already delegates, and 3.x doesn't carry these two methods on its interface, so the 4.x-only scope is right. The `parsePlanWithParameters` fallback to the Lance grammar dropping parameters is fine, since the extension statements carry no `?` markers. Looks correct.

LuciferYang (approved):
Approving. Delegation completeness verified as noted above.

LuciferYang:

PR #844, opened 2026-09-21
fix: re-parse write options on the row-level DML path

jackylee-ch:
## Summary

`LancePositionDeltaOperation` builds its `LanceSparkWriteOptions` with `storageOptions()` alone, so
the typed write settings are never parsed out of the option map. `max_row_per_file`,
`max_rows_per_group`, `max_bytes_per_file`, `batch_size`, `queue_depth`, `use_large_var_types` and
the rest stay at their defaults for UPDATE / DELETE / MERGE, while the same keys take effect on
INSERT, which goes through `fromOptions()` in `LanceDataset#newWriteBuilder`. Catalog-level settings
are affected too: `Utils` merges them into the read options' storage options, and this path copies
that map without parsing it.

`file_format_version` is already hand-forwarded here from the table, which is what made the missing
`fromOptions()` look accidental rather than intended.

The map now goes through `fromOptions()` on this path as well. The table's `file_format_version` and
`enable_stable_row_ids` only apply when the options do not set them, matching the precedence the
INSERT path already had — before this, a per-write `file_format_version` was overwritten by the
table's value. Spark write options are merged in too, with the internal blob-source-contexts key
stripped the way the INSERT path strips it.

## Testing

New `PositionDeltaWriteOptionsTest` in both 3.4 and 3.5. Against the old code two of its three cases
fail with `expected: <7> but was: <null>` (option dropped) and `expected: <2.2> but was: <2.0>`
(table version overriding an explicit option); the third pins that the table value still applies
when the options are silent.

- `mvn test -pl lance-spark-3.5_2.12` — 1444 tests, 0 failures
- `mvn test -pl lance-spark-3.4_2.12` — 1297 tests, 0 failures
- `mvn test -pl lance-spark-4.1_2.13` — 1501 tests, 0 failures
- `make lint` clean


jackylee-ch:
cc @LuciferYang

LuciferYang:
Reviewed this one closely since it changes the write-option plumbing. The core change is correct: `fromOptions` copies the whole option map verbatim into `storageOptions`, so credentials and object-store settings survive on UPDATE/DELETE/MERGE, and the merge order plus the blob-key strip match `LanceDataset#newWriteBuilder` (the INSERT path). 4.0/4.1/4.2 compile this same shared source via build-helper, so there's no 4.x twin still needing the fix.

One optional test-coverage note: the `file_format_version` precedence is pinned in both directions, but the symmetric `enable_stable_row_ids` branch added at `LancePositionDeltaOperation.java:158` (`!mergedOptions.containsKey(CONFIG_ENABLE_STABLE_ROW_IDS)`, same in the 3.4 copy) has no test, and it does change behavior: the table property used to always apply, and now a write-option `enable_stable_row_ids` overrides it. A mirror case in `PositionDeltaWriteOptionsTest` would cover it. Non-blocking.

jackylee-ch:
Added in b88b3e3: `PositionDeltaWriteOptionsTest` now has both cases — the table `enable_stable_row_ids` applies when the write options omit it, and a write-option value overrides it — for 3.4 and 3.5. Dropping the `containsKey` guard makes the override case fail (expected false but was true), so it's not vacuous.

LuciferYang (approved):
Approving. The core change is correct as noted above; the `enable_stable_row_ids` test case is a non-blocking suggestion.

LuciferYang:

PR #845, opened 2026-09-21
fix: count buffered blob v2 bytes in the per-batch write budget

jackylee-ch:
## Summary

`BlobV2StructWriter` does not override `estimatedBufferedBytes`, so it returns the base class's `0`
while its data child (`LargeBinaryWriter`) buffers blob references that resolve to real blob bytes in
`finish()`. `SemaphoreArrowBatchWriteBuffer` adds that number to `currentBatchBytes` precisely
because the buffered references are tiny and their resolved blobs are not — its comment says the
guard would otherwise "let the batch grow until resolution OOMs the executor".

For a blob v2 column that is exactly what happens: the byte budget sees nothing, `max_batch_bytes`
never fires, and the batch fills to `batch_size` rows before resolving every reference at once. The
same copy from a blob v1 source flushes normally, because there the bytes land in the vector as they
are written.

The mechanism arrived with the shuffle support in #355; blob v2 write support came later in #560 and
added this writer without wiring it up. It now delegates to the data child, the way `ArrayWriter`,
`MapWriter` and `StructWriter` already do.

## Testing

New case in `BlobV2StructWriterTest`: two rows carrying references that declare 4096 and 1024 bytes,
asserting the root writer reports 5120. Without the override it reports `0`.

- `mvn test -pl lance-spark-4.1_2.13` — 1499 tests, 0 failures
- `mvn test -pl lance-spark-3.5_2.12` — 1442 tests, 0 failures
- `make lint` clean


jackylee-ch:
cc @LuciferYang

LuciferYang:
The override matches how `ArrayWriter`/`MapWriter`/`StructWriter` delegate, and it composes when a blob-v2 column is nested inside them. The `uri`/`position`/`size` siblings aren't missed: they're synthesized on read and buffer nothing on the write path, so `dataWriter` holds the only externally-buffered bytes. `LargeBinaryWriter` zeroes its pending bytes in `finish()`/`reset()`, so the per-batch figure stays correct across batches. Looks correct.

LuciferYang (approved):
Approving. Verified as noted above.

LuciferYang:

PR #846, opened 2026-09-21
fix: point the benchmark submit scripts at the built jar name

jackylee-ch:
## Summary

`benchmark/pom.xml` sets `<finalName>${project.artifactId}</finalName>`, so `make benchmark-build`
produces `benchmark/target/lance-spark-benchmark.jar`. The four `submit-*.sh` scripts still default
to `target/lance-spark-benchmark-0.3.0-beta.1.jar`, a name the build has not produced since #370
decoupled the module from the project version — the five wrapper scripts (`generate-data.sh`,
`run-benchmark.sh`, `create-indexes.sh`, and the TPC-H pair) were updated then, the submit defaults
were not, and #403 copied the stale path into the two new TPC-H submit scripts.

Because the default never matches a file, the `[[ ! -f ]]` guard is always true: the script runs a
full Maven rebuild and then aborts with `Error: benchmark jar not found at ...`, so none of the four
cluster-submit paths documented in `benchmark/README.md` work unless `--benchmark-jar` is passed.

The README's eight `spark-submit` examples were stale the same way: four name the versioned jar, and
four glob `benchmark/target/lance-spark-benchmark-*.jar`, which does not match the unversioned name
either.

## Testing

`make benchmark-build SPARK_VERSION=3.5 SCALA_VERSION=2.12` produces exactly
`target/lance-spark-benchmark.jar` (plus the shade plugin's `original-` copy).

Driving `submit-benchmark.sh` with a stand-in `spark-submit` on `SPARK_HOME`:

- before: `Building benchmark jar ...` then `Error: benchmark jar not found at
  .../target/lance-spark-benchmark-0.3.0-beta.1.jar`
- after: `Benchmark jar: .../target/lance-spark-benchmark.jar`, and the submit line carries that path

`bash -n` clean on all four scripts.


geruh (approved):
lgtm

PR #847, opened 2026-09-21
docs: fix the catalog class in the LanceDataSource javadoc

jackylee-ch:
The catalog example in `LanceDataSource`'s javadoc still names
`com.lancedb.lance.spark.LanceNamespaceSparkCatalog`. That class does not exist — it is
`org.lance.spark.LanceNamespaceSparkCatalog`, which is what `docs/src/config.md` documents — so
copying the published javadoc gives a `ClassNotFoundException`.

It arrived with #131 and the package rename left it behind; it was the last `com.lancedb` reference
in the tree. `make lint` and `make install-base` clean.


geruh (approved):
LGTM!

PR #848, opened 2026-09-21
fix: fail the TPC-DS index builder when index creation fails

jackylee-ch:
## Summary

`TpcdsIndexBuilder` catches every `ALTER TABLE ... CREATE INDEX` failure, counts it, prints
`=== Index creation complete ===` and returns normally, so the JVM exits 0. `create-indexes.sh` runs
under `set -euo pipefail`, so it announces completion too, and the operator goes on to benchmark
tables that have no indexes — measuring Lance without the zonemap pruning this script exists to set
up, which is exactly the comparison the benchmark is for.

Every statement is still attempted so all failures stay visible; only the reported outcome changes.
`System.exit` sits after the `finally` so `spark.stop()` still runs.

## Testing

Ran the built jar through `spark-submit --master local[1]` against a `--data-dir` with no Lance
tables, so all 23 statements fail:

- before: `=== Index creation complete ===`, `Succeeded: 0, Failed: 23`, `spark-submit exit=0`
- after: `=== Index creation finished with failures ===`, same counts, `spark-submit exit=1`

`make benchmark-build` and `make lint` clean.


jackylee-ch:
cc @LuciferYang

LuciferYang:
`failed` is correctly hoisted out of the `try` so it's readable after the `finally`, and `System.exit(1)` runs after `spark.stop()`. Every statement is still attempted, so the operator still sees which ones failed. The paths that skip `System.exit` (an exception out of `loadIndexStatements` or `spark.stop()`) still exit non-zero through the uncaught exception, so the exit-0-on-failure hole is closed either way. Looks correct.

LuciferYang (approved):
Approving. Verified as noted above.

LuciferYang:

